PR #3, opened 2023-07-27
sqlx migration + debugging

sebadob:
This is a big WIP.  
`sqlx` screwed up quite a few things about the `Any` driver with the 0.7 release and it does not work anymore like it did before. 

This PR is about testing out a possible split of the code into basically 2 versions behind feature flags to get rid of the `Any` driver and possibly be able to use native drivers further down the road as well, like `rusqlite` or `tokio-postgres`, if things don't work out.  

I opened up a few issues about it (https://github.com/launchbadge/sqlx/issues/2592, https://github.com/launchbadge/sqlx/issues/2644, https://github.com/launchbadge/sqlx/issues/2657), but sticking with the `Any` driver and waiting for fixes seems too uncertain to me, because it might even happen again in the future, since this driver is just not that wide spread as all other drivers.


sebadob:
Everything has been migrated away from the `Any` driver since the issues does not seem to resolve any time soon and just using the default drivers is a lot more stable in the future, since they will receive a lot of care for sure.

Tests are all green again, just needs some more real world testing before a release.

PR #131, opened 2023-10-31
helm chart creation

sebadob:
The goal is to finally create a Helm chart for Rauthy for a much easier deployment inside Kubernetes in the future.

This is a lot of work and will probably take some time and be done in between other things, when it makes sense.

### DIFF
--- a/justfile
+++ b/justfile
@@ -299,3 +299,9 @@ publish-latest: publish-versions
     docker pull ghcr.io/sebadob/rauthy:$TAG
     docker tag ghcr.io/sebadob/rauthy:$TAG ghcr.io/sebadob/rauthy:latest
     docker push ghcr.io/sebadob/rauthy:latest
+
+
+# prints out the raw computed helm template for easier development and debugging
+helm-debug:
+    #!/usr/bin/env bash
+    helm install --debug --dry-run rauthy ./rauthy-helm
--- a/rauthy-book/src/config/config.md
+++ b/rauthy-book/src/config/config.md
@@ -179,8 +179,8 @@ SMTP_FROM=
 ###### ENCRYPTION / HASHING #########
 #####################################
 
-# Format: "key_id/enc_key another_key_id/another_enc_key" - the
-# enc_key itself must be exactly 32 characters long and and
+# Format: "key_id/enc_key another_key_id/another_enc_key"
+# The enc_key itself must be exactly 32 characters long and
 # should not contain special characters.
 # The ID must match '[a-zA-Z0-9]{2,20}'
 #ENC_KEYS="bVCyTsGaggVy5yqQ/S9n7oCen53xSJLzcsmfdnBDvNrqQ63r4 q6u26onRvXVG4427/3CEC8RJWBcMkrBMkRXgx65AmJsNTghSA"
--- a/rauthy-helm/.helmignore
+++ b/rauthy-helm/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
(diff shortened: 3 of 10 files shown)


sebadob:
The Helm chart has been postponed for the moment until the config has been stabilized more towards v1.0.0.  
I might maybe even build a Kubernetes operator directly which may be able to do auto-updates in the future.

wt:
Would you be interested in a very basic helm chart that expects you to load a secret containing the config before you install the chart? I have one that I have been using for a few weeks to play with rauthy, and I would love to help you with this task.

sebadob:
Yeah regarding this issue, I am not sure if a Helm chart is still the right approach, or if Kustomize would be better.

What I want to avoid, and what most Helm charts do very badly, is re-defining every config variable (even worse when location / names change) and then create the config from it. It's not only a huge maintenance overhead, but also very unnecessary. Just because you can do this, does not mean it's a good idea. So I guess the initial approach can be thrown away.

Having the config as a separate file makes the most sense, yes. Then you only need very few templating values for setting up the services and deployments. I think Kustomize would be a better fit for that and more straight forward.

But , if you have a working Helm chart though, yes sure go ahead and add it, then maybe in combination with a section in the book. The only important thing for me is, that you should be able to do everything with it and not just some "hello world" examples. And a single install / update of the helm chart should do the trick without the need to do any manual things beforehand. It should be easy and straight forward to use it.

---

Rauthy is reaching a point where it actually makes sense to take a 2nd look at this, because it's mostly feature complete and somewhat close to the first `v1.0.0-rc` I would say. I don't expect too many breaking config changes at this point, and when the config is separate (which is where Kustomize would come in handy), you don't even need to touch the other files and just provide a copy of the complete reference config, beaucse all documentation already exists for each value.

bartekus:
We have been running Rauthy on Kubernetes in production for a while behind a
StatefulSet-based chart that is built around the hiqlite/Raft constraints you
called out here and on #1131: no autoscaling, scale-up-only semantics, stable
pod identity for Raft membership, and secrets referenced from existing K8s
Secrets rather than generated into chart values. We would like to contribute it
as the official chart and carry it to merge, including a `book/` section.

On the design question you raised in this thread (Helm vs Kustomize, and not
wanting a chart that re-templates every config variable): the chart deliberately
does **not** re-template the config. The user supplies the complete reference
`config.toml` (the documented one) as a ConfigMap or Secret and the chart mounts
it at `/app/config.toml` verbatim. The only things the chart templates are the
deployment plumbing and the handful of values that genuinely cannot live in a
static file: the Raft membership list (`HQL_NODES`, derived from the replica
count as `<pod>.<headless-svc>:8100/:8200`) and the secret references. That keeps
the maintenance surface to "services + StatefulSet + peer discovery", which is
the same small set Kustomize would template, while still giving operators a
single `helm install` / `helm upgrade` with no manual steps beforehand. If you
would still rather have a Kustomize base, the manifests are the same and we are
happy to ship that instead or in addition.

Design summary:

| Aspect | Approach |
|---|---|
| Workload | `StatefulSet` (stable identity for Raft); no `Deployment` |
| HA / Raft membership | `HQL_NODES` rendered from `replicas` via a headless service; `HQL_NODE_ID_FROM=k8s` |
| Config | full reference `config.toml` mounted from a user-supplied ConfigMap/Secret; individual vars are **not** re-templated |
| Secrets | referenced from an existing Secret (`ENC_KEYS`, `ENC_KEY_ACTIVE`, Raft/API secrets, bootstrap admin password); nothing generated into values |
| TLS | terminate at the ingress by default; in-cluster HTTPS supported by pointing the mounted config at mounted certs |
| SMTP | optional, off by default, sourced from a Secret when enabled |
| Image | pinned by tag (`appVersion`), `pullPolicy: IfNotPresent` |
| Scaling | scale-up only; **no HPA, ever**; scale-down is unsupported by Raft and documented as such |
| Out of scope | autoscaling, a bundled SMTP server, generating secrets into values |

Credit where due: @wt's MVP in #1131 is the right starting shape (StatefulSet,
headless service, ingress) and we have leaned on it as prior art. The two points
you flagged there carry straight into this design: the HPA is gone entirely (not
just gated off), and there are no default variables for values that are hardcoded
anyway.

A few questions before we open a PR, so it lands where you want it:

1. **Home** : in-repo under `charts/` (as #1131 proposed), or a separate
   `rauthy-charts` repo? Either is fine with us.
2. **Release / versioning** : chart-releaser to a `gh-pages` index, an OCI push
   to ghcr alongside the image, or just the source in-tree for now? And do you
   want `version`/`appVersion` coupled to the image release?
3. **Chart CI** : would you like a kind-based install + 1->3 scale-up + upgrade
   smoke test in the same PR, or kept as a separate commit/workflow you can drop?

Happy to send the chart for review as soon as you point us at the preferred home.

PR #137, opened 2023-11-02
Fix typos

twistedfall:
Just fixed some typos that I noticed during setup


twistedfall:
I’m perfectly fine with the license change. Thanks for sharing your work!

twistedfall:
Just one note, after merging this you will need to regenerate docs. I’ve tried doing this too, but probably my mdbook version is too different and it produces too many changes.

sebadob:
Up until now I did all these things locally and have not set up github actions yet.
I will re-generate the docs with the next PR tomorrow, thank you.

PR #138, opened 2023-11-02
Construct issuer correctly when https only serving is enabled

twistedfall:
When `LISTEN_SCHEME` is set to `https` the issuer URL is constructed using "http://" prefix, which doesn't seem right.


sebadob:
Thanks

PR #147, opened 2023-11-05
Make sure to strip binary when doing a release build

twistedfall:
That saves around 20% of the binary size in my case (55MiB → 44MiB)


sebadob:
Since we already use `panic = "abort"` it should have no negative impact to strip the binary.
Have you tested cross-compilation with a stripped one too? I sometimes run into problems when doing this.

sebadob:
I just tested all 4 release builds and it looks good so far.
Thanks!

PR #158, opened 2023-11-09
New webids logic and features

sebadob:
This PR will care about implementing a basic `webid` endpoint for each user.
If it was activated / allowed beforehand in den account page, Rauthy will provide a public webid document / endpoint in the future with optional metadata / k/v pairs that can be set by the user.


melvincarvalho:
you might want to look at recent development in the webid community group -- we're redoing the spec to allow things like JSON-LD

https://lists.w3.org/Archives/Public/public-webid/2023Nov/thread.html

sebadob:
That's actually a nice thing. I am not a big fan of XML anyway.
We are currently still optimizing the webid feature and it is implemented in a way that we can easily serve JSON-LD based on the Accept header later on.

PR #164, opened 2023-11-10
fix: webid document serialization and deserialization

damooo:
Adds serialization, deserialization for webid models. In perticular:

- Updates `WebId` model to have proper type for custom data.
- `WebidResponse`turtle serialization.
- openapi config update to include webid endpoints.

Also adds a `.devcontainer` config, so that contributors can launch in github codespaces, with required setup (rustup, cargo, musl-tools for musl target, node, npm, just, sqlx-cli, etc.). I hope it is ok.


damooo:
I have following requests though.

- Webid doc endpoint: Current endpoint that serves webid document is `auth/v1/users/{user_id}/webid`. This has a major issue that api version number `c1` exists in the path. Unlike other endpoints which are interface to crud on data, webid's uri must be permanent identifier for a user regardless of software update. The interface will never change, as data is always a dynamic turtle/other rdf formats based on content-negotiation. Thus the request is to have an endpoint without a version number. Also `webid` means just `uri`string. The document is known as `webid document` or `webid card` in ecosystem. Thus it would be great if uri has `card` instead of `webid`, like existing webid-idps.
- When `ENABLE_WEBID` is `true`, that means service admin wants all their users to have webid. The`is_open` config will not be constructive for service owners in such cases. Instead `webid_open_fields` can be provided as a user facing config for them to chose which of the fields he wants to be in his webid.

sebadob:
Thank you very much for this PR.

- Regarding the link, I will just move it into `auth/webid/{user_id}/profile` or something like that with another PR later on. If there is an API version change of Rauthy at some time, this just needs a proper reverse proxy rule, if you want to host both API versions for some time side-by-side.
- I guess the trailing `/` has to be removed from `OIDC_ISSUER`. It would match the issuer from the `openid-configuration` then.
- I would like the `resolve_webid_uri` and `resolve_webid_card_uri` better as functions on the `WebId` in `rauthy-models/src/entity/webids.rs`. These functions only make sense in conjunction with a WebId and are not really generic.
- I had a 2nd thought about the `is_open` flag and yes, your suggestion makes sense. I think we can re-name this into `expose_email` and then use it in `WebIdResponse::serialize()` as you mentioned in the TODO for the `mbox`.
- The fact that you changed the type on the `WebIdRequest.data` to the `NTriplesGraph` breaks the UI part completely and is way more complicated than just "some key value pairs". Like it is now, the user would need to provide full triples which are 100% correct and you just cannot expect that from anyone who is not really into the topic. We absolutely need to simplify this, so the user can only provide basic strings with a foaf:Property that Rauthy supports. Otherwise such a thing would lead to huge amounts of explanation for an Admin, or only technical persons would be able to use that.  
- About simplifying the custom data part, what data or actually Properties would you need there? Because most of the basic stuff like `nick`, `homepage` or `title` will come as "normal" user values anyway in the future and can be added automatically. You talked about publishing some public keys there, which would be essential. How would such an NTripleGraph look like and to which kinds of Properties can we simply this custom data and narrow it down? Because I don't find something about public keys under for instance http://xmlns.com/foaf/0.1/#term_name .

damooo:
> If there is an API version change of Rauthy at some time, this just needs a proper reverse proxy rule, if you want to host both API versions for some time side-by-side.

The issue is not just about getting response from the endpoint, but construction of webid uri itself. I.e. webid uri "string" value will be embedded into all kinds of documents, as rdf "name" of the user. For e.g. The webid profile-card served has used the webid uri reconstructed from only user_id, (and no reverse proxy info known). Thus One must know all params "at the time of uri construction itself."

> I guess the trailing / has to be removed from OIDC_ISSUER. It would match the issuer from the openid-configuration then.

Didn't checked that. I should remove that constant.

> I would like the resolve_webid_uri and resolve_webid_card_uri better as functions on the WebId

Yes.

> We absolutely need to simplify this, so the user can only provide basic strings with a foaf:Property that Rauthy supports.

> About simplifying the custom data part, what data or actually Properties would you need there?

> Because I don't find something about public keys under for instance http://xmlns.com/foaf/0.1/#term_name .

The predicates need not be just from `foaf`. They can be from any namespace. For existing properties, foaf values are standard, and hence used. There can be any number of predicates from schema.org, or any other namespace.

For example,
 - to link to user's solid pod, `solid:storage`.
 - to link to profile pic, `vcard:photo`,
 - to link to his [ldn](https://www.w3.org/TR/ldn/) inbox, `ldn:inbox`,  
 - And terms from [cert ontology](`https://www.w3.org/ns/auth/cert#`) for all things certificates.
 - terms from [Security vocab](https://w3c.github.io/vc-data-integrity/vocab/security/vocabulary.html) for keys and jwks, etc.

For eg, we can describe a jwk to be used for pub key in http-signatures as
```ttl
@prefix security: <https://w3id.org/security#> .
@prefix rdf: <http://www.w3.org/2000/01/rdf-schema#> .

<#> a security:JsonWebKey2020 ;
   security:controller </people/alice#i> ;
   security:publicKeyJwk """{
     "alg": "PS512",
     "warning": "Don't use this key for real world examples, as the private key is publicly known. This is the key from <https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures-13#page-82>.",
      "use": "sig",
      "kty":"RSA",
      "e":"AQAB",
      "n":"r4tmm3r20Wd_PbqvP1s2-QEtvpuRaV8Yq40gjUR8y2Rjxa6dpG2GXHbPfvMs8ct-Lh1GH45x28Rw3Ry53mm-oAXjyQ86OnDkZ5N8lYbggD4O3w6M6pAvLkhk95AndTrifbIFPNU8PPMO7OyrFAHqgDsznjPFmTOtCEcN2Z1FpWgchwuYLPL-Wokqltd11nqqzi-bJ9cvSKADYdUAAN5WUtzdpiy6LbTgSxP7ociU4Tn0g5I6aDZJ7A8Lzo0KSyZYoA485mqcO0GVAdVw9lq4aOT9v6d-nb4bnNkQVklLQ3fVAvJm-xdDOp9LCNCN48V2pnDOkFV6-U9nV5oyc6XI2w"
   }"""^^rdf:JSON .
```

There should be a custom ui component, so that users can add statements in familiar fashion. I should make a pr for that in near future.

sebadob:
Okay, got it.

I guess for now, we could do it the following way:

- Populate the webid document with existing default values automatically. I plan on adding a `user_profile` table or similar in the future. When a user is logged into the account page, additional values for OIDC default claims like for instance `phone` may be added here. We could then collect data from this table and populate default values for the webid endpoint from these.

- Allow custom data triples. This could be done in the current `WebID` section from the Account page. I would get rid of the `is_open` switch there and change the behavior to `expose_email` or something like that. Below, with some kind of description paragraph, allow any amount of custom triples. The custom input could be again behind another `bool` in the database like `is_expert`, which should be switched on by the user, basically provide the functionality that the `is_open` is doing now, so it does not confuse normal users that don't even know what this is. These users would never need it anyway, I guess, so this should be good.

Does this make sense?

sebadob:
> Didn't checked that. I should remove that constant.

You could re-use the `isser` from the `AppState` as well, when you just pass it to the `serialize` fn

damooo:
> The custom input could be again behind another bool in the database like is_expert, which should be switched on by the user.

Can that be just a ui-pattern instead? Where we can hide advanced-input in an accrordion, to be toggled? For example, many uis have a search bar for simple search, and a form for "Advanced search" etc.

sebadob:
> Can that be just a ui-pattern instead? Where we can hide advanced-input in an accrordion, to be toggled? For example, many uis have a search bar for simple search, and a form for "Advanced search" etc.

Yes, totally

sebadob:
Ah I totally forgot: Should I merge this PR and you would do the other stuff in a second one, or how should we do it?

damooo:
I pushed with basic suggestions incorporated, and can be merged i hope. We can iterate further on remaining.

PR #169, opened 2023-11-13
fix: take turtle as input for custom triples

damooo:
It now parses and formats data in turtle, translating to and from n-triples when required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2149,6 +2149,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "efd193069b0ddadc69c46389b740bbccdd97203899b48d09c5f7969591d6bae2"
 
 [[package]]
+name = "futures-timer"
+version = "3.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e64b03909df88034c26dc1547e8970b91f98bdb65165d6a4e9110d94263dbb2c"
+
+[[package]]
 name = "futures-util"
 version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2205,6 +2211,12 @@ name = "gimli"
 version = "0.28.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6fb8d784f27acf97159b40fc4db5ecd8aa23b9ad5ef69cdd136d3bc80665f0c0"
+
+[[package]]
+name = "glob"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
 
 [[package]]
 name = "gloo-timers"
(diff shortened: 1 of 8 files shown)


damooo:
Ideally, i would also like to

- Remove `Webid`, `WebidResponse` structs. And merge fields `expose_email`, `custom_triples` into user's info, and serialize and deserialize user object as ttl on webid-card endpoint.
- Remove `auth`, `webid` from webid uri. Webid is a string just like name, which users display as their identity. It would be great, if it is something like `https://ex.org/profile/{user_name}#me`. If `user_name` can be something simple, user relatable, much better.

sebadob:
I did not add it to the normal `users` table on purpose, because most of the time when a User is being fetched, this additional data is just obsolete. This is only needed for the `webid` endpoint and nowhere else. I would rather have it separated. 
What we actually could do in advance now is to rename that table into `users_profile`. This is another TODO on my list. This  table should care about optional additional user values like `phone` and so on. We could then merge this with the `webids` table. If we do this now, it would not require a big database migration with an upcoming release, but just an extension to the existing table.

The `/auth` in front needs to stay there. The reasons is, that with the `/auth` added to just every endpoint and even the UI part, you can easily merge a Rauthy backend into an already existing application, as long as the existing part has nothing under `/auth`. This makes merging multiple backends easy and does not require expensive path rewrites on your reverse proxy. What we could do is to expose a different URL in addition. But then we need to manage that again and show the correct address everywhere, which just makes things more complicated.  
The other thing why I did it with the `id` instead of just the `email` is that it gives some kind of passive protection from bots and stuff, because they simply cannot guess the ID's. This is a bonus. Another thing is, that there just simply is no user name. You could do it with an email, but that might change as well. The only stable thing there is the id.

damooo:
May be `webid` part can be removed from uri?, And if possible we can take custom prefix as configuration value, default suggestion being `auth`?

> The only stable thing there is the id.

Yes. Is it possible to allow custom "[slug](https://datatracker.ietf.org/doc/html/rfc5023#section-9.7)" at the time of registration?

sebadob:
Yes that's true. You can implement such a config prefix if you like. We just need to mention this properly in the config then, but that is no big deal.

damooo:
> this would prevent the removal of the /webid/ part

May be we can have webid at `/profile/<user>#me`

>  Yes that's true. You can implement such a config prefix if you like. We just need to mention this properly in the config then, but that is no big deal.

I should implement that in another PR _/\\\_

sebadob:
Okay, so then I can merge this one and you do the URL moving and so on in a new one, right?
Because the code looks good so far.

damooo:
Indeed _/\\\_.

sebadob:
I had another thought about merging the `webids` table into the upcoming  `user_profile` or something like this and I would rather not do this.

In most scenarios, the webid feature will just be disabled and not be used at all. If the current `webids` table would be merged into something else, it would mean that we would fetch unnecessary columns each time, that we don't use anyway and therefore waste precious network and cache resources.
The webid feature is opt-in and as long as you don't enable it, it should just not come with any additional overhead as much as possible.

I saw a bug in a postgres query that is related to the latest migration of the column names. I will fix that tomorrow and then have another look and test the latest implementation from your side. If everything is fine so far, I will publish v0.19.0 soon, as this will include one very important bugfix for the JWKS returned by the `openid-configuration` endpoint.

PR #220, opened 2023-12-25
Migrate to spow

sebadob:
This migrates to the way more efficient new [spow](https://github.com/sebadob/spow) crate for Proof-of-Work's, which is another project I created recently.


erlend-sh:
spow is probably preferable due to its lightness, but fyi there's also https://github.com/mCaptcha/mCaptcha

sebadob:
Yes I know about mcaptcha.
The thing is, that `spow` is using a wasm module on the client side. This makes it possible to chose way (!) higher difficulties and therefore have a way higher threshold for bots and spammers.

I have a small [example](https://github.com/sebadob/spow/tree/main/examples/svelte-wasm) in the spow repo showing the performance difference between the wasm module and a JS-only version, which also comes with `spow`.
The code is really simple in the end, but the wasm version was in tests constantly ~20 times faster than JS only.

erlend-sh:
Nice! Maybe throw in this quick explainer under the existing mini-comparison to Turnstile. Since there was no mention of mcaptcha at all I thought perhaps you hadn’t even heard of it.

sebadob:
I will mention a bit more about it with the 0.20 release I guess.
To make a better comparison to mCaptcha, I need to take a look at their code to have a better understanding, before I can write something about it. All I know for now is that they are using sha256 as well, and JS on the client side.

sebadob:
So, I took a quick look at the core code of mCaptcha, and they do things completely different.
On their side, PoW's do not have an expiry and you cannot force a difficulty beforehand. You need to communicate that in some way. Their algorithm works on any data that is serializable, which might have its use-cases for sure.

- they cannot check for a given version, which makes updating in the future most probably not easily possible.
- they can only verify, that a minimum difficulty was chosen by the client, but how high it should be and so on must be communicated outside
- They cannot verify an expiry on a PoW, which means you can use any pre-computed PoW as long as you want. You can simply send the same PoW over and over again and it will always be valid until end of time, basically. This means you would need to implement mechanisms to prevent that yourself and handle it manually.

All of this comes baked-in with `spow`. You create a PoW with a limited lifetime and the validation will check the version of `spow`, the lifetime, the challenge and verify, that the `spow` has actually been created by this very backend. In the end it will validate the difficulty and it is not possible to tamper with the original PoW at all.

mCaptcha could probably be used even on very large files (I guess?), but I don't really see a reason for doing this. It does for sure not make any sense when used in a UI. They provide a lot more stuff apart from just requesting a PoW, like rate limiting and all that stuff, which is simply not needed in Rauthy's case.

One benefit of mCaptcha though is, that the time is always constant for a given difficulty. With `spow`, you will get a median time to "solve the puzzle" and it will differ, depending on how lucky or unlucky you get with the random salt.

PR #281, opened 2024-02-17
build ui mac fix

sjud:
This ended up being the fix
https://stackoverflow.com/questions/12272065/sed-undefined-label-on-macos
Now I don't know if you would want a seperate just command for macs in particular or if there's cleaner way to specify it. But let me know!
Thanks!

### DIFF
--- a/justfile
+++ b/justfile
@@ -208,65 +208,8 @@ build-ui:
     npm run build
     cd ..
 
-    # set correct values in html outputs for pre-rendering
-    PAGES=(
-    "templates/html/*.html"
-    "templates/html/admin/*.html"
-    "templates/html/error/*.html"
-    "templates/html/error/error/*.html"
-    "templates/html/error/error/error/*.html"
-    "templates/html/oidc/*.html"
-    "templates/html/users/*.html"
-    "templates/html/users/{id}/reset/*.html"
-    "templates/html/users/{id}/email_confirm/*.html"
-    )
-    for folder in "${PAGES[@]}"; do
-        for html in $folder; do
-          # set correct document language
-          sed -i 's/lang="en"/lang="{{{{ lang }}"/g' "$html"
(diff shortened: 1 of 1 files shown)


sjud:
I don't know why this works or what it does, according to the comments 
>An alternative could be to use -i='' this worked for me (as part of a Snakemake rule) when -i '' didn't.
This also might work on linux and possibly windows? I'm not sure.

sebadob:
Thanks for your PR.

> So my sed is different than yours. :D

That's weird. Didn't know that. :D 

A nicer solution would be to actually have the same recipe and make it check what the current OS is.  
As I said, I am no Mac user, but is there something like `/etc/os-release` or something else, which could be used to check if its any Apple OS? Then it could simply be wrapped in an if clause and we keep one recipe.

In an earlier version, I used the `build.rs`, but this may not get triggered sometimes and you need workarounds to make it work properly.

Something else that would be fine for me is if we would simply use `sd` (`cargo install sd`). This should work on any system the same way, which means we would be 100% portable and it would not be a big thing to install this for local development, since you need other tools anyway.  
As I am thinking about it, this would actually be the preferred solution for now. Would you mind adding this?  
https://crates.io/crates/sd

Both solutions would work nicely and not require an additional script. If you can check for an Apple OS reliably via something like the `os-release`, then this is maybe "nicer" because you don't need an additional `cargo install sd`.

sebadob:
If you fix this small `cargo install`, I can merge it.

sjud:
> If you fix this small `cargo install`, I can merge it.

I wasn't 100% sure what you meant here but I understood this as "deleting the cargo install sd line". Please let me know if your intention was something else. Thanks! :)

sebadob:
Yes, that's exactly what I meant with it. :)

We just need to run the install cmd once for each dev environment.
Thanks.

Issue #166, opened 2023-11-12
Support social sign in


damooo:
Signing in with google, github , etc, through oidc will be highly convinient for many service instances.

sebadob:
This is something I thought about, that it may come, but this would be quite far into the future, because there are a lot of other things that are way more important ant this moment.
But yes, this is on the "it may come" TODO list already.

erlend-sh:
Social sign-ins are less oppositional to Rauthy’s value proposition if you think of them strictly as an onboarding tool rather than a default method of sign in.

For a service like [Weird](https://blog.erlend.sh/weird-web-pages), we would like to use Rauthy to eventually become our own standalone OIDC provider to compete with the IdP incumbents (Google, GitHub et.al.) in our own small way.

But for first-time user registrations, it’d be a mistake to not support the “usual suspects” as a signup method, as that’s what the vast majority of people are used to, including on Weird’s competitors like linktree and biolink.

The proprietary login options could be challenged post-signup; in Weird’s case we’d wait maybe 2-3 months for the user to fully commit to our service, and then pop the question:

> “Do you want to upgrade your login with more security and privacy features? Switch to our Weird-ID for passkey login, MFA and zero tracking!”

sebadob:
Yes, for an initial registration, it would be a nice feature, that's true. Just to grab some data that already exists.
At the moment, the benefit is not yet there, since you only can add your E-Mail and Name in the Account view, but this will change in the future.

If everything goes to plan, I will release v0.19.0 today with Solid OIDC support and some bugfixes.  
Adding more user values is planned for the release afterwards, most probably.

sebadob:
Just as an update for this issue.

I have to do quite a bit of other work, before I can implement something like this.
The next (way more important) steps are to get fully compliant with the OIDC RFC.

Some smaller things are still missing for the mandatory party and I am thinking about making Rauthy compliant with the dynamic OIDC provider spec as well. This would however be a big one, since I would need to fully support the hybrid flow + implement dynamic client registration in a way, that the DB would not get filled by bots and spammers.

When this is done, I will most probably look at implementing upstream auth providers.

Freyert:
Put another way is this: supporting [OIDC Connect discovery](https://openid.net/specs/openid-connect-discovery-1_0.html)?

sebadob:
> Put another way is this: supporting [OIDC Connect discovery](https://openid.net/specs/openid-connect-discovery-1_0.html)?

[...]

PR #286, opened 2024-02-26
Upstream Auth Providers Part 1

sebadob:
This is the first PR with the goal to implement upstream auth providers #166 .
For now, this is only a collection of thoughts and TODO's about the implementation to get it right from the beginning.

The list is not complete yet and will evolve over the next days.


sebadob:
1. part of it done.

There is not a basic openid-configuration lookup function in the UI and the basics are there for the new nav section.  
In the next PR part 2, I will add database migrations and a registration endpoint for new auth providers.

PR #293, opened 2024-03-22
Upstream Auth Providers Part 3

sebadob:
#166 

Provider auto config lookup, registration, modification and deletion has been implemented in parts 1 and 2.  
This part will take care about showing a registered provider as a login option and then handle the whole login flow.

Even though some features will be missing at the end of part 3, the basics should be implemented to make the login via an external OIDC provider work.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,7 @@ dependencies = [
  "actix-macros",
  "actix-rt",
  "actix_derive",
- "bitflags 2.4.2",
+ "bitflags 2.5.0",
  "bytes",
  "crossbeam-channel",
  "futures-core",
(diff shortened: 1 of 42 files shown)


sebadob:
I got the whole thing basically working for my dev env.  
A few things are still TODO like cleanup and resolving a few bugs that came up, because we can now basically log into an account without having any password or passkey locally. These are minor UI bugs and easily resolved.

What's missing as well is something like the possibility to "unlink" an account from an upstream provider, since this would be very handy to have. If I will add something like "link to provider" for an existing account is unsure right now.

Also, there have not been written any integration tests for the new stuff yet, which should cover everything.  
I need to think about a way how I even want to do this, since I would need an external provider / instance somewhere, if I want to test without mocking stuff.

All the left overs will be done in a final PR, so that a single one does not get too big.

PR #325, opened 2024-04-15
Feat: Add `ClaimMapping`

kerkmann:
This adds the ability to choose between `any`, `or`, `and` and `none` `JwtClaims`. This is helpful to set the `.is_admin` and `.is_user` field more flexibly.


kerkmann:
@sebadob Maybe you have some better namings, feel free to force push or mention it. :D

sebadob:
@kerkmann as we already talked about it, I like this a lot.  
I would probably rename:

`JwtClaimsSelector` -> `ClaimMapping`
`*.is_in()`  -> `*.matches()`

Thank you a lot!

kerkmann:
Thanks for the feedback, it's solved. :)

PR #334, opened 2024-04-16
DX: try to improve LSP freak out in clean environments

sebadob:
This PR is about trying to not make the LSP freak out in clean environments.
This should make it easier for new contributors to get into the project.

We cannot simply set `DATABASE_URL` for instance for sqlx compile time checked queries, because we cannot use the `Any` driver. However, to be able to easily switch between databases in dev without the need to modify the config each time, `just` has been providing this information until now.  
This works fine, but it makes the rust LSP freak out each time in environments, where you cannot set vars without interfering with the default setup.

This PR is about exploring solutions to make getting into the project smoother for new contributors.

### DIFF
--- a/justfile
+++ b/justfile
@@ -92,12 +92,12 @@ pull-latest-cross:
 
 
 # clippy with sqlite features
-clippy:
+clippy ty="sqlite":
     #!/usr/bin/env bash
     set -euxo pipefail
 
     clear
-    cargo clippy --features sqlite
+    cargo clippy
 
 
 # clippy with postgres features
@@ -106,7 +106,7 @@ clippy-postgres:
     set -euxo pipefail
 
     clear
-    DATABASE_URL=$DB_URL_POSTGRES cargo clippy
+    DATABASE_URL=$DB_URL_POSTGRES cargo clippy --features postgres
 
 
 # re-create and migrate the sqlite database with sqlx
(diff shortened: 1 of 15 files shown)


sebadob:
The logic for setting the database feature has been inverted.  
SQLite is now the default defaure instead of Postgres, since I guess that's the DB people are using for dev mostly.

The setting of the `DATABASE_URL` has been modified as well. The way it works now does not affect production deployments but only development.
- The `DATABASE_URL` will now be set in the `rauthy.cfg` by default and will always point to SQLite.
- An additional `DATABASE_URL_POSTGRES` has been added which may point to a postgres. This variable however is only
active, when also `DEV_MODE=true` in the config.
- The `.env` file you can create on your own and which is added to `.gitignore` so you don't accidentially push it, will always have the higher priority. Thie means values are read from the (default) `rauthy.cfg` first, but can be overridden if you simply re-define the same value in your `.env` file.  
Doing it this way, you don't have do modify the default config and only care about your personal `.env` with maybe additional SMTP access keys and so on.

How it works now:
When rauthy starts with  `DEV_MODE=true` and `--features postgres`, it will overwrite the `DATABASE_URL` it found in the env with the ` `DEV_MODE=true`.
This seems a bit silly at first, but it makes it possible to always have the environment set up for sqlite, so the LSP "should" not complain anymore even in a clean environment. Additionally, starting postgres is super easy as well and you can switch back and forth without the need to modify the config each time.

I also modified the `just run` recipe and got rid of `run-postgres` and `run-sqlite`.
You now simply type:
- `just run` for sqlite
- `just run postgres` for the postgres backend
- `just run ui` for the ui  
This syntax is a bit more cargo-like.

I also imroved the testing recipes. When full tests are being run, the recipe starts rauthy in test mode in the backend and then compiles all the frontend stuff. This is mandatory for the in-depth integration tests against real databases and so on.
However, when a test failed beforehand, you had to manually find the process and kill it. The PID is now being written into a file when the process has been spawned and you can `just test-backend-stop`, which will automatically run as well before starting new tests.

I hope these additions help a bit.

PR #336, opened 2024-04-17
fix: PoW CORS

### LINKED ISSUE #335: Add the ability to add a `ALLOW_ORIGIN` for the `GET /auth/v1/pow` endpoint
Right now the `GET /auth/v1/pow` endpoint is only usable (with CORS) within the web ui. We would love to get a pow challenge without disabling CORS in the client. This would add the opportunity to create a custom frontend for user registration, just fetching a pow challenge, solve that challenge and create a `POST /auth/v1/users/register` for the registration. :)

Solution 1: Add a global `ALLOW_ORIGIN` config entry, which will be used for all that endpoints (`GET /auth/v1/pow`, `POST /auth/v1/users/register`).

Solution 2 (my personal preferred solution): Adding a `client_id` to the `query` could help to set the `ALLOW_ORIGIN` for the client. So this would mean that a `pow challenge` could be obtained by a `GET /auth/v1/pow?client_id=test` and their `ALLOW_ORIGIN` entry from that specific client.

Note: I am not sure about the `POST /auth/v1/users/register` endpoint, but I think the frontend would have the same issue there as well and would ran into some CORS issues (maybe adding a client there as well?). ^^"

sebadob:
fixes #335

### DIFF
--- a/frontend/src/utils/dataFetching.js
+++ b/frontend/src/utils/dataFetching.js
@@ -116,7 +116,7 @@ export async function getProvidersTemplate() {
 
 export async function getPow() {
     return await fetch('/auth/v1/pow', {
-        method: 'GET',
+        method: 'POST',
         headers: HEADERS.json,
     });
 }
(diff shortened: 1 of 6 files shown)


sebadob:
Marging it now before the verification from @kerkmann .
I need to release a new patch anyway to fix https://rustsec.org/advisories/RUSTSEC-2024-0336.html

PR #356, opened 2024-04-22
fix: minor documentation, request body for /users/register

kerkmann:
Really a minor thing, we was confused when we looked into the swagger ui. :D


sebadob:
Oh yes thats actually nice. If you find more of these, please submit PR's.
The OpenAPI / Swagger macros cannot be compile-time checked, which is a little bit annoying.

PR #358, opened 2024-04-22
fix: Title is always showing `Password Reset` for registration

kerkmann:
Register a new user shows a `Password Reset`, this should not be the case. The String Should be in the i18n model as well. This MR is fixing this.

### DIFF
--- a/frontend/src/routes/users/{id}/reset/reset/+page.svelte
+++ b/frontend/src/routes/users/{id}/reset/reset/+page.svelte
@@ -331,7 +331,11 @@
 </script>
 
 <svelte:head>
-    <title>Password Reset</title>
+    {#if requestType.startsWith('new_user')}
+        <title>{t.newAccount}</title>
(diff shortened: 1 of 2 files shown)


kerkmann:
I am not sure if that is enough, can you please double check it? :)

sebadob:
> I am not sure if that is enough, can you please double check it? :)

That looks about right. Thanks!

sebadob:
@kerkmann please check that the UI actually builds fine next time. It was unable to compile it because of `t` being undefined. The reasons is that it was outside of the `<WithI18n>` component. I fixed this now in another PR.

kerkmann:
@sebadob Sorry, I am still struggling around with the setup, I'll take more care next time, sorry. >.<

sebadob:
No worries, was a tiny thing.

If you need to make sure everything is fine, you can run

```
just verify
```

or if you only did stuff on the UI, then 

```
just build-ui
```

is sufficient.

Issue #363, opened 2024-04-23
Feature request: add bootstrapping mechanism that can be automated instead of having to fetch generated password from the logs


taladar:
It would be great if rauthy bootstrap could be automated with software like Terraform/OpenTofu, e.g. by passing the initial admin credentials (email, password, possibly an admin API token) as environment variables the way other software does for bootstrap.

That way installing and then configuring rauthy via something like a Terraform provider could be fully automated without any manual steps.

sebadob:
I thought about adding something like this as well, yes.

taladar:
It might be a bit tricky to generate an sqlite DB file with a rauthy specific schema as part of the deployment process.

taladar:
Preparing the file manually might be fine for testing but for something like an instance accessible from the public internet it would be useful if each deployed instance had a distinct, auto-generated password (and API token if that is applicable).

sebadob:
> ...it would be useful if each deployed instance had a distinct, auto-generated password...

This is the case. That's why you see the password in the logs at the very first startup.  
The only use case I see for something like this is when providing an OIDC SaaS.
Where else what you need all of this to be fully automatic and externally handled?

Such a feature will probably come when I start developing the K8s operator at some point.

taladar:
Not really as SaaS. I am currently experimenting with Kubernetes clusters after years of using Puppet for Infrastructure as Code and I am trying to get clusters set up in a fully (apart from specifying some access tokens and names as parameters) automated and repeated way and want to use some OIDC provider in the cluster to protect observability tools like Grafana and Prometheus, ideally something with low resource usage so the VPS used as basis for the Kubernetes cluster do not need to be too large just for that kind of basic cluster infrastructure that every cluster might need.

sebadob:
Ahh I see, okay... yes in that case it makes sense.

I can add something to achieve this before starting with the k8s operator.

sebadob:
You can test with 

```
ghcr.io/sebadob/rauthy:0.22.2-20240424-2-lite
```

if you like.

taladar:
I did test this today and at least the _PLAIN variant seems to work as expected, I haven't been able to find a way to generate the password hash yet so I didn't test the other one.

Now I just need to figure out a way to bootstrap my way from there to an API key automatically to configure the new instance.

sebadob:
> Now I just need to figure out a way to bootstrap my way from there to an API key automatically to configure the new instance.

[...]

PR #365, opened 2024-04-24
feat: remove email from password reset form

### LINKED ISSUE #357: Feature request: Remove the e-mail field for `new account` page
The user needs to set their password after creating a new user account. Do we really need the e-mail field there? Because the e-mail is already known, otherwise the user couldn't click on the link in the mail. :D

image

sebadob:
fixes #357

Originally, the idea was to add improved security by requesting the E-Mail inside the password reset form. The situations where this has a benefit for most users though are pretty unlikely for "just the normal user", apart from having an evil email provider.

I talked to quite a few people about it and the tiny security benefit in certain situations is just not worth the extra confusion for users, because basically no other service I know of does it in this way, which is why it hurts the UX pretty much.

Additionally, the binding cookie for the reset form will still be set, but by default not block, only throw warnings in the logs, if a form has been requested multiple times from different hosts. Some big email providers scan their customers emails and even use links inside them, which makes the form inaccessible to the user in that case.  
So to improve the UX again, this additional security is opt-in instead of opt-out.

To fix git conflicts, this supersedes #364 for #363


sebadob:
The warning in the logs will look like this:

grafik

We could maybe add an event for this in the future if requested.

The old config var 

```
UNSAFE_NO_RESET_BINDING
```

has been removed in favor of

```
# By default, Rauthy will log a warning into the logs, if an active password
# reset form is being access multiple times from different hosts. You can set
# this to `true` to actually block any following request after the initial one.
# This hardens the security of the password reset form a bit more, but will
# create problems with E-Mail providers like Microsoft, which cans the customers
# E-Mails and even uses links inside, which make them unusable with this set to
# true.
# This feature works by setting an encrypted cookie to the host whichever opens
# the password reset form for the very first time. All subsequent requests either
# need to provide that cookie or would otherwise be rejected.
#PASSWORD_RESET_COOKIE_BINDING=true
```

This will make the setup quite a bit easier and should avoid confusion on the user side in certain situation.

PR #368, opened 2024-04-25
UI: show additional client info + user reg during login

### LINKED ISSUE #308: UI improvement, Register button in Login page (and vice versa), Go back to the source page (client)
Right now there is a encapsulated Register Screen and a encapsulated Login Screen, but both of them are not really "connected together", they are more or less standing on their own. For example, a User is getting redirected to the login screen of Rauthy, then remembers that he doesn't have an account, then he has no chance to create a new account. A small `register` button besides the login button would be awesome.

A second thing is, the user is more or less "stuck" on the login/register screen, if a user is on the login page and would like to cancel the transaction, he's completely stuck. There is no "go back" button, which would redirect the user to the source application. Of course he could use the "back function" of his browser, but the redirects are making it impossible and it's not really user friendly. ^^"

PS: I would like to help out and create a PR. The frontend and backend is running, but I have no idea how I can get test data into it. There is no rauthy client creation, no admin account, and I am a little bit stuck there. Would be awesome if someone could help me out, so I could create a PR. :heart:  Thanks!

sebadob:
fixes #308

Prerequisite work has been done with the release of v0.22 already, so we can show now a "client home page" during login, if this data exists. We also can show a link to the user registration below the login form, if it is open.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,15 @@ send huge payloads each time.
 [9f87af3](https://github.com/sebadob/rauthy/commit/9f87af3dfb49b48300b885bf406f852579470193)
 [e6d39d1](https://github.com/sebadob/rauthy/commit/e6d39d1e1118e18aeb020fbbb477a944fcd1467a)
 
+#### UX Improvement on Login
+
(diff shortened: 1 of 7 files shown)


sebadob:
This is how it looks now:

grafik

PR #371, opened 2024-04-25
Fix typo in openapi spec ErrorRresponse -> ErrorResponse

taladar:
I noticed this small typo in part of the openapi spec when trying to create a terraform provider based on it.


sebadob:
Thanks!
Interesting, that the swagger macro did not throw an error... apparently it simply does not check these.

btw when you have a provider ready, if you would want ot share it, we could add it to the repo.

taladar:
I have thought about that, this is the first time I am experimenting with writing a custom provider though so I was going to wait to see how it turns out before I was going to ask if you were interested in the result. I had planned to share though if it works okay once I am done.

PR #372, opened 2024-04-25
Fix typos in openapi spec /clients/{id}}/colors -> /clients/{id}/colors

taladar:
Another typo in the spec. Seems utoipa does not check syntax in these bits either.

I also noticed that the openapi spec does not mention authentication so the swagger-ui does not offer a field for the API key to use for testing and that the calls do not mention which API key permission they need (though that should be fairly obvious) in the generated docs.

I am not far enough yet to know if the missing authentication part will become a problem for generating the Terraform provider but if it does I can patch it locally for now and let you know.

### DIFF
--- a/rauthy-handlers/src/clients.rs
+++ b/rauthy-handlers/src/clients.rs
@@ -294,7 +294,7 @@ pub async fn put_clients(
 /// - rauthy_admin
 #[utoipa::path(
     get,
-    path = "/clients/{id}}/colors",
+    path = "/clients/{id}/colors",
     tag = "clients",
     responses(
         (status = 200, description = "Ok", body = Colors),
@@ -321,7 +321,7 @@ pub async fn get_client_colors(
 /// - rauthy_admin
 #[utoipa::path(
     put,
-    path = "/clients/{id}}/colors",
+    path = "/clients/{id}/colors",
     tag = "clients",
     request_body = ColorsRequest,
     responses(
@@ -353,7 +353,7 @@ pub async fn put_client_colors(
 /// - rauthy_admin
 #[utoipa::path(
     delete,
-    path = "/clients/{id}}/colors",
+    path = "/clients/{id}/colors",
     tag = "clients",
     responses(
         (status = 200, description = "Ok"),


sebadob:
> Another typo in the spec. Seems utoipa does not check syntax in these bits either.

That's weird. Usually, when you have typos in there, errors will pop up in the swagger ui and I always check that, when I added new endpoints. Would be really nice if the macro would compile-time check these for sure.

> I also noticed that the openapi spec does not mention authentication so the swagger-ui does not offer a field for the API key to use for testing and that the calls do not mention which API key permission they need (though that should be fairly obvious) in the generated docs.

That's correct. When I checked my options there the last time, it was just not possible. Because Rauthy does not allow admin API access with only a bearer token and the API Keys uses `API-Key` as prefix instead of `Bearer`, so I can actually distinguish them better. In the last ~6 months I have just been updating the utoipa deps but have not checked again since, if I can make this work now.

What I would like a lot better would be if I can generate the HTML from the macro at compile time and place it behind my own authentication / ui. Then I could use a default Rauthy session and CSRF token from within the Admin UI.

I will leave this PR open in case you find more typos. :D 
Thanks!

sebadob:
Regarding the API Key permissions, they should be absolutely intuitive, I hope at least.
If the permissions should match the actions you perform on the idfferent resources 1:1.

And if you do a request with an API Key and have insufficient permissions, you will actually get an answer which ones were to be expected.
But yes, the whole documentation needs some love. I also added a lot of features over the last month's that are not documented in the book yet, only in the reference config.

taladar:
I think the API key permissions aren't really a problem, I just got confused why it said
```
Permissions:
rauthy_admin
```
in the Swagger UI in so many places but it makes sense when you look at the source code and realize that the swagger UI docs are generated from the doc comments and I don't think that is really a problem given how simple the mapping from permissions to resources in the API is.

sebadob:
Ahh okay got it. Yes I did not find a nicer or prettier way to define which API endpoints require a Rauthy admin session and which ones can be accessed with an authenticated user session.

API Keys work in almost all places, except for API Keys themselves, so you would not be able to do a priv esc when a key leaks somewhere, and for upstream authentication providers for the same reason.

taladar:
Is there a particular reason the listing of providers is a POST and the creation is on /providers/create instead of the way the other entities are doing it (GET for listing, POST on the same path for creation)?

sebadob:
> Is there a particular reason the listing of providers is a POST and the creation is on /providers/create instead of the way the other entities are doing it (GET for listing, POST on the same path for creation)?

Yes, security. The "problem" with just a `GET` for `/providers` would even include the saved provider secret. But a `GET` can be issues cross origin because of LAX cookies and a `GET` will not check the CSRF token header to be present. This is just a little bit more secure, because this secret is pretty sensitive information. `POST` will only allow the request with a valid CSRF header.

PR #407, opened 2024-05-03
fix: too strict password input validation on `/authorize`

### LINKED ISSUE #406: bug: Password reset has different password policy than the login itself
A user can reset their password and use characters like `"` in the `reset` (for example `sj3478Dhksdk&jhB'\"6(4b7TB`). After redirecting the user to the login page hes unable to login with with password. The server response with a
```
Validation errors in fields:
	password: Validation error: [a-zA-Z0-9,.:/_-&?=~#!$'()*+%]+$ [{"value": String("sj3478Dhksdk&jhB'\"6(4b7TB")}]
```

This should be allowed in the `authorize` endpoint or `disallowed` in the set password endpoint. :)

sebadob:
fixes #406

### DIFF
--- a/frontend/src/components/account/AccModPwd.svelte
+++ b/frontend/src/components/account/AccModPwd.svelte
@@ -71,6 +71,11 @@
             return false;
         }
 
+        if (formValues.new.length > 256) {
+            err = 'max 256';
+            return false;
+        }
+
         if (formValues.new !== formValues.verify) {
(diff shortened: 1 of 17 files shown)


kerkmann (approved):
lgtm; Awesome! :D

sebadob:
> lgtm; Awesome! :D

...the clippy from the 1.78 update was the most amount of work here... :D

PR #441, opened 2024-05-20
feat: Experimental FedCM compatibility

sebadob:
This PR will take care of implementing experimental support for [FedCM](https://fedidcg.github.io/FedCM/#idp-api) behind a new config setting `EXPERIMENTAL_FED_CM_ENABLE`


sebadob:
It is complete and should be working now.  
However, for the implementation to make sense with OIDC, [this issue](https://github.com/fedidcg/FedCM/issues/578) should be resolved beforehand.

sebadob:
The FedCM spec currently requires you to set auth cookie with `SameSite=None`. There is an [open issue](https://github.com/fedidcg/FedCM/issues/587#issuecomment-2124092499) about this.
This PR will be on hold depending on the outcome.

anderspitman:
@sebadob is this in a basically working state? If so what are you currently returning as a token? I suspect we're basically going to want to do the same thing with Rauth and obligator (LastLogin) and would love to start testing interoperability.

sebadob:
> @sebadob is this in a basically working state? If so what are you currently returning as a token? I suspect we're basically going to want to do the same thing with Rauth and obligator (LastLogin) and would love to start testing interoperability.

I deployed a first nightly version this morning, but I am getting a weird error about some DNS issues I need to figure out.
But apart from that it "should" work.
A nightly instance with the FedCM stuff enabled is running at https://iam.sebadob.dev

Currently, I am simply returning the `id_token`, but I hope they make it possible to return any JSON so you could stick to OIDC or whatever.

sebadob:
@anderspitman  the DNS error is coming from the FedCM call inside the browser btw. I tested the API and it should be fine. All requests manually via `curl` are good so far.

anderspitman:
Awesome! I was able to create an account with GitHub federation. How do I initiate the FedCM stuff? Does rauthy implement a FedCM client, server, or both?

> Currently, I am simply returning the id_token, but I hope they make it possible to return any JSON so you could stick to OIDC or whatever.

What advantage is there of doing full OIDC? In my implementation all I want is an OIDC ID token, which is sufficient to prove the IdP vouches for the user's identity.

sebadob:
I was able to get the small "Login Start Window" now. It returned some error from the assertion endpoint I need to look into, but that means that all the others beforehand seem to be fine so far. Maybe it was just a temporary thing beforehand.

> Awesome! I was able to create an account with GitHub federation. How do I initiate the FedCM stuff? Does rauthy implement a FedCM client, server, or both?

Currently it only does the server side. I did not create any client UI yet. I am simply testing via the JS console to see if it is throwing errors.

> What advantage is there of doing full OIDC? In my implementation all I want is an OIDC ID token, which is sufficient to prove the IdP vouches for the user's identity.

You could use FedCM to handle a code flow for instance, which would make it possible to never let the token travel through the browser, but only to the RP only which would be a small security boost. Or you could actually pass refresh tokens as well, but as I am thinking about this, FedCM will probably make them kind of obsolete, because you just need 1 additional click to get a new token.

sebadob:
I think the best thing would be, if the FedCM defines a generic answer for that endpoint with maybe a superset of values that make sense, while all of them could be optional. In that case the RP can decide what to do with them and it would be compatible with a lot of IdPs directly.

sebadob:
grafik

That's how far I get now. I will add more trace / debug logging and deploy a new version to see why Rauthy is returning the 403, but I think I am almost there.

I have no idea about how to handle the login endpoint though. From a UX perspective, does it do any redirects, or just open the login in an iframe or something, so I should maybe show something like "Logged in successfully - you can close this window now" ?

anderspitman:
> Currently it only does the server side. I did not create any client UI yet. I am simply testing via the JS console to see if it is throwing errors.

:+1: 

> You could use FedCM to handle a code flow for instance, which would make it possible to never let the token travel through the browser, but only to the RP only which would be a small security boost. Or you could actually pass refresh tokens as well, but as I am thinking about this, FedCM will probably make them kind of obsolete, because you just need 1 additional click to get a new token.

That makes sense

> I think the best thing would be, if the FedCM defines a generic answer for that endpoint with maybe a superset of values that make sense, while all of them could be optional. In that case the RP can decide what to do with them and it would be compatible with a lot of IdPs directly.

Yeah there's been a lot of [discussion](https://github.com/fedidcg/FedCM/issues/585) about how to match up clients and servers. I tend to agree with you in that my instincts tell me if this isn't more strongly specified we're going to end up in a similar situation as OAuth2, where it's useful for making walled gardens but not for decentralized solutions in the same way OIDC is.

> I have no idea about how to handle the login endpoint though. From a UX perspective, does it do any redirects, or just open the login in an iframe or something, so I should maybe show something like "Logged in successfully - you can close this window now" ?

From what I can tell it opens a new browser window. When you're done you need to call `IdentityProvider.close()` which closes the window and continues the FedCM flow. The only way I've been able to get the auto login page to trigger is by manually deleting my IdP session cookies. Apparently it's only for cases where the FedCM logged in status is out of sync. Seems a bit weird to me because in the case where you're not logged in to any IdPs the request is simply going to fail, and my intuition would be that it would be better to do the auto login flow. But what if you have multiple IdPs that are all logged out? There's probably a lot of details to be worked out here.

sebadob:
> From what I can tell it opens a new browser window. When you're done you need to call IdentityProvider.close() which closes the window and continues the FedCM flow. The only way I've been able to get the auto login page to trigger is by manually deleting my IdP session cookies. Apparently it's only for cases where the FedCM logged in status is out of sync. Seems a bit weird to me because in the case where you're not logged in to any IdPs the request is simply going to fail, and my intuition would be that it would be better to do the auto login flow. But what if you have multiple IdPs that are all logged out? There's probably a lot of details to be worked out here.

Okay, that helps already.  
Well for that reason I am maintaining a second cookie for now (without having it tested beforehand). This is very long lived by default and only contains the `user_id`. In the other cookie, I have the FedCM session, which is short lived.
The problem was, when the session expired, there was no way of knowing if the user has been logged in before, if the session just expired, if the user has been deleted, and so on. So I maintain the `user_id` of the last successful login so I can at least return a `logged-out` status correctly, when the FedCM does a request to the `/accounts`.

> Does you implementation accept any value for clientId or does it have to be "fedcm"? What I'm currently doing is using the client URI, which results in a traditional ID token with the aud set properly. But I suppose if the client is using a nonce they're safe against attackers trying to use a fedcm token from a different session.

For now, this only works with pre-registered clients, because I had no idea how I can accept a URI (I would love to...). Isn't the `approved_clients` the problem here, or did I get that wrong? From my understanding, the FedCM would only even try to log in, if this contains some values it "knows" or something like that.  
Exactly that question would be me last big issue with it - how to provide any `client_id` for the login. Or can the `approved_clients` be empty and it would still work? That would be nice.

Maybe it makes sense to create a tiny UI for testing. :D

anderspitman:
> For now, this only works with pre-registered clients, because I had no idea how I can accept a URI (I would love to...). Isn't the approved_clients the problem here, or did I get that wrong? From my understanding, the FedCM would only even try to log in, if this contains some values it "knows" or something like that.
Exactly that question would be me last big issue with it - how to provide any client_id for the login. Or can the approved_clients be empty and it would still work? That would be nice.

I haven't implemented `approved_clients` at all and it's been working great. I believe if the IdP doesn't return `approved_clients` then it will show up in any client as long as they specify `configURL: "any"`, like this:

```javascript
const identityCredential = await navigator.credentials.get({
    identity: {
      context: "signin",
      providers: [
        {
          configURL: "any",
          clientId: window.location.origin,
          nonce: "fake-nonce",
        },
      ]
    },
  })
```

For example if you go to https://lastlogin.net, add at least one identity with an upstream provider, and register FedCM, then go to https://webmention.io, your identity will show up in the list. If you try to log in it will fail though, because webmention is using the IndieAuth variant of FedCM, whereas obligator is simply returning a ID token.

anderspitman:
Note the email flow for LastLogin is pretty terrible at the moment. I was trying to make it as secure as possible and went way too far down that rabbit hole. It doesn't even make sense to me anymore and I wrote it.

sebadob:
> I haven't implemented approved_clients at all and it's been working great. I believe if the IdP doesn't return approved_clients then it will show up in any client as long as they specify configURL: "any", like this:

That's really good to know - thanks! Now it makes sense how I can accept any client_id. The good thing is, all of this dynamic resolving is already implemented in Rauthy.

> Note the email flow for LastLogin is pretty terrible at the moment. I was trying to make it as secure as possible and went way too far down that rabbit hole. It doesn't even make sense to me anymore and I wrote it.

I know such situations so good... :D

sebadob:
When I do it like you mentioned:

```
const identityCredential = await navigator.credentials.get({
    identity: {
      context: "signin",
      providers: [
        {
          configURL: "any",
          clientId: window.location.origin,
          nonce: "fake-nonce",
        },
      ]
    },
  })
```

How does the FedCM know I am logged in "somewhere" ?
Do I need to implement something on the Rauthy side and after any successful login, try to register the provider with the internal FedCM API? Because until now, I did not need to do anything in the JS code on Rauthy's side.

anderspitman:
> The good thing is, all of this dynamic resolving is already implemented in Rauthy.

That was one of the things that really stuck out to me when @erlend_sh first pointed me to Rauthy. IMO these anonymous flows are really important, especially for things like self-hosted apps. You can write a self-hosted service with millions of installations, and as long as it supports OIDC everyone can use LastLogin.io without anyone needing to register.

> I know such situations so good... :D

The reality is you can't protect users from phishing with an email login. You can mitigate, but that's it. I'm hopeful things like FedCM will eventually replace email login flows.

sebadob:
> Yeah, you have to call IdentityProvider.register("https://iam.sebadob.dev/auth/v1/fed_cm/config") from a page hosted on https://iam.sebadob.dev/. And it has to be a user gesture like a click. I didn't even realize FedCM would work at all without that call. Also make sure you've followed the other steps here: https://github.com/fedidcg/FedCM/issues/240#issuecomment-2004650817

It probably worked because I returned a non-empty array for the `approved_clients` and referred to the config directly.
But that exactly was my issue and why I got stuck at the point where I wanted to accept any client. All of this makes more sense now.

However, I am a bit stuck right now, because the suggested code from the linked PR comment does not work for me.
The `IdentityProvider` does not have a `register` function. I just installed the latest Chrome DEV (canary does not exist for linux) and am on `127.0.6485.0` which should be fine. Maybe I will just ask in the linked PR.

sebadob:
@anderspitman I pushed a tiny, buggy and partly hardcoded test UI: https://iam.sebadob.dev/auth/v1/fedcm  
It is showing a network fetch error in the console for the client config, but apparently it does fetch it successfully, because the flow is working now and it is using the correct endpoints from the config itself.  
Its super hard to debug any of these messages in the browser though, because you don't get any valuable information at all, just the error without anything else. The fetch does not even show up in the network tab.

I would now just need to add an input there where you can put a URL (and fix another origin check bug I found), and then it should work with any client, that just hosts its own config. The only thing I am thinking about now is how I would want to validate the `Origin` for ephemeral clients. My first idea (and current impl) was to make sure the Origin and the `client_id` are the same, which makes sense at first. But it may also be possible, that you want to host your client config somewhere else and re-use it in a few places. This approach is simpler at first, when you have multiple ones, but less secure, because you would not be able to validate the client origin at all.  
Something else I am doing now is, when the Origin is not in the `allowed_origins` saved inside Rauthy, that I compare the allowed `redirect_uri`s and split off the origin. If one of these matches, I allow the request. I guess this makes sense and it would make it possible to host your client config in only one location and possibly re-use it for many clients, as long as you properly set all the allowed redirect URIs.  
But these are just some details that can be adjusted easily at any time.

The login is not working as well, because I added one `/` more than needed to the URL, but at least the message is shown correctly and the window opens. ^^  
I was not able to test this on localhost, because for some reason, chrome deleted the cookies each time and the whole thing only works, when it is deployed with proper TLS certificates.

anderspitman:
Nice! I'm able to register it for FedCM, but it still doesn't show up in my list of login options. I'm guessing because you're still returning the `allowed_clients`?

> It is showing a network fetch error in the console for the client config, but apparently it does fetch it successfully, because the flow is working now and it is using the correct endpoints from the config itself.
Its super hard to debug any of these messages in the browser though, because you don't get any valuable information at all, just the error without anything else. The fetch does not even show up in the network tab.

I forgot to warn you the console errors are worse than useless in my experience. At one point someone linked (probably in the big  register IdP thread) a tool that apparently shows you more network information, but I've managed to avoid using it.

> I would now just need to add an input there where you can put a URL (and fix another origin check bug I found), and then it should work with any client, that just hosts its own config.

What format would this client config need to be in?

> The only thing I am thinking about now is how I would want to validate the Origin for ephemeral clients. My first idea (and current impl) was to make sure the Origin and the client_id are the same, which makes sense at first.

This is what I'm currently doing, but I want to support fully anonymous logins, so I'm hoping we get a way for IdPs to opt out of having the Origin header sent at all. In that case the client_id becomes opaque and will probably be randomly generated by RPs that want to protect user privacy.

zicklag:
For some reason in the latest release of brave the page says that FedCM is not enabled despite me enabling every FedCM flag in the chrome://flags page. Here's my version for reference:

[Version 1.66.115 Chromium: 125.0.6422.112 (Official Build) unknown (64-bit)](https://brave.com/latest/)

Not ready to spend time investigating myself, but in case this info is useful. :)

anderspitman:
> For some reason in the latest release of brave the page says that FedCM is not enabled despite me enabling every FedCM flag in the chrome://flags page. Here's my version for reference:
> 
> [Version 1.66.115 Chromium: 125.0.6422.112 (Official Build) unknown (64-bit)](https://brave.com/latest/)
> 
> Not ready to spend time investigating myself, but in case this info is useful. :)

I think FedCM should at least partially be working in that case, but I'm pretty sure you need at least Chrome 126 in order to test the IdP registration stuff.

sebadob:
> Nice! I'm able to register it for FedCM, but it still doesn't show up in my list of login options. I'm guessing because you're still returning the allowed_clients? Is there any reason not to accept connections from any client, seeing as Rauthy already supports ephemeral clients?

It worked first, then I deleted the session cookie to see if the login works, and then it stopped working, even after a manual login. I don't know why, but maybe it had the config cached somehow.  
Yes I did return `approved_clients`, but just because this valid is strictly not specified as being optional in the spec. I absolutely want to support ephemeral clients. I think this is a critical part to make FedCM worth to use.

> I forgot to warn you the console errors are worse than useless in my experience. At one point someone linked (probably in the big register IdP thread) a tool that apparently shows you more network information, but I've managed to avoid using it.

Yeah they are really the worst. You can't even look into any code, nothing. What was weird about the above config error is that the request did not even show up in my access logs, so something must still be wrong on the browser side.

> What format would this client config need to be in?

Currently, I am using the definition from [Solid OIDC](https://solidproject.org/TR/oidc#clientids-document), but maybe the dynamic client request from the [official OIDC extension]( https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata) would make more sense.  
I haven't played around with this yet, wanted to make it work with static clients first.

I will deploy a new version where I can modify the registration values via Inputs.

sebadob:
So... after a lot of testing with these super nice debugging capabilities, I found the issue.

All the errors I got, which did not make any sense to me, were actually coming from bad / wrong provider registrations from the past. For instance, it even remembered a config URL from testing on localhost and saved it internally. It seems so me that the FedCM API then internally just iterates over the list and throws all kinds of errors on configs that do not work, and when a faulty one has been reached, it would simply not go on.  
Now it makes sense that it complained about not being able to resolve the config URL, because it (I guess) had a localhost testing url with a higher priority internally and stopped at that point. I only found it out, because I now did some tests with self-signed certs on a remote host so I don't need to do a full deployment each time, but then it complained about self-signed certs. After I registered this IdP though, I got the same error on the deployed instance with proper TLS.

@anderspitman do you know a way of clearing all registered FedCM IdP's? Because I have reset the browser to factory defaults and it still did not remove the broken entries. I then uninstalled and `rm -rf`ed `~/.config/google-chrome-unstable` to get rid of it, which is of course super annoying to do each time when you are just testing.

Also, the logging out and in is a bit bumpy. I am returning the `Set-Login` header on the `/accounts` and it works, until I actually log out and want to log in again. If I interrupt the login procedure, it kind of gets stuck and just reports `Not signed in with the identity provider.` even when I log in manually afterwards. Need to figure that out as well, probably when I built a dedicated login procedure for FedCM.

sebadob:
I got pretty far with everything after cleaning up the browser.  
The only thing left right now is that it does not show up the login in a new window, when the FedCM recognizes that I am logged out. It shows the error in the console correctly, but it does not open the window.

All other parts are fine so far.  
- I can register
- I can fetch tokens and get the type
- I get feedback if I am logged in or logged out
- If I login manually again (actually by just returning the proper headers from the backend at any point), I will get tokens again

Why it is not opening the login window currently, I have no idea though. I can work around that manually in Rauthy's case, but that would not work on some remote client that does not know about it. At that point I need to rely on the FedCM API to open that window, which is does not right now, only logging the error.

anderspitman:
Excellent progress. No time to work on this at the moment, but a couple things.

* I still can't see sebadob.dev in the list of accounts presented on lastlogin.net. No idea why.
* However, I *can* see lastlogin.net on sebadob.dev.
* If you haven't run into this already, if you bail out of a login (ie click the X instead of selecting an account), the RP domain will get blacklisted. Go to chrome://settings/content/federatedIdentityApi to reset. The error to look for is "Third-party sign in was disabled in browser Site Settings."

sebadob:
> I still can't see sebadob.dev in the list of accounts presented on lastlogin.net. No idea why.

Maybe because it has been registered in the past with these `approved_clients` returned and now it simply would not update the value, if you try to re-register? Could you try cleaning this value somehow?

> However, I can see lastlogin.net on sebadob.dev.

That is most probably coming from the change to `configUrl: "any"` - perfect.

> If you haven't run into this already, if you bail out of a login (ie click the X instead of selecting an account), the RP domain will get blacklisted. Go to chrome://settings/content/federatedIdentityApi to reset. The error to look for is "Third-party sign in was disabled in browser Site Settings."

Thanks! That will come in handy at some point. Never got into that situation so far.

I just triggered a new pipeline and will do some ephemeral client tests and then test with other clients like lastlogin.net and hopefully find out, why the login window does not appear. But I guess I am almost there. Thanks for your help!

anderspitman:
> Maybe because it has been registered in the past with these approved_clients returned and now it simply would not update the value, if you try to re-register? Could you try cleaning this value somehow?

Good point. That's likely it.

anderspitman:
Is there a reason you need the additional client information beyond just the client id? I'm not using any of that in my implementation. Just shoving the client id into an oidc id token as the aud and sending it over fedcm. Very simple.

sebadob:
> Is there a reason you need the additional client information beyond just the client id? I'm not using any of that in my implementation. Just shoving the client id into an oidc id token as the aud and sending it over fedcm. Very simple.

Every value from `EphemeralClientRequest` that is not wrapped in an `Option<_>` is mandatory. This means in addition to the `client_id`, you need to provide at least:
- `redirect_uris`
- `grant_types`

For all missing values, Rauthy will just use defaults.

The `redirect_uris` are mandatory to make sure that you can only use this client from origins that actually belong to you or should be able to receive a token. If this would not be validated, any website in the internet would be able to get a token that is only meant for your client, which would be a huge security issue ofc.

The `grant_types` are mandatory to  make sure that an initiating client only uses the flows that it is supposed to use. Most often this will just be `authorization_code`. I could make this an option as well and assume `authorization_code` by default, but what is mandatory and whats optional is actually derived from the official RFC.

For a simple setup, you just need to:
- set the `client_id` to the exact same URI as the location where the JSON document is hosted
- set the `redirect_uris` to for instance `["https://lastlogin.net/*"]` to simply whitelist everything on that domain

sebadob:
@anderspitman Since I pushed a small bug for the accounts view, I will do another deploy soon. At the same time, I made the `grant_types` an option as well, so you will not need them once this is deployed. I had a second look and they are just there for completeness. Rauthy actually ignores this value anyway and only allows flows, that have been configured on Rauthy's side from the admin beforehand.

The pipeline will take ~40 minutes I guess.

anderspitman:
> The redirect_uris are mandatory to make sure that you can only use this client from origins that actually belong to you or should be able to receive a token. If this would not be validated, any website in the internet would be able to get a token that is only meant for your client, which would be a huge security issue ofc.

For OIDC there's an alternative design which [LastLogin uses](https://lastlogin.io/developers/) and I learned from [this excellent article](https://aaronparecki.com/2018/07/07/7/oauth-for-the-open-web) is to simply require the `client_id` to be a URI and for it to be a strict prefix of the `redirect_uri`, or even require them to be exactly equal. This is still secure, and is vital for self-hosted apps where the app might not have a public URI where it can host a client config. It's also nice and simple to implement.

For FedCM, you don't even need that much. The client_id can be completely opaque (preferably random). You never do an authorization code flow. Instead, you simply return an ID token directly through the FedCM ID assertion endpoint. Protocol-wise, this is similar to using the OIDC implicit flow.

I'm [lobbying](https://github.com/fedidcg/FedCM/issues/595) for them to add a way for IdPs to opt out of the browser sending the Origin header to the assertion endpoint. This should enable completely anonymous logins, which would be excellent for self-hosted apps.

zicklag:
> The idea is to simply require the client_id to be a URI and for it to be a strict string prefix of the redirect_uri, or even require them to be exactly equal.

I really like this idea! So far I've setup client JSON's for 5 different apps ( list [here](https://git.weird.one/zicklag/oidc-config/src/branch/main) if interested ). If we could just made the client ID equal to the redirect URI, then that would mean the JSON is completely unnecessary, _and_ it avoids an extra network fetch, _and_ it avoids the need to cache the client.

The only app this wouldn't work for is the one that I had to set the signing algorithm explicitly because it didn't support ECDSA or whatever it was.

Even then, though, you could configure the signing algorithm with query string parameters in the client ID?

Or we could avoid the JSON by going JWT style and base64 encode the client JSON and use that as the client ID? I've wondered if that would work.

As long as it's secure, there's nothing that stops us from being creative here right? So far the 5 apps I've tried integrating with ephemeral clients have not had any related built-in functionality for it, it's all invisible to the app, which is really awesome.

sebadob:
> The idea is to simply require the client_id to be a URI and for it to be a strict string prefix of the redirect_uri, or even require them to be exactly equal.

This is exactly what I am already doing.  
The reason you need to specify them currently, is that you can use this json document for multiple clients this way without requiring you to host a new document for each app, which would be very annoying. But it's true that this param could be made optional and then just assume that the `client_id` must be an exact match. But you don't gain that much from it. In the end it is one line.

> I'm https://github.com/fedidcg/FedCM/issues/595 for them to add a way for IdPs to opt out of the browser sending the Origin header to the assertion endpoint. This should enable completely anonymous logins, which would be excellent for self-hosted apps.

I am not a big fan of this tbh. No origin checking may work for simple ephemeral clients, but all static clients would stop working immediately. All confidential clients for a boosted security as well.  
It also makes it impossible for you to even create security audits (see where you are logged in, which apps do have access to your personal information, ...), makes refresh tokens impossible, and so on. If a login would not share any personal information about you, I would be fine with this, but this is not the case. You could literally log in to 1000 apps with this, and if you don't remember all of them, would never know where your personal data went, how long they have access to it, what they are doing with it, and so on.

At the end of the day, if you don't want your IdP to track you, use another one, its that simple. This is one of the reasons why I created Rauthy.

> I really like this idea! So far I've setup client JSON's for 5 different apps ( list [here](https://git.weird.one/zicklag/oidc-config/src/branch/main) if interested ). If we could just made the client ID equal to the redirect URI, then that would mean the JSON is completely unnecessary, and it avoids an extra network fetch, and it avoids the need to cache the client.

> The only app this wouldn't work for is the one that I had to set the signing algorithm explicitly because it didn't support ECDSA or whatever it was.

This would work for super simple setups and "Hello World" examples, yes. But most often you want to configure stuff, which will not be possible without an additional fetch.   
Regarding the 5 docs, you don't need to host all of them, if you don't like that, 2 would achieve the same outcome in this case. Just have one with the changed token algs, and all other ones could use the same document and you just need to add the redirect_uri accordingly.

Just one other thing - I would not use the ephemeral docs in prod, if you don't need them. Static, pre-registered clients will always be faster and more secure.

> Even then, though, you could configure the signing algorithm with query string parameters in the client ID?

All possible, but totally worthless when every IdP is doing its own thing. The document above is coming from the official OIDC DCR extension and it is not even nearly complete in case of config params you can set. It contains only the ones Rauthy currently supports. If you would allow all of them as query params, you could easliy end up with a URI with thousands of chars.

> Or we could avoid the JSON by going JWT style ( loosely speaking ) and base64 encode the client JSON and use that as the client ID? I've wondered if that would work.

Basically, you can put anything in there you like. Something that a lot of people forget is, that you want to try to keep your tokens small, because they are sent with every single request and can become a really huge overhead. For instance, just compare 2 tokens while one of them has been signed with RSA512 and the other one with EdDSA.  
And then the JWT must be some standard, just the same as the JSON document and I would always prefer the JSON because of less overhead in the end.

> As long as it's secure, there's nothing that stops us from being creative here right? So far the 5 apps I've tried integrating with ephemeral clients have not had any related built-in functionality for it, it's all invisible to the app, which is really awesome.

Yes that was the idea behind it.

anderspitman:
@sebadob rather than address all your points individually, I'll just say that I think our differences are mainly around how much power an IdP should have. You're interested in using more of the available features of OAuth2/OIDC (which is totally valid). I'm interested in an IdP that does as little as possible (asserting X user controlled Y identity at Z time). Yes there are tradeoffs. Always are. But I value privacy pretty highly so I'm willing to trade quite a bit for it.

One thing I will point out since you're concerned about where the user's data is being shared. In the case I'm talking about the surface area isn't increased, because the IdP is only sharing basic, relatively immutable data such as email address and name. Once those things are shared with an app, there's no unsharing it. Contrast this with something like an OAuth API for accessing online storage and that's a different story. You definitely need to have revocation there.

The problem is OIDC and OAuth often go together, when I think in many cases they could and should be separated. For example, currently if you log in to google you're logged in to google drive automatically, because google controls both your identity and your online storage. I envision a future where (privacy-centric) IdPs are kept simple and separate from data hosting. So I want to log in to both my online storage and my apps with something like LastLogin, which can have a tiny surface area and strong privacy guarantees, instead of one entity controlling both. Note that this is already the case for non-google drive online storage, except they usually use google as one of the default IdPs...

anderspitman:
Hopefully this whole thing becomes moot because people end up using Weird/Rauthy/TakingNames.io to manage their identity on a domain they control (which I think is the logically conclusion of your suggestion to pick an IdP you trust). But I suspect there will be some intermediate steps between where we are and there, and I want to nudge protocols like FedCM in the direction of supporting more possibilities.

sebadob:
OICD is just a superset of OAuth, so yes they do the same thing, except that OIDC can do more and usually has more information about the user.

> One thing I will point out since you're concerned about where the user's data is being shared. In the case I'm talking about the surface area isn't increased, because the IdP is only sharing basic, relatively immutable data such as email address and name. Once those things are shared with an app, there's no unsharing them. 

Absolutely true, which is why I would dislike anonymous logins even more. Currently I have configured the scope for all my Rauthy clients to only provide the data they actually need, not more not less. When you are allowing anonymous logins, you can only always provide a default dataset to the downstream app that always have to contain all data an app would need. But the problem is, that some apps don't need the full information about a user. When you have a default set, you are sending it anyway without additional user consent.  
When I use an app that only needs my email and that's it, I do not want my IdP to also send full name, address, phone number, and so on. This is a way higher privacy concern than validating an origin.

sebadob:
Merging this now as it is. The login window popup is the only issue left so far, but there are a few other things left that would need some more implementation on the Google side. As soon as they deploy updates, I will do these in additional PRs.

PR #534, opened 2024-08-05
fix: token introspection

### LINKED ISSUE #530: Support `/introspect` endpoint
**Short-version:** I suggest to drop `/tokenInfo`, and instead implement `/introspect` which appears to be missing support despite being the official OIDC endpoint for this purpose?

---

## `/tokenInfo`

Presently there is a `/tokenInfo` endpoint:

> image

When querying `/.well-known/openid-configuration` for the `/introspect` token introspection endpoint, it reports `/tokenInfo` endpoint instead:

```json
{
  "introspection_endpoint":"http://localhost:8080/auth/v1/oidc/tokenInfo"
}
```

That endpoint however is not equivalent and appears to be non-standard / inconsistent across implementations? (_note all the following references have deprecated this endpoint_):
- https://auth0.com/docs/api/authentication#get-token-info
- https://help.salesforce.com/s/articleView?id=000927357&type=1
- https://www.microfocus.com/documentation/access-manager/developer-documentation-5.0/oauth-application-developer-guide/tokeninfo-endpoint-details.html (_related: [`/introspect` endpoint](https://www.microfocus.com/documentation/access-manager/developer-documentation-5.0/oauth-application-developer-guide/t48d97uwgvk1.html)_)
- https://backstage.forgerock.com/docs/am/7/oauth2-guide/legacy-oauth2-endpoints.html#varlist-oauth2-tokeninfo-endpoint (_related: [`/introspect` endpoint](https://backstage.forgerock.com/docs/am/7/oauth2-guide/varlist-oauth2-introspect-endpoint.html)_)
- https://developer.onewelcome.com/documentation/iwelcome-endpoint-details#token-info-on-tokeninfo-endpoint (_notes support ceased by this service at the end of 2020_)

I could not find any RFC where `/tokeninfo` is defined as a standardized endpoint (_despite rauthy swagger docs implying it is OIDC standard?_):
- [This](https://backstage.forgerock.com/knowledge/kb/article/a66128300#F6J8l3) claims the endpoint is part of the OAuth2 spec, but no valid reference link.
- It appears likely that [`/tokeninfo` is an endpoint defined by Google](https://cloud.google.com/docs/authentication/token-types#access-contents) that became more widely adopted before better standardization landed? Similar to what Google had done with [`XOAUTH2`](https://developers.google.com/gmail/imap/xoauth2-protocol#the_sasl_xoauth2_mechanism) before [`OAUTHBEARER`](https://datatracker.ietf.org/doc/html/rfc7628) was standardized.

I did a [brief search](https://github.com/search?q=repo%3Asebadob%2Frauthy%20tokeninfo&type=code) before raising this issue and found no context for why this was implemented. The endpoint was already present with your open-source release commit, so no context from that history either.

---

## `/introspect`

Given the above, it should be communicated more clearly that this is a legacy endpoint, or potentially considering dropping it as I'm not sure what value it might provide for rauthy? If there is demand for it for some reason, I'm sure those users would engage here to clarify why.

Instead `/tokenInfo` should be replaced by `/introspect` endpoint, with the implementation conforming to [RFC 7662 - Section 2](https://datatracker.ietf.org/doc/html/rfc7662#section-2) where this endpoint is defined.

References for the `/introspect` endpoint documented by alternative open-source services:
- https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/introspectOAuth2Token (_Ory Hydra differs from others with protected routes, [delegating to the sysadmin to decide](https://www.ory.sh/docs/hydra/self-hosted/production#exposing-administrative-and-public-api-endpoints)_)
- https://zitadel.com/docs/apis/openidoauth/endpoints#introspection_endpoint (_like most implementations, basic auth tied to client credentials, the RFC only requires the route to be protected but not how, some software like Dovecot expects basic auth support_)

---

## Feedback - OpenAPI docs

I realize any standardized APIs should probably avoid straying from case conventions, but I did find it a little odd to encounter some endpoints that used `camelCase`, and a little bit for those that used `snake_case`.
- `kebab-case` is most appropriate for these when you have the control to decide, so that might be a breaking change you'd want to consider phasing in at some point.
- For `/tokenInfo` I believe elsewhere it was most commonly `/tokeninfo`, although one resource did have `/token-info`. Mostly a non-issue given OIDC provides an endpoint to easily query these 👍 

**NOTE:** It'd also be neat if the endpoints were documented publicly online instead of needing to run `rauthy` to lookup via swagger docs. I know we're all limited with time, so it's probably just wishful thinking 😄 

If you're curious at all, the Ory Hydra OpenAPI docs are implemented via using [Redoc](https://redocly.com/docs/redoc) instead of Swagger, and [Docusaurus](https://docusaurus.io/) instead of mdBook for docs in general, to integrate those two together they've used [Redocusaurus](https://redocusaurus.vercel.app/docs/).
- Docusaurus does supports some more advanced features, but for the most part is markdown focused like most doc generators, so it shouldn't be too different if you were to consider migrating.
- I'm not sure why by both rauthy and kanidm docs on Github Pages seem rather slow to load when I would visit them each time, which isn't the case with Github Pages for docs generated with MkDocs Material (_another which I have worked with_).

sebadob:
fixes #530


CHANGELOG.md
```diff
@@ -4,6 +4,25 @@
...
+application. If so, you can disable the authentication on this endpoint with
+
+```
+# Can be set to `true` to disable authorization on `/oidc/introspect`.
+# This should usually never be done, but since the auth on that endpoint is not
+# really standardized, you may run into issues with your client app.
+# If so, please open an issue about it.
+# default: false
+DISABLE_INTROSPECT_AUTH=true
+```
```
polarathene: For anyone in future landing here, the only implementation I know of that doesn't have basic auth by default is [Ory Hydra](https://www.ory.sh/docs/hydra/self-hosted/production#exposing-administrative-and-public-api-endpoints).

The referenced lack of standardized endpoint protection is covered in [RFC 7662 - Section 2.1](https://datatracker.ietf.org/doc/html/rfc7662#section-2.1), where protecting the endpoint is required but how that is approached is flexible. A benefit of opt-out support is that you can implement alternative support such as mTLS.

I'm not sure about clients, as basic auth would always be valid AFAIK as part of the URL (_although [some clients are deprecating support](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#access_using_credentials_in_the_url)_). For `/introspect` this would be the only way for [Dovecot](https://doc.dovecot.org/configuration_manual/authentication/oauth2/#full-config-file) to pass credentials for authorization (_the [`POST` request doesn't use the `client_id` / `client_secret` settings](https://github.com/dovecot/core/blob/03316d828c31ec736d87ba51e494bcef7a259b56/src/lib-oauth2/oauth2-request.c#L323-L330), only the [`introspection_mode = auth` setting supports adding the `Authorization: Bearer` header](https://github.com/dovecot/core/blob/03316d828c31ec736d87ba51e494bcef7a259b56/src/lib-oauth2/oauth2-request.c#L250-L267) which is a `GET` request for `/userinfo` not `/introspect`_).

There may be other clients 😅

CHANGELOG.md
```diff
@@ -21,6 +40,10 @@ How Rauthy parses config variables now correctly:
 
 [28b2457](https://github.com/sebadob/rauthy/commit/28b2457a53bf31163e94a363f2009b811e1b0b76)
 
+### Bugfixe
```
polarathene, line 43:
```suggestion
### Bugfixes
```

CHANGELOG.md
```diff
@@ -21,6 +40,10 @@ How Rauthy parses config variables now correctly:
...
+
+- The token introspection endpoint was only accepting `Json` instead of `Form` requests
```
polarathene, line 45:
```suggestion
- The token introspection endpoint was only accepting requests with `Json` data, when it should have instead been with `Form` data.
```

PR #547, opened 2024-08-29
Add Simplified Chinese translations

xtexChooser:
This PR adds Simplified Chinese (zh-Hans) translations, and updates some English strings.

The next step is to let CJK characters not regarded as special character and add an option to make family name optional (Mongolian people does not have family names, and also some people may not want to use their real name).

### DIFF
--- a/frontend/src/components/account/AccInfo.svelte
+++ b/frontend/src/components/account/AccInfo.svelte
@@ -171,7 +171,7 @@
     </div>
 
     <div class="row">
-        <div class={classLabel}><b>{t.user} {t.enabled}:</b></div>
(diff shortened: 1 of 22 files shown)


xtexChooser:
I hope to distinguish from Traditional Chinese (zh-Hant). zh-Hans and zh-Hant are two writing system variants of Chinese (note that zh-{CN,HK,TW,SG,MY,MO} are wording variants, like en-US, en-GB). there are also some differences in wording habits between two. Most people can read one fluently and understand the other (not as fluent as the other).

Most projects/softwares I have seen translates them differently. (except for MediaWiki, they have a variant conversion system.

PR #548, opened 2024-08-29
Allow more CJK characters to appear in user name

xtexChooser:
This PR adds more code-blocks in Unicode Basic Multilingual Plane (BMP) to allow CJK characters.

This should include most of the common characters in names of who lives in East Asia.
Some symbols that should never appear in people's names are excluded.


sebadob:
You probably need to add the same to `src/common/src/constants.rs:140` -> `RE_USER_NAME`, right?

xtexChooser:
ah, thanks for the review! fixed.

PR #575, opened 2024-09-26
WIP: Support for extra origin schemas

gilliangoud:
Currently I'm working on a Tauri desktop application and running into the following issue:
When using deep links (a.e. tauri://anything, myapp://anything) Rauthy does not allow any login. Even when forcefully including the deeplink origin by editing the sqlite database manually.

This is marked as a WIP as the solution I've implemented is merely a bandaid for my use case and I'd like some advice on possible ramifications and needed precautions for supporting custom origin schemas.

The current state of the PR adds an environment variable named ORIGIN_SCHEME_ACCEPT_ANY and adds a check on to the schema_ok branch of the 'client.rs entity' to skip "no matching header found" if so desired.

Additionally (And I realize this should have been a different PR) I found it necessary to run 'just migrate' before 'just setup' as it was not completing due to not being able to connect to the SQLite db.

I'd like to help in a productive way instead of just asking more features from you.


sebadob:
Did you try setting `Allowed Origin` for your client in the Admin UI to something like `127.0.0.1` (don't know what Tauri sends as `Origin`) ? This should fix your issue, as it will add additional allowed origins for each client. This is usually only used, when you want to do CORS requests and do something like embed a login to Rauthy directly on another origin.

Allowing any origin with a global varibale would be a very bad approach, becuase it would open up the application to CSRF attacks.

> Additionally (And I realize this should have been a different PR) I found it necessary to run 'just migrate' before 'just setup' as it was not completing due to not being able to connect to the SQLite db.

Oh, I didn't notice that yet. I tested on my existing setup but not on a fresh clone yet. The `setup` recipe is still very new.
If we need a migration to make it work, the migration should be added to the `setup` recipe directly instead of adding to the `CONTRIBUTING.md` to keep everything simple.

gilliangoud:
So my current bandaid does not allow any and every origin to pass; it'll still have to be in the allowed origins list, it just allows non http/https origins to be allowed as possible values.
Additionally, the frontend will still think that the origin 'tauri://localhost' is a invalid value and prevent you from saving. So any alteration of the 'client' object I'm doing at database level at the moment as I'd feel discussing this was a better path before altering a lot.

Tauri will send 'tauri://localhost' when built to a binary. This can be altered/worked around by running a separate local server or the localhost plugin that receives the redirect. I'm not too happy with that route, but it'll be the next thing to roll if I cannot get Rauthy to build on my machine :)

sebadob:
> So my current bandaid does not allow any and every origin to pass; it'll still have to be in the allowed origins list, it just allows non http/https origins to be allowed as possible values.

Sorry was a bit tired yesterday when taking a look. Yes of course it does.  Instead of allowing `any` though I would rather introduce a config var which allows very specific schemes to be parsed from a `\s` list, like for instance

```
ALLOW_ADDITIONAL_ORIGIN_SCHEMES="tauri somethingelse andanotherone"
```

This should be parsed into a `Vec<_>` inside the lazy init and can be checked against in the `match` above. 

I always like to be specific when allowing exceptions rather than just accepting anything, which usually includes a huge amount of possibilities you never even thought of. For instance, even though I don't have an example right away, allowing anything in a situation like this could maybe allow an attacker to do something bad with carefully forged input value. 

But then, yes of course it makes sense to have a feature like that. You can get around this issue by proxying everything through the backend which I guess the mentioned plugins do, but that's a bit tedious.

src/models/src/entity/clients.rs
```diff
@@ -1110,13 +1110,13 @@ pub fn is_origin_external<'a>(
...
     } else {
-        match listen_scheme {
+        (match listen_scheme {
             ListenScheme::Http => scheme == "http",
             ListenScheme::Https => scheme == "https",
             ListenScheme::HttpHttps => scheme == "http" || scheme == "https",
             ListenScheme::UnixHttp => scheme == "http",
             ListenScheme::UnixHttps => scheme == "https",
-        }
+        } || ADDITIONAL_ALLOWED_ORIGIN_SCHEMES.contains(&scheme.to_string()))
```
sebadob, line 1119: This would be cleaner if integrated into the `match` directly as a catch-all like `x => ...` and no conversion with `to_string()` to avoid allocation, but use a reference.

src/common/src/constants.rs
```diff
@@ -229,6 +237,11 @@ lazy_static! {
         };
         format!("{}://{}", scheme, *PUB_URL)
     };
+    pub static ref ADDITIONAL_ALLOWED_ORIGIN_SCHEMES: Vec<String> = env::var("ADDITIONAL_ALLOWED_ORIGIN_SCHEMES")
+        .unwrap_or_else(|_| String::from(""))
+        .split(' ')
+        .map(|scheme| scheme.to_string())
```
sebadob, line 243: Not sure if we need a `filter_map()` here and check for empty `String`s, or if the `.split()` filters these out already, even when we would have whitespace at the front or end. Another solution to make sure this cannot happen would be to do a `trim()` before the `split()`.

To be 100% sure to never allow any whitespace would be to add a very simply unit test.

I am just not sure if it could be possible with an input like for instance ` tauri something  `.

sebadob:
Apart from the 2 small comments, the changes look good so far. If this solves your issue, I could merge it next week and include it in the next `0.26.0` release which would be ready now.

src/models/src/entity/clients.rs
gilliangoud: Sadly the contains function is not happy receiving a &str, but the allocation can be resolved by using 
ADDITIONAL_ALLOWED_ORIGIN_SCHEMES.iter().any(|s| s.as_str() == scheme)

I don't think it can exist within the match statement, as the listenscheme is from the server itself and would make it unreachable.

src/common/src/constants.rs
gilliangoud: Split seems to return empty values when extra spaces are used; a filter_map with a test on empty values would also get rid of unwanted extra spaces in between values.

gilliangoud:
Super, I did manage to build it for myself; so there's no rush for me.

src/models/src/entity/clients.rs
sebadob: You're right of course, yes.

PR #581, opened 2024-10-02
Allow query params in OIDC authorization endpoint

xtexChooser:


sebadob:
Do you have an example where this would be necessary?

xtexChooser:
Twitch requires a custom param to get email claims:
```
https://id.twitch.tv/oauth2/authorize?claims=%7B%22id_token%22:%7B%22email%22:null%7D%7D
```

( btw Cloudflare also have a quirk that they send HTTP Basic Auth headers without Base64 padding which is a disobey of RFC

PR #582, opened 2024-10-02
Some justfile improvments

xtexChooser:
- Only install missing tools when setup.
- podman compatibility
- pnpm compatibility

### DIFF
--- a/justfile
+++ b/justfile
@@ -34,11 +34,11 @@ setup:
     clear
 
     echo "Installing necessary tools: just, sd, mdbook, mdbook-admonish, sqlx-cli"
-    cargo install just
-    cargo install sd
-    cargo install mdbook
-    cargo install mdbook-admonish
-    cargo install sqlx-cli --no-default-features --features rustls,sqlite,postgres
+    command -v just &>/dev/null || cargo install just
+    command -v sd &>/dev/null || cargo install sd
+    command -v mdbook &>/dev/null || cargo install mdbook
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -34,20 +35,20 @@ setup:
...
-    cargo install just
-    cargo install sd
-    cargo install mdbook
-    cargo install mdbook-admonish
-    cargo install sqlx-cli --no-default-features --features rustls,sqlite,postgres
+    command -v just &>/dev/null || cargo install just
+    command -v sd &>/dev/null || cargo install sd
+    command -v mdbook &>/dev/null || cargo install mdbook
+    command -v mdbook-admonish &>/dev/null || cargo install mdbook-admonish
+    command -v sqlx &>/dev/null || cargo install sqlx-cli --no-default-features --features rustls,sqlite,postgres
```
sebadob, line 42: Before, it would update the tools as well if there are updates and just do nothing, if they are installed and up-to-date, since there is no reason not to update I would say. Now it would skip updates and you have to manually update each tool.

xtexChooser: Maybe there needs some check to determine if a binary is installed by Cargo

sebadob: Ahh got it, yes that makes sense. In that case its less error prone this way and could be improved in a future recipe which just updates these tools.

xtexChooser:
ohhhhhhh wait there are some updates

PR #592, opened 2024-10-15
Migrate `sqlx` SQLite to `hiqlite`

sebadob:
This is the 2nd issue for the migration to [Hiqlite](https://github.com/sebadob/hiqlite).  

It will take care of starting the database layer, adopting all DB access fn's and implementing migration capabilities to migrate from an existing SQLite instance.


sebadob:
Migration is complete and tests are green again.

Some things are not covered by integration tests like everything including E-Mail sending and passkeys interactions.
These have been tested manually and at first glance, they are fine too.

An automatic migration from an existing, older version sqlite has not been implemented yet, but since this PR is so big already, I will do this in another one.

PR #599, opened 2024-10-24
fix the ingress name

zzl221000:
The same name for HTTP and HTTPS routes causes them to overwrite each other.


sebadob:
Oh yes that's true of course, thanks.

PR #601, opened 2024-11-04
rework migrations between differerent databases

sebadob:
This is a follow-up issue for the migration to [Hiqlite](https://github.com/sebadob/hiqlite).

It will take care of the migrations between different databases. The whole `MIGRATE_DB_FROM` will be reworked internally to make it less error prone and work with `Hiqlite`.


sebadob:
The following migrations have been tested and are working now:

- sqlite -> hiqlite
- sqlite -> postgres
- hiqlite (sqlite) -> postgres
- postgres -> hiqlite

sebadob:
The migrations are working fine.

At the current stage, Rauthy will always create (and need) a Postgres connection, even when using Hiqlite.
This is due to the design from before and the connection pool being inside the `AppState`.
This needs to be dropped and moved out, which is already prepared, but dropping it completely will result in many things that need fixing, so this will be done in another PR again.

PR #623, opened 2024-11-20
fix: build new nightly with increased cache shutdown timeout

sebadob:
This will (hopefully) fix an annoying error that may come up in some situations when doing a rolling release of a Rauthy HA deployment inside K8s using Postgres.

This situation is a bit special, because the Raft cluster for the caching layer has no state persistence. This is usally no issue (and even wanted for an in-memory cache), but sometimes one node can get stuck because of race conditions.

I increased the shutdown wait delay inside Hiqlite from 5 to 10 seconds to see if this solves the issue (it should).
This is not the nicest solution, and I am working on a more robust and cleaner one, but this is very time consuming, because tests can only be done manually in real-world scenarios, not in integration tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -840,7 +840,7 @@ dependencies = [
  "http 1.1.0",
  "http-body 1.0.1",
  "http-body-util",
- "hyper 1.5.0",
+ "hyper 1.5.1",
  "hyper-util",
  "itoa",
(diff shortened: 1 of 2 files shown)


sebadob:
After first tests, the issues seem to be gone now with the 10 second shutdown delay.

PR #644, opened 2024-12-10
ATProto support

avdb13:


src/models/src/app_state.rs
```diff
@@ -154,6 +163,52 @@ impl AppState {
...
+            let listen_scheme = match listen_scheme {
+                ListenScheme::Http | ListenScheme::UnixHttp => "http",
+                ListenScheme::Https | ListenScheme::HttpHttps | ListenScheme::UnixHttps => "https",
+            };
+
+            let client_metadata = AtprotoClientMetadata {
+                client_id: format!(
+                    "{listen_scheme}://{public_url}/auth/v1/atproto/client_metadata"
+                ),
+                client_uri: format!("{listen_scheme}://{public_url}"),
```
sebadob, line 178: I don't know about atproto internalals in detail, but this `client_uri` could lead to issues in scenarios, where Rauthy is hosted side-by-side with another application and only routes to `/auth/v1` are routed properly by a reverse proxy. So I guess better would be to appen `/auth/v1` just as for the token issuer.

src/models/src/app_state.rs
```diff
@@ -154,6 +163,52 @@ impl AppState {
...
+            Arc::new(
+                OAuthClient::new(OAuthClientConfig {
+                    client_metadata: client_metadata.clone(),
+                    keys: None,
+                    resolver: OAuthResolverConfig {
+                        did_resolver: CommonDidResolver::new(CommonDidResolverConfig {
+                            plc_directory_url: DEFAULT_PLC_DIRECTORY_URL.to_string(),
+                            http_client: http_client.clone(),
+                        }),
+                        handle_resolver: AtprotoHandleResolver::new(AtprotoHandleResolverConfig {
```
sebadob, line 198: For everything in this block that's coming from an external dependency, I would like to have one additional module definition added, just to make very clear that it's an external function or struct.

For instance, something like just `AuthMethod` on its own makes sense in many different contexts and by writing it as `atproto::AuthMethod`, it's clear where it belongs to. Another way to solve these issues is to push this initialization into a separate file.

src/models/src/entity/atproto.rs
```diff
@@ -0,0 +1,368 @@
...
+        let link_cookie = ApiCookie::from_req(req, PROVIDER_LINK_COOKIE)
+            .and_then(|value| AuthProviderLinkCookie::try_from(value.as_str()).ok());
+
+        if email.is_none() {
+            todo!()
+        }
+
+        let claims_user_id = did.to_string();
+
+        let user_opt = match User::find_by_federation("atproto", &claims_user_id).await {
```
sebadob, line 223: The hardcoded `provider_id` here could lead to issues in case someone created an `atproto` provider beforehand.

There 2 options:
- Define `atproto` as a blacklisted id for upstream providers in UI + API POST endpoint validation
- Create a section in the book for atproto setup and mention how an atproto provider needs to be set up and added manually

The latter would provide more control and freedom but could create issues of course (have not seen all code yet).

src/models/src/entity/atproto.rs
```diff
@@ -0,0 +1,368 @@
...
+                            return Err(ErrorResponse::new(
+                                ErrorResponseType::BadRequest,
+                                "bad user_id in link cookie".to_string(),
+                            ));
+                        }
+
+                        if link.user_email != user.email {
+                            return Err(ErrorResponse::new(
+                                ErrorResponseType::BadRequest,
+                                "Invalid E-Mail".to_string(),
```
sebadob, line 254: You don't need the `.to_string()` allocation for the error reponse message.

src/models/src/entity/atproto.rs
```diff
@@ -0,0 +1,368 @@
...
+                        )));
+                    }
+                } else {
+                    None
+                }
+            }
+        };
+        debug!("user_opt:\n{:?}", user_opt);
+
+        let now = OffsetDateTime::now_utc().unix_timestamp();
```
sebadob, line 275: You can't know this yet, but it would be nice if you use `chrono` instead of `time`. I am getting rid of the `time` dependency whenever I stumble over leftovers as you only need / want one of them usually.

src/models/src/entity/atproto.rs
```diff
@@ -0,0 +1,368 @@
...
+            user.last_login = Some(now);
+            user.last_failed_login = None;
+            user.failed_login_attempts = None;
+
+            user.save(old_email).await?;
+            user
+        } else {
+            let new_user = User {
+                email: email.as_ref().unwrap().to_string(),
+                given_name: "N/A".to_string(),
```
sebadob, line 316: Can we provide something else / more meaningful than `N/A` for the `given_name` here?
The `N/A` was something I used for Github login, but it's not a nice solution, as it will throw an error in UI validation. I need to change it for Github as well so something like `Unknown`.

src/models/src/entity/atproto.rs
```diff
@@ -0,0 +1,368 @@
...
+                family_name: None,
+                roles: Default::default(),
```
sebadob, line 318: For `roles` mapping, we should either have a proper `atproto` upstream provider configured, so they can be mapped, or add a config variable with default roles for atproto logins.

Most often, a downstream client will only accept logins if certain roles / groups match. Maybe @zicklag can provide information how they use this information internally.

src/models/src/app_state.rs
avdb13: Good point, I think it's best to just make the type `AtprotoClient` a newtype struct with a `new` method to avoid the clutter here.

src/models/src/entity/atproto.rs
avdb13: I just copy-pasted most of this code from `AuthProviderIdClaims`.

src/models/src/app_state.rs
sebadob: So, if you just host Rauthy on a particular domain, this will work like expected, because it will return an `HTTP 301` to `/auth/v1` if you access `/`. But Rauthy is also deployed in siutations, which explains the sub-domain style for the UI too, where it shares the domain with the main application.

For instance, the application is running on `example.com` but does (on purpose) not utilize any route like `/auth/*`. Then Rauthy + your app can run on the exact same domain (even though not recommended usually).
The reverse proxy then will route any traffic coming in to `/auth/*` to Rauthy and everything else to your app. They will share the same domain, but this also means, that if you define `/` as the `client_uri`, it would point to your application instead of Rauthy.

src/models/src/entity/atproto.rs
sebadob: Oh okay, if this still has some `.to_string()`, they can all be removed. I did this optimization some time ago and am not sure, if I catched all of the updates.

src/models/src/entity/atproto.rs
zicklag: Sorry I didn't see this mention.

Right now the only thing we use roles for is to recognize admin users and give them access to the admin panel in Weird.

Actually I'm a little confused by the difference between roles and groups, but right now we only distinguish between users that can login at all, and users that are also admins.

sebadob: Depending on the setup, there is no difference, if you only take a single value into account. The roles + groups  become more powerful for complex situations when you combine them like `groups == engineers && roles == admin`.

@avdb13 if the typical `atproto` clients fits into the scheme of currently existing upstream IdP's, I would prefer the solution to add them there, but I guess they don't, because each user provides their own handle / callback, right?
In that case, blacklisting `atproto` as an upstream IdP name would be the best solution.

src/models/src/entity/atproto.rs
sebadob: I know it doesn't depend on auth providers, that's exactly the problem here.
Because if someone would add an upstream IdP with the id `atproto` manually, this would produce in conflicts.
You are using the `auth_provider_id` column here without it being a user-configured auth provider.

That's why I would simply blacklist `atproto` and mention it in the UI + API, that it's a reserved keyword, and all confusion would be gone. You can't misunderstand that.

ghost:
I refactored the backend to reuse `api_types::auth_providers` in order to avoid duplicating as ATProto only requires one extra field for `/authorize` (`handle`) as well as `/callback` (`iss`). These can simply be made optional. 

Not complete as there's still some refactoring left for the UI and the state check during the callback (`callback_id != payload.state`) seems to fail.

sebadob:
> I refactored the backend to reuse api_types::auth_providers in order to avoid duplicating as ATProto only requires one extra field for /authorize (handle) as well as /callback (iss). These can simply be made optional.

That's reasonable, since `atproto` kind of is an upstream provider anyway.

In `src/models/src/entity/atproto.rs` you can probably avoid a few `String` allocations, but that's an optimization for later and getting it to work first makes sense.

Just wondering why you out-commented jemalloc? Did you have any issues with it being the default?

ghost:
> Just wondering why you out-commented jemalloc? Did you have any issues with it being the default?

Didn't seem to build on my platform, `--no-default-features` suffices though. It's ready to be merged, although I would like to extract the added logic out of the authorization page as this approach feels quite messy.

Let me know what you think.

sebadob:
I need to release a `v0.30.2` patch release first and will then have a look. Thanks.

src/models/src/entity/atproto.rs
```diff
@@ -0,0 +1,292 @@
...
+        let state = serde_json::from_slice(&value)?;
+
+        Ok(Some(state))
+    }
+
+    async fn set(&self, key: String, value: InternalStateData) -> Result<(), Self::Error> {
+        let value = serde_json::to_vec(&value)?;
+
+        Self::hql()
+            .put_bytes(Cache::Atproto, key, value, CACHE_TTL_AUTH_PROVIDER_CALLBACK)
```
sebadob, line 247: `InternalStateData` implements `Serialize` + `Deserialize`, so you should be able to use `put()` and `get()` directly without serializing into and from json. Hiqlite would use `bincode` internally which is much faster and more efficient. 

Is there a reason why you are doing JSON manually?

The same would be the case for the `store::session::Session` below.

src/models/src/entity/auth_providers.rs
```diff
@@ -827,15 +853,15 @@ impl AuthProviderCallback {
...
-        }
+        // if callback_id != payload.state {
+        //     Self::delete(callback_id).await?;
+
+        //     error!("`state` does not match");
+        //     return Err(ErrorResponse::new(
+        //         ErrorResponseType::BadRequest,
+        //         "`state` does not match",
+        //     ));
+        // }
```
sebadob, line 864: Why is the state validation section still out-commented?

sebadob (changes requested):
Most comments are small things.

---

I don't know why and where this is coming from, but when I want to `just run` your branch, the global HTTP client cannot be built with min TLS version 1.3 for some reason:

```
thread 'main' panicked at src/bin/src/init_static_vars.rs:123:25:
called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Builder, source: "invalid minimum TLS version for backend" }
```

It only starts when I set the min version to `1.2`. Not sure if this only happens to this branch or if it's something else on my side. If I am not mistanken you did not even change anything for this client.

---

The Login form needs some CSS fixing:

grafik

We at least need some space between these buttons and change the level of the back button to either 2 or 3 to better guide the eye.

And I would not show the link to the user registration, when the `atproto` form is shown. This registration has nothing to do with atproto.

To avoid additing another translation for `Handle or DID`, I would probably rename it to `Handle / DID`, which works in any language. Sometimes being lazy makes sense. :)

ghost:
> It only starts when I set the min version to `1.2`. Not sure if this only happens to this branch or if it's something else on my side. If I am not mistanken you did not even change anything for this client.

https://github.com/seanmonstar/reqwest/issues/1901

sebadob:
Thanks, that looks good now.

I cannot test if it works, since I don't have any bluesky / atproto, but I guess you did testing and @zicklag will yell as well if something does not work. :)

Edit:

Do you want to do something about the Cache `put` / `get` instead of manually handling serialization?

ghost:
> Do you want to do something about the Cache `put` / `get` instead of manually handling serialization?

```
2025-06-22T11:57:37.955623Z ERROR post_provider_callback{callback_id="MZuqVPW8i0oZj8woR17mNw"}: rauthy_models::entity::auth_providers: failed to complete authorization callback for ATProto error=state store error: Bincode: Serde(AnyNotSupported)
```

sebadob:
> ```
> 2025-06-22T11:57:37.955623Z ERROR post_provider_callback{callback_id="MZuqVPW8i0oZj8woR17mNw"}: rauthy_models::entity::auth_providers: failed to complete authorization callback for ATProto error=state store error: Bincode: Serde(AnyNotSupported)
> ```

Oh it does the annoying any deserialization internally. Yes then you need to keep it as json unfortunately.

ghost:
Everything should be fine now, just confirmed that it works again.

sebadob:
> Everything should be fine now, just confirmed that it works again.

Will have another look tomorrow.

sebadob:
Thanks!

sebadob:
@zicklag do you need a nightly image, or can you build your on and test it?

zicklag:
A nightly would be helpful. Will it be a safe upgrade from 0.28.3?

sebadob:
It won't let you go from `0.28` to `0.30` directly, you need to start `0.29` at least once in between for some programmatic migrations. Afterward, you can use `ghcr.io/sebadob/rauthy:0.30.3-20250623-atproto`. This is safe to use on a `0.30` prod DB.

PR #653, opened 2024-12-13
feat: add `DISABLE_REFRESH_TOKEN_NBF` to disable `refresh_token.nbf`

### LINKED ISSUE #651: Unable to redeem refresh_token
I'm using oauth2-proxy together with rauthy to self-host, but I have issues redeeming refresh_tokens, as I get an error: {"error": "Unauthorized"}, with now apparent further explanation in the logs.

I use Rauthy ghcr.io/sebadob/rauthy:0.26.2-lite (I am unable to upgrade as of yet), and to try to make a more minimal reproduction, I was able to trigger the same behaviour easily using [oauth2c](https://github.com/cloudentity/oauth2c/releases/tag/v1.17.0).

Is there anything I may have missed with regards to refresh_tokens that may be the reason why I am not able to redeem them? Is there a rate-limit? or dont-use-before or maybe I'm doing something obviously wrong? I can make a more automated minimal reprodution if needs be, just wanted to get the question out before in case it is something obvious I have missed.

```
# scripts and commands needs oauth2c and jq installed
# run ghcr.io/sebadob/rauthy:0.26.2-lite locally
# set up client as screenshots below show
# update the client-secret in the scripts below
# then execute these two scripts in succession
./start_session.bash > tokens.json
./refresh_session.bash $(cat tokens.json | jq -r '.refresh_token')
```

#### start_session.bash
```bash
oauth2c \
    --client-id test \
    --client-secret '7mLGUT2bfyA0Gln0b22j9dEks60WBEXnjyalFVxQ4zXMTmkQNET7bTFNzwNdasDk' \
    --insecure \
    --pkce \
    --scopes "openid profile email offline_access" \
    --response-types code \
    --response-mode query \
    --grant-type authorization_code \
    --auth-method client_secret_post \
    https://localhost/auth/v1
```
#### refresh_session.bash
```bash
oauth2c \
    --client-id test \
    --client-secret '7mLGUT2bfyA0Gln0b22j9dEks60WBEXnjyalFVxQ4zXMTmkQNET7bTFNzwNdasDk' \
    --insecure \
    --pkce \
    --scopes "openid profile email offline_access" \
    --response-types code \
    --response-mode query \
    --grant-type refresh_token \
    --auth-method client_secret_post \
    --refresh-token $1 \
    https://localhost/auth/v1
```

This is my client setup

image
image

sebadob:
fixes #651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,35 @@
 # Changelog
 
+## UNRELEASED
+
+### Changes
+
+Even though not recommended at all, it is not possible to opt-out of the `refresh_token` nbf claim, and disable it.
+
+By default, A `refresh_token` will not valid before `access_token_lifetime - 60 seconds`, but some (bad) client
+implementations try to refresh `access_tokens` while they are still valid for a long time. To opt-out, you get a new
+config variable:
+
+```
+# By default, `refresh_token`s will have an `nbf` claim, making them valid
+# at `access_token_lifetime - 60 seconds`. Any usage before this time will
+# result in invalidation of not only the token itself, but also all other
+# linked sessions and tokens for this user to prevent damage in case a client
+# leaked the token by accident.
+# However, there are bad / lazy client implementations that do not respect
+# either `nbf` in the `refresh_token`, or the `exp` claim in `access_token`
+# and will refresh early while the current access_token is still valid.
+# This does not only waste resources and time, but also makes it possible
+# to have multiple valid `access_token`s at the same time for the same
+# session. You should only disable the `nbf` claim if you have a good
+# reasons to do so.
+# If disabled, the `nbf` claim will still exist, but always set to *now*.
+# default: false
+DISABLE_REFRESH_TOKEN_NBF=false
+```
+
+[]()
+
 ## v0.27.1
 
 ### Bugfix
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2464,8 +2464,9 @@ dependencies = [
 
 [[package]]
 name = "hiqlite"
-version = "0.3.1"
-source = "git+https://github.com/sebadob/hiqlite.git#05a89f8d1f2f88a056d5b055f3471140cbfca260"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4278d3e5e0093f7da889443336a82aef39d57d286ac6a0d6933ec45d1bd4dd3b"
 dependencies = [
  "argon2",
  "axum",
--- a/book/src/config/config.md
+++ b/book/src/config/config.md
@@ -62,6 +62,22 @@ deploying with Kubernetes, extract these values into Kubernetes Secrets.
 # default: false
 #DANGER_DISABLE_INTROSPECT_AUTH=false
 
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
 # Changelog
 
+## UNRELEASED
+
+### Changes
+
+Even though not recommended at all, it is not possible to opt-out of the `refresh_token` nbf claim, and disable it.
```
andoks, line 7: > it is not possible

"it is now possible"?

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+# leaked the token by accident.
+# However, there are bad / lazy client implementations that do not respect
+# either `nbf` in the `refresh_token`, or the `exp` claim in `access_token`
+# and will refresh early while the current access_token is still valid.
+# This does not only waste resources and time, but also makes it possible
+# to have multiple valid `access_token`s at the same time for the same
+# session. You should only disable the `nbf` claim if you have a good
+# reasons to do so.
+# If disabled, the `nbf` claim will still exist, but always set to *now*.
+# default: false
```
andoks, line 27: nbf = now() sounds like a great compromise for bad clients =)

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+
+```
+# By default, `refresh_token`s will have an `nbf` claim, making them valid
+# at `access_token_lifetime - 60 seconds`. Any usage before this time will
+# result in invalidation of not only the token itself, but also all other
+# linked sessions and tokens for this user to prevent damage in case a client
+# leaked the token by accident.
+# However, there are bad / lazy client implementations that do not respect
+# either `nbf` in the `refresh_token`, or the `exp` claim in `access_token`
+# and will refresh early while the current access_token is still valid.
```
andoks, line 21: I have the impression that refresh_tokens usually are opaque (which seems to be supported by at least one of the [standards](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5)), so I am not sure nbf-checking of refresh-tokens usually would be related to a projects quality :sweat_smile: 

AT expiry though, should probably best be checked (at least for clients like oauth2-proxy)

andoks (commented):
Hope you don't mind, I had a look in liù of not being able to test it. Only thing I could find was some potential "typos" in the documentation.

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
 # Changelog
 
+## UNRELEASED
+
+### Changes
+
+Even though not recommended at all, it is not possible to opt-out of the `refresh_token` nbf claim, and disable it.
+
+By default, A `refresh_token` will not valid before `access_token_lifetime - 60 seconds`, but some (bad) client
```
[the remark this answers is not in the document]
sebadob: thanks

CHANGELOG.md
sebadob: thanks

CHANGELOG.md
sebadob: I made the refresh token just another JWT to give clients the possibility to inspect it and gain further information. Even though it's non-standard, it's an additional feature.

PR #654, opened 2024-12-13
feat: don't return Ok on `/ready` during the first 5 seconds

sebadob:
To further imrove K8s rolling releases, the `/ready` endpoint will return `HTTP 503` during the first 5 seconds after startup to give the new node enough time to join an existing raft cluster.


sebadob:
Tests have shown, that this approach does not help smoothing out k8s rolling releases.

PR #667, opened 2024-12-30
fix: remove non-existent just migrate call

gepbird:
> error: Justfile does not contain recipe `migrate`

Migration is done in `backend-start` -> `postgres-start` -> `migrate-postgres`


sebadob:
Oh yeah that is a left-over from the pre-Hiqlite phase, when the additional Sqlite migration needed to be done as well. Thanks.

Issue #642, opened 2024-12-10
UI Migration + Prettify


sebadob:
This issue will keep track of the UI migration to Svelte 5 and making it prettier at the same time.

The current UI setup is a few years old and has some issues regarding DX and maintenance, as well as some CSS problems, accessibility is missing in lots of places, does not provide light / dark mode (tricky in combination with client branding), and uses a pretty complicated setup for i18n.

All of this will be solved and the work will be split into multiple PRs:

- [x] migrate everything to Svelte 5 (deps + syntax + features)
    - [x] first step: make it work again with Svelte5 while using partially legacy syntax and function
    - [x] migrate everything from JS to TS (where it makes sense)
    - [x] re-create a new Svelte5 + TS version of each component in a separate folder  
    - [x] delete old style components and make sure no errors are thrown during compilation
    - [x] move the new Svelte5 components into the original place after all dependencies have been migrated
    - [x] finalize: get rid of all left-over legacy functions everywhere
    - [x] set `runes: true` for the svelte compiler to be 100% sure that all legacy code has been migrated
- [x] change the way theming / branding is done and provide light + dark mode possibilities - the tricky part here is the branding  of the login page which is individual for each client
    - [x] create new theming backend with DB migrations, endpoints, caching, pre-compression
    - [x] include the new theme in the UI and always preload the stylesheets
    - [x] create a new component to modify custom themes
    - [x] make it possible to manually override system prefs for theming
- [x] create a new structure for API communication in a type-safe way
- [x] simplify the i18n setup to make it easier to understand
- [x] improve accessibility
    - [x] improve automatic focus and event handling in many situations
    - [x] update all components to use proper `aria` labels and other things for screenreaders
- [x] improve performance and efficiency
    - [x] move global CSS out of inline-style in each doc into cacheable, pre-compressed external css
    - [x] move all static HTML serving endpoints into a separate module and implement dynamic compression + caching
    - [x] Change the way Client Logos are served to make them infinitely cacheable
- [x] prettify the UI
    - [x] build a new Nav + SubNav for the Admin UI to make it fully responsive even down to mobile
    - [x] improve visuals and make it prettier
- [x] get a better logo which makes it easier to identify the Rauthy tab in browsers
- [x] fix #672 during the migration
- [x] Migrate the templating engine from `askama` to `rinja`
- [x] Update the Changelog

PR #671, opened 2025-01-01
Svelte 5 Migration 2

sebadob:
#642 

This PR will take care of the way theming will be done in the future.

The old / current way of handling it is pretty inefficient and does not allow to easily use light / dark theming in combination with client branding. It also does not compress dynamically created HTML with SSR inserted values.


sebadob:
The current code implements the whole backend for the new theming approach with all database migrations, structs, API endpoints, caching, pre-compression, and so on. The new (default rauthy) theme has been implemented into the `app.html` in the UI and will from now on always be preloaded and fetched (and cached).

PUT and DELETE endpoints do exist as well already and are working. However, a UI component to modify the theme has not been implemented yet and will be built in a future PR, when the clients config view in the Admin UI is being migrated to Svelte5 as well. Doing it now would mean to do the work twice, which does not make much sense.

PR #690, opened 2025-01-08
[post_logout] fix state etc using form

cocoon:
After our discussion (https://github.com/sebadob/rauthy/discussions/682), going back and forth, validating everything again, I still come to the point where I think that the state field is not getting passed to the POST method.

Please verify again if this is wrong ^^

I think it is missed to passthrough at "LogoutHtml::build", the state is available before that in the "logout_request.state".

I now tried to fix the code (and it is working like expected now) to add it, but adding it to the file:
`\frontend\src\app.html`

required to add it to all templates, too, but I don't think it would be required everywhere.
In my first test I manually added it only to the logout.html that was already generated.

What would be a better way to only add it to the logout.html?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## UNRELEASED
+
+### Changes
+
+#### I18n - Korean
+
+Korean has been added to the translations for all user-facing UI parts.
+[#670](https://github.com/sebadob/rauthy/pull/670)
+
+### Bugfix
+
+- When using the `MIGRATE_DB_FROM` in combination with a Sqlite DB as base that has been created with
+  Rauthy < v.0.20.0, the migration might fail for you in v0.27.0 because of an old value in the `config`
+  table that might not have been cleaned up properly in some cases.
+  [#669](https://github.com/sebadob/rauthy/pull/685)
+- Fixed a regression from v0.27.0 which made it impossible to use `zh-Hans` as a users language. The deserialization
+  of the database value would fail when using Hiqlite.
+  [#693](https://github.com/sebadob/rauthy/pull/693)
+
 ## v0.27.3
 
 ### Changes
--- a/frontend/src/app.html
+++ b/frontend/src/app.html
@@ -15,10 +15,6 @@
     <link rel="preload" href="/auth/v1/theme/{{client_id}}" as="style"/>
     <link rel="stylesheet" href="/auth/v1/theme/{{client_id}}"/>
     <style>
-        * {
-            box-sizing: border-box;
-        }
-
         :root {
             --col-act1: #6b3d99;
             --col-act1a: #714d99;
@@ -51,12 +47,6 @@
         a:hover {
             color: var(--col-act2a)
         }
-
-        .content {
-            width: 100%;
-            height: 100vh;
-            padding: 20px 10px;
-        }
     </style>
     %sveltekit.head%
 </head>
@@ -66,6 +56,6 @@
 <input name="rauthy-csrf-token" id="{{ csrf_token }}" type="hidden"/>
 <input name="rauthy-data" id="{{ data }}" type="hidden"/>
 <input name="rauthy-action" id="{{ action }}" type="hidden"/>
-<div>%sveltekit.body%</div>
+<div style="display: contents">%sveltekit.body%</div>
 </body>
 </html>
--- a/frontend/src/app.html
+++ b/frontend/src/app.html
@@ -66,7 +66,6 @@
 <input name="rauthy-csrf-token" id="{{ csrf_token }}" type="hidden"/>
 <input name="rauthy-data" id="{{ data }}" type="hidden"/>
 <input name="rauthy-action" id="{{ action }}" type="hidden"/>
-<input name='state' value='{{ state }}' type='hidden'/>
 <div>%sveltekit.body%</div>
 </body>
 </html>
--- a/frontend/src/app.html
+++ b/frontend/src/app.html
@@ -66,6 +66,7 @@
 <input name="rauthy-csrf-token" id="{{ csrf_token }}" type="hidden"/>
 <input name="rauthy-data" id="{{ data }}" type="hidden"/>
 <input name="rauthy-action" id="{{ action }}" type="hidden"/>
+<input name='state' value='{{ state }}' type='hidden'/>
 <div>%sveltekit.body%</div>
 </body>
 </html>
--- a/frontend/src/components/Error.svelte
+++ b/frontend/src/components/Error.svelte
@@ -1,51 +1,58 @@
 <script>
-    import { slide } from 'svelte/transition';
+    import {slide} from 'svelte/transition';
     import IconChevronRight from "$lib/icons/IconChevronRight.svelte";
     import WithI18n from "$lib/WithI18n.svelte";
     import BrowserCheck from "./BrowserCheck.svelte";
     import LangSelector from "$lib/LangSelector.svelte";
+    import Main from "$lib5/Main.svelte";
+    import ContentCenter from "$lib5/ContentCenter.svelte";
 
     let t = $state();
     let showDetails = $state(false);
 
 </script>
 
-<BrowserCheck>
-    <WithI18n bind:t content="error">
-        <h1>{t.error}</h1>
-        <p>{t.errorText}</p>
+<Main>
+    <ContentCenter>
+        <WithI18n bind:t content="error">
+            <div class="flex-col">
+                <h1>{t.error}</h1>
+                <br>
+                <p>{t.errorText}</p>
 
-        {#if t.detailsText}
-            <div
-                    role="button"
-                    tabindex="0"
-                    class="showDetails"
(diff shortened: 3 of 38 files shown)


sebadob:
The state is already added.

This code block builds the location header if a `post_logout_redirect_uri` exists:

https://github.com/sebadob/rauthy/blob/a92374eed0fe082cd328cb4b2a54c6e8b9677357/src/api/src/oidc.rs#L675

But there is actually a tiny bug in the UI I just saw when looking into this.
The UI is not using the location header, but only the redirect URI.
This means the fix can be done in even a single line.

Everything that should be needed is something like

```js
window.location.href = res.headers.get('location');
```

in https://github.com/sebadob/rauthy/blob/a92374eed0fe082cd328cb4b2a54c6e8b9677357/frontend/src/routes/oidc/logout/%2Bpage.svelte#L51

Please have a look if this fixes your issue.

cocoon:
OK learned, I can debug the "res" object there with:

`$: res, console.log(res);`

and in my case there is no location:



I will try to find out why not ...

cocoon:
Up to that line, the req object contains all data:

https://github.com/sebadob/rauthy/blob/a92374eed0fe082cd328cb4b2a54c6e8b9677357/frontend/src/routes/oidc/logout/%2Bpage.svelte#L33

So what also was working for me as a quick'n'dirty test is, to save the state and append it at the location url here:

 
```
 let mstate = ""; <------------

    onMount(async () => {
        const params = getQueryParams();
        postLogoutUri = params.post_logout_redirect_uri;
        const token = params.id_token_hint;
        const state = params.state;
        mstate = params.state; <--------------
...
async function handleRes(res) {
...
            var loc = postLogoutUri + "?state=" + mstate;
            alert(loc);
            window.location.href = loc;

```






In this screenshot the following objects are printed:

```
onMount(async () => {
        const params = getQueryParams();
        postLogoutUri = params.post_logout_redirect_uri;
        const token = params.id_token_hint;
        const state = params.state;
        
        $: state, console.log(state);

        const csrf = window.document.getElementsByName('rauthy-csrf-token')[0].id
        saveCsrfToken(csrf);

        const immediateLogout = window.document.getElementsByName('rauthy-action')[0].id
        if ('true' === immediateLogout) {
            isLoading = true;

            const req = {
                id_token_hint: token,
                post_logout_redirect_uri: postLogoutUri,
                state: state,
            };

            $: req, console.log(req);

            let res = await logout(req);
            
            $: res, console.log(res);
            
            await handleRes(res);
        }
```

sebadob:
Pass in the correct `req` data object when doing the manual logout, in the same way as for the automatic one. The backend will only build and append the location header if a `post_logout_redirect_uri` is given. This is less error prone in the future than building the redirect  URI manually again, but would work as well in this case. This UI will receive a major rework anyway, as this is still Svelte 4 syntax.

cocoon:
Found something else here:

https://github.com/sebadob/rauthy/blob/a92374eed0fe082cd328cb4b2a54c6e8b9677357/src/api/src/oidc.rs#L659

I replaced it with 
`Json(params): Json<LogoutRequest>,`

and changed:
https://github.com/sebadob/rauthy/blob/a92374eed0fe082cd328cb4b2a54c6e8b9677357/frontend/src/utils/dataFetching.js#L93

with:
`body: JSON.stringify(data),`

Then I get the data in the POST method,

but end here:

`The page’s settings blocked the loading of a resource (connect-src) at https://win10pro.lab:5001/signout-callback-oidc?state=CfDJ8... because it violates the following directive: “connect-src 'self'”`

sebadob:
> I replaced it with
Json(params): Json<LogoutRequest>,

You must not do this. In fact, the `Query` is wrong as well, it should be `Form` by the spec, that's actually a bug. With it being `Query`, it will never accept the body properly. I also just saw that the `POST /logout` may need a bit bigger rework, because after another look at the RFC, a RP should be able to use the POST endpoint directly, as long as it sends an `id_token_hint`, which is not possible right now.

The `ReqPrincipal` needs to be chaned to be optional and then check inside the endpoint if either the principal is given or the `id_token_hint` exists. But you could also just take the `state` from the params for now and append it yourself, if it exists.

cocoon:
Great, looks like there is some light.

OK, will try the FORM.

And just one more thing, the Cookie "COOKIE_SESSION_FED_CM" needs to be flagged as secure now.

> Cookie rejected because it has the “SameSite=None” attribute but is missing the “secure” attribute.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie

sebadob:
> Great, looks like there is some light.

Yeah, I actually missed it from the RFC that it should work with a direct POST from an RP as well.

> And just one more thing, the Cookie "COOKIE_SESSION_FED_CM" needs to be flagged as secure now.

Yeah, the `SameSite=None` was a major concern from my side, when I implemented it, but they didn't care so far.

sebadob:
fyi: https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout

sebadob:
> and that results in a error 415 Unsupported Media Type.

```js
export async function logout(data) {
    let body = new FormData();
    for (let key in data) {
        body.append(key, data[key]);
    }

    return await fetch('/auth/v1/oidc/logout', {
        method: 'POST',
        headers: getCsrfHeaders(),
        body,
    });
}
```

cocoon:
You had to post it quickly, right? So I had to compile again ^^
But still fails with the content error.

cocoon:
**Progress**

Found the issue from yesterday:

`let body = new FormData();`
creates multipart Form and
`getCsrfHeaders()`
sets 
`'Content-Type': 'application/json'`

I reverted my changes in this PR and replaced wit with some new changes just for easier reading:
`let body = new URLSearchParams();`
`getCsrfHeadersForm()`

Now it can extract the Form data and sets a location with state:




But I get stuck here:

cocoon:
Found out more.

The post_logout uses MOVED_PERMANENTLY if it got a state, setting a LOCATION in headers:
```
return Ok(HttpResponse::build(StatusCode::MOVED_PERMANENTLY)
.append_header((header::LOCATION, loc))
```

https://github.com/sebadob/rauthy/blob/3b8cecbd46650a8aca71172863940384f1bb11d5/src/api/src/oidc.rs#L686

MOVED_PERMANENTLY = 301


The JavaScript part you pointed me to here:
https://github.com/sebadob/rauthy/pull/690#issuecomment-2577331068

where you thought a change like this could fix it:
`window.location.href = res.headers.get('location');`

will not work because the check "if response.ok is true" is not because 301 is not in the range for OK:

https://github.com/sebadob/rauthy/blob/a92374eed0fe082cd328cb4b2a54c6e8b9677357/frontend/src/routes/oidc/logout/%2Bpage.svelte#L51

> https://developer.mozilla.org/en-US/docs/Web/API/Response

> 
> Response.ok Read only
> 
>     A boolean indicating whether the response was successful (status in the range 200 – 299) or not.


And that is correct I think, because if a location header is set it should redirect on it's own and would not need the java script "window.location.href "?

But using the location header to redirect to another URL seems to be blocked by policy as shown.
This is how it looks in Chrome:




I have seen it does an HTTP OPTIONS request then:




Instead of a LOCATION header, i set it as a header named "loc" and used this in the javascript to set window.location.href to it.
That would work.

I pushed the changes as example.
But using MOVED_PERMANENTLY without LOCATION is not so nice, but working.

sebadob:
Instead of `301`, it makes sense to change it to a `200` to make it useable by RPs directly. These will probably expect an answer in the `2xx` range anyway. The `res.ok` check in the UI to extract the location is unnecessary, because a `301` would autoamtically redirect without the need for a manual redirection.

Just return a `200` and in the UI, check for an existing `location`. If it exists, redirect to it, and to `/auth/v1` otherwise. It makese sense to use location replace as well, because you can't go "back", since the logout is a one-time operation.

cocoon:
> Just return a `200` and in the UI, check for an existing `location`

Yes just thought the same, changed it now.

cocoon:
> use location replace as well, because you can't go "back", since the logout is a one-time operation

works 👍

sebadob:
> works 👍

If you remove the `WIP` and resolve the git conflicts, I can merge it.

cocoon:
> > works 👍
> 
> If you remove the `WIP` and resolve the git conflicts, I can merge it.

done.

You have seen I had to add another header to use form? It is a bit duplicated now.

```
function getCsrfHeadersForm() {
    return {
        ...HEADERS.form,
        'csrf-token': getCsrfToken(),
    }
}
```

sebadob:
> You have seen I had to add another header to use form? It is a bit duplicated now.

Yes I did. No worries though, the CSRF token will be removed anyway, when I rework the `/logout` endpoint that it will work with a direct `POST` from the RP in the future. I created an issue for this with #692 .

But this PR still has git conflicts, which is why I can't merge it. You should resolve these locally and push a merge from main for instance.

cocoon:
> But this PR still has git conflicts, which is why I can't merge it. You should resolve these locally and push a merge from main for instance.

🙈 resolved the conflicts, marked them but did not commit.

PR #701, opened 2025-01-18
Fix cleanup of static files in `build-ui` target

Sp1l:
The static files cleanup in the `justfile` doesn't do anything: it runs in the wrong directory.

https://github.com/sebadob/rauthy/blob/20e447c6a2dbe68689a47b8538d79b0aa0380cc1/justfile#L294

This runs in the root of the `rauthy-0.27.3` extracted files, not in the `frontend` dir.

### DIFF
--- a/justfile
+++ b/justfile
@@ -290,6 +290,8 @@ build-ui:
     #!/usr/bin/env bash
     set -euxo pipefail
 
+    cd frontend
+
     # cleanup old files
     rm -rf static/v1/*
     rm -rf templates/html/*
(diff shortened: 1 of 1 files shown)


sebadob:
> This runs in the root of the rauthy-0.27.3 extracted files, not in the frontend dir.

Yes, and that's correct. Check https://github.com/sebadob/rauthy/blob/20e447c6a2dbe68689a47b8538d79b0aa0380cc1/frontend/svelte.config.js#L15

Sp1l:
Sorry for the noise @sedabob 😆 

Thanks

PR #712, opened 2025-01-24
make compilable on windows removing unixsockets if windows

cocoon:
As I have seen you added the fix in hiqlite to make it compatible with windows here:
https://github.com/sebadob/hiqlite/commit/c6113b57b96cae2551fa3c590bc97d11ee0c9e96

I thought maybe you would also add the small fixes in rauthy, as it is not much and would make life for me easier, so that I don't have to patch it ^^

The only thing left is to target the hiqlite branch with the fix currently
`hiqlite = { git = "https://github.com/sebadob/hiqlite.git", features = ["full"], rev = "626c6877c54f4c6bad360d02053853f6d3d4fefc" }`

as the latest commits had a problem that I think will be resolved in future:
`failed to select a version for `libsqlite3-sys``

### DIFF
--- a/src/api/src/openapi.rs
+++ b/src/api/src/openapi.rs
@@ -322,6 +322,15 @@ impl ApiDoc {
         // contact.email = Some(ADMIN);
         // doc.info.contact = Some(contact);
 
+        #[cfg(target_os = "windows")]
+        let scheme = if !*PROXY_MODE && app_state.listen_scheme == ListenScheme::Http
+        {
+            "http://"
+        } else {
+            "https://"
+        };
+        
+        #[cfg(not(target_os = "windows"))]
         let scheme = if (!*PROXY_MODE && app_state.listen_scheme == ListenScheme::Http)
             || app_state.listen_scheme == ListenScheme::UnixHttp
         {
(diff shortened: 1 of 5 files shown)


sebadob:
> I thought maybe you would also add the small fixes in rauthy, as it is not much and would make life for me easier, so that I don't have to patch it ^^

I don't own a single windows PC and in fact, I don't even have windows toolchains and all that stuff installed for rust, so it's not that "quick and easy" to do something for an OS I don't even have available.

The `asm` issue is fixed for `hiqlite`, but I can't update Rauthy currently to the latest version, because `sqlx` blocks the upgrade (once again...) because of sqlite dependencies under the hood. To finally solve it, I would need to get rid of `sqlx` for all DB migrations, or wait a little bit until they hopefully update their sqlite under the hood.
On the long run, I want to drop it completely, because it's so annoying.

sebadob:
> You could use Github action to compile it for windows and see if it fails, would also make it again easier if binaries are available directly.

Yes creating some github actions at least for tests and clippy checks is still on my TODO.
Thanks for the PR.

Issue #742, opened 2025-03-05
Missing i18n translations


sebadob:
After finishing the recent migration to Svelte 5 (#642), additional necessary keys for i18n have been added. The `zh-Hans` and `ko` files are missing some translations. I copy & pasted most values from the old setup, but can't do anything about the new ones, because I don't know anything about them.

The files themselves are complete, but some entries contain english as placeholders. The files that need updates can be found in [frontend/src/i18n/common](https://github.com/sebadob/rauthy/tree/main/frontend/src/i18n/common): `zh.ts` + `ko.ts`

@xtexx do you have some time to update the `zh` translations?
@GoodLucky777 can you do the korean ones?

- [x] `zh-Hans` updated
- [x] `ko` updated

xtexx:
Sorry for being late. I was busy with school things last week. #769 should update translations for Simplified Chinese.

athene20:
I have a question about translations.

![Image](https://github.com/user-attachments/assets/3b2bd228-d5af-41af-848e-4783c5638f3c)

Invalid E-Mail Address or Verify E-Mail Address seems to be the correct message, but is Valid E-Mail Address the correct wording?
Is this message used in other places?

xtexx:
ah. Thanks for your finding. If so, the zh translation should be `无效的电子邮件地址`.

sebadob:
> Invalid E-Mail Address or Verify E-Mail Address seems to be the correct message, but is Valid E-Mail Address the correct wording?
Is this message used in other places?

Yes and no. Usually you will see `Invalid E-Mail Address`. Only on the login form though, I changed it a bit, because I found out, that some people have gotten confused when they would immediately see `Invalid E-Mail Address` when they start typing, because they though they did something wrong, even though everything works as expected.

Maybe changing this message to `Provide valid E-Mail address` makes more sense in that case and would reduce confusion.
What do you think about that? That was the one thing I was not really sure about and wanted to wait for some external input.

athene20:
> Maybe changing this message to `Provide valid E-Mail address` makes more sense in that case and would reduce confusion. What do you think about that? That was the one thing I was not really sure about and wanted to wait for some external input.

I think this is a more appropriate message.

PR #775, opened 2025-03-15
i18n: update Korean translations

athene20:
Updated Korean to address #742.
Didn't change the default value of FILTER_LANG_ADMIN.


sebadob:
Thank you!

PR #822, opened 2025-04-07
feat: Drop `sqlx` in favor of `tokio-postgres`

sebadob:
`sqlx` has been a blocker many times in the past, like it is since a few months again, and it has created quite a few issues. It has some very nice features, but the headaches are not worth it.

This PR will take care of dropping `sqlx` completely. It has already been done for `sqlite` with the migration to `hiqlite`. Now it's about dropping the Postgres part of it as well. This will be a huge amount of work, but pay off on the long run.


sebadob:
The `sqlx` dependency has been removed completely now and integration tests are green again.
However, there are some flows, which are not covered by integration tests. Most things that require human interaction or complex flows like additional deployments and stuff. These will need more manual testing and if fixes are needed, they will come in new PRs.

PR #837, opened 2025-04-11
Feat: SCIM support

### LINKED ISSUE #666: feat: SCIM support
After the [UI Migration](https://github.com/sebadob/rauthy/issues/642), [User Avatars](https://github.com/sebadob/rauthy/issues/558) and [Backchannel Logout](https://github.com/sebadob/rauthy/issues/619) are done, I will add the most important [SCIM](https://www.rfc-editor.org/rfc/rfc7644) client functionality, which is `/Users` CRUD operations. Additionally, `/Groups` will probably be implemented as well.

- [x] UI Migration finished
- [x] User Avatars finished
- [x] Backchannel Logout finished

**Resources**

- SCIM v2 has been finished in [RFC7643](https://www.rfc-editor.org/rfc/rfc7643) and [RFC7644](https://www.rfc-editor.org/rfc/rfc7644)
- Check some well known resources like [aws SCIM setup](https://docs.aws.amazon.com/singlesignon/latest/userguide/provision-automatically.html) or [Okta SCIM Documentation](https://developer.okta.com/docs/api/openapi/okta-scim/guides/scim-20/) and use them as a rough guideline.
- [scim_v2](https://crates.io/crates/scim_v2) has already implemented all the core schemas

sebadob:
fixes #666


sebadob:
Everything should be working as expected now.

SCIM tests cannot be easily done inside integration tests, because you need an external service provider if you want to do proper testing. In the future, the idea is to integrate SCIM functionality into the `rauthy-client` which could then be used inside integration tests.

For now, the implementation has been manually tested against https://scim.dev and everything looks fine so far and works as expected, under the assumption that the scim.dev implementation is correct of course, but even before testing, everything has been implementing while following the RFC closely.

The next release will therefore have SCIM support in beta state, which will then be ready for real world testing, but I am pretty confident that everything works out nicely.

PR #840, opened 2025-04-12
feat: add SMTP_PORT option and remove SMTP_DANGER_INSECURE_PORT.

zicklag:


sebadob:
Hey, thanks for the issue.

Are you still working on it? Because it's not a Draft, but has conflicts.

zicklag:
Ah, yeah, I can fix the conflict, I had it based on the latest release tag because I was trying to get it built for a hotfix, but never figured out how to get the build to work.

I'll rebase it on to main.

zicklag:
There rebased.

sebadob:
> I was trying to get it built for a hotfix, but never figured out how to get the build to work.

What do you mean with that? What were your issues with building?

sebadob:
Ah okay got it.
Then I have good news for you, I guess.
I have fully droppex `sqlx` recently. This means the build process recipes could be updates, as they don't need a database anymore, if you skip testing. This means you could test on local (at least if it's just for a hotfix), and skip testing during `build` with small modifications.

zicklag:
I can wait 'till the next release, but thanks!

I've got to to contact Digital Ocean support. Apparently they randomly started blocking outbound SMTP ports without telling us. 🤦‍♂️

This PR was to let us use the non-standard 2525 port that postmark allows us to connect under so that we avoid the block.

sebadob:
> I've got to to contact Digital Ocean support. Apparently they randomly started blocking outbound SMTP ports without telling us. 🤦‍♂️

Probably to avoid spammers? But without telling is really bad.

> This PR was to let us use the non-standard 2525 port that postmark allows us to connect under so that we avoid the block.

I was actually wondering what anyone would need a custom SMTP port in production for, but that makes sense. ^^
At least if it really helps and their blocks are that simple by port only and nothing else.

PR #890, opened 2025-05-02
remove erroneous character

bkw:
I believe this was added in 9bbbb8d by accident.


sebadob:
Oops, yes that's true. Thanks!

PR #946, opened 2025-05-17
onboarding improvements

### LINKED ISSUE #945: New developer experience / on-boarding is a bit difficult
### Discussed in https://github.com/sebadob/rauthy/discussions/944

<div type='discussions-op-text'>

<sup>Originally posted by **lvaroqui** May 17, 2025</sup>
Hello @sebadob,

As I've began to work on rauthy in the context of #872, I just wanted to share a bit of feedback on my dev experience so far.

**DISCLAIMER**: I don't want to be mean at all, I want to share info that could help make the on-boarding of new devs a better experience. To be honest my first few minutes with the project almost made me abandon the idea of contributing. This feedback is based on my own sensibility and I totally understand every one would not agree.

1. The `CONTRIBUTING.md` is missing the `just setup` step (which was in it before but was removed?).
2. Without a `.env` file with the `DEV_MODE=true` in it, the front-end will not actually build and this is not stated in `CONTRIBUTING.md`.
3. The generation of UI files that makes the git repo instantly dirty with ~700 changes is maybe the worst part:
    Committing generated artifact is commonly accepted as a bad practice. At least in general, the generated file are stable, meaning that if I do modify the UI, I shouldn't have modified files in my repo. Having to deal with such shenanigans at work, committed generated files are especially annoying when rebasing as every rebase must rebuild the files (because git does not understand how to solve the issue).
    I understand this is done to be able to build on FreeBSD but I feel like generating the files in CI in a Linux env and sending them to a FreeBSD env afterward is a much better solution than forcing developer to deal with committed generated files.
4. I don't like that even after stopping the `just run` command some docker container still run, I should not have any process remaining when I stop my dev env.
   For this, I really think that having a docker-compose flow would help immensly and would also avoid running `just` command to setup everything. The whole dev env could be embedded in a compose setup and a simple `docker compose up` would spin up everything.
   This would also avoid the "container already started issue".
5. The way `hiqlite` is integrated is a bit annoying to work with as all DB related stuff is duplicated. That's not too bad and understandable for things like migrations, but way harder to deal with when every access to the DB has to do twice the work with the infamous `if is_hiqlite()`.

So these are my main complaints and I think points 1 through 4 could be resolved quite easily!

There also a lack of dev documentation (regarding architecture, db migration, ui internationalization, ui styling...) but that's understandable as this is mostly a one man project and I was able to get around quite easily in the end, so not that bad!</div>

sebadob:
fixes #945


sebadob:
@lvaroqui Do these changes and doc update make sense to you?

CONTRIBUTING.md
```diff
@@ -184,6 +218,91 @@ rebuilt the UI, you will have the static HTML files updates, and you see the lat
...
+
+> The **most important part** is, that you NEVER modify already existing files that have been published in any stable
+> version before. These migrations will be hashed at startup and compared to the already applied hashes, and they will
+> `panic` if there is any difference, to never operate on a database in an unknown state.
+
+### I18n
+
+#### UI
+
+Internalization has been reworked a few times until it reached the current, hopefully easy to understand state:
```
lvaroqui, line 262:
```suggestion
Internationalization has been reworked a few times until it reached the current, hopefully easy to understand state:
```

lvaroqui (approved):
Wow! I did not expect such fast reaction to my thread, this is really really great and I think new comers will be grateful!

I tested setting up the dev env from a fresh repo and it looks fine!

 Quick neat-pick there is still a part referring to generated file around line 100:
```
The pre-built files checked into version control are usually only updated during releases to not have
```

sebadob:
Thanks for the feedback. This is always welcome and helps a lot.

PR #947, opened 2025-05-17
Improve dev env commands

lvaroqui:
As discussed in #944, here are:
- the `watch` command (based on [watchexec](https://watchexec.github.io/))
- renamed `backend-*` commands to `dev-env-*`


sebadob:
Thanks!

PR #995, opened 2025-06-07
fix windows build, add cross-env to package.json

cocoon:
Because of the addition of setting env vars in front of the build command, it broke compiling under Windows.

https://www.npmjs.com/package/cross-env?activeTab=readme

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,8 +3,8 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "dev": "DEV_MODE=true vite dev",
-    "build": "DEV_MODE=false vite build",
+    "dev": "cross-env DEV_MODE=true vite dev",
+    "build": "cross-env DEV_MODE=false vite build",
     "preview": "vite preview",
     "check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
     "check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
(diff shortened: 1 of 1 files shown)


sebadob:
Thanks. I don't really like adding dependencies just to fix Windows, but on the long run, we can have another look if maybe Svelte 5 provides a native way how to check this, so we could drop the env var. I had a solution for this with Svelte 4, that unfortunately doesn't work anymore since Svelte 5.

I do have one, as long as the app is running, but we also need one for the `svelte.config.js` (or maybe not anymore?).
I avoid Javascript dependencies as much as possible, because they break so often.

But this could be fixed in a future PR.

sebadob:
Btw, when you are testing on Windows right now, does `hiqlite` work for you with the new `hiqlite-wal` (used by default)? I was not sure about the memory mapping, if the `advise` could throw errors on Windows.

If Rauthy starts, it will be fine, it's used right away.

cocoon:
> I don't really like adding dependencies

yes I totally understand that.

An alternative would be another build command for windows and use "set myenvvar=value"


> memory mapping, if the advise could throw errors on Windows.

:) I already openened a PR @hqlite because of advise etc

sebadob:
fyi I fixed it with #998 . The only situation where the Svelte native approach did not work, was inside the Svelte config itself. But we don't need the information there anymore. The CSP works during local development just fine because of the proxying.

cocoon:
great thanks! 
had still some compiler errors with Windows, will continue tomorrow. one was the fix you already removed the debug line for vars.

PR #1006, opened 2025-06-09
Add missing english translation for scopes deleteDefault

marlyx:
The i18n translation file for the admin UI have one string (deleteDefault) in German. This PR changes it to the equivalent string in English.


sebadob:
Oops, thanks!

PR #1051, opened 2025-06-17
add get config from Vault

cocoon:
So this is the basic first draft, how it currently works.

I will add more information and thoughts the next days.

---

Found this nice guide, that shows what I plan to achieve.
Todo: I will need to test the wrapping:
https://www.linkedin.com/pulse/centralized-secret-management-hashicorp-vault-short-lived-richard-pal-7uktf
https://developer.hashicorp.com/vault/docs/concepts/response-wrapping

Regarding cert auto renewal with vault, I will need to experiment later.

Currently I think it might be just easy to use the coming file system watcher importer together with something like this:
https://discuss.hashicorp.com/t/vault-agent-to-renew-cert/63332/4

pkiCert or secret:
https://developer.hashicorp.com/vault/docs/agent-and-proxy/agent/template#certificates

---

Prepare a config.toml for rauthy.

~~If testing certs, ad these lines:~~
~~cert_path_vault = 'cert.pem'~~
~~key_path_vault = 'cert.key'~~
~~(or key_path_vault = 'cert.key.der')~~

---

For simple testing, one can deploy an Vault Docker container dev env with a preconfigured root token with unlimited lifetime, that just makes it easier to keep commands for testing always the same without needing to find out the current token.
In this example I use a different port (8202) than the default (8200) because I have something already running on 8200.

(These commands are escaped for Windows, so for Linux you might need to set single quotes around the env var values)

`docker run --cap-add=IPC_LOCK -e VAULT_DEV_ROOT_TOKEN_ID=unsafe -e VAULT_ADDR=http://0.0.0.0:8202 -e VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:8202 -p 8202:8202 -d --name=dev-vault hashicorp/vault`

Copy your config and maybe certs into the vault (replace <your_container_id> with your id):

```
docker cp /yourpath/to/config.toml <your_container_id>:/config.toml
docker cp /yourpath/to/cert.pem <your_container_id>:/cert.pem
docker cp /yourpath/to/cert.key <your_container_id>:/cert.key
docker cp /yourpath/to/cert.key.der <your_container_id>:/cert.key.der
```

Open a shell (or run every command invididually):

```
docker exec -it container_id sh
```

Login with the specified root token and create the secrets using the copied files:
```
vault login unsafe
vault status
```
Sealed should be automatically false for dev env.

Putting the config.toml as secret to path rauthy_config with key "config":

```
vault kv put /secret/rauthy_config config.toml=@/config.toml 
```

Adding multiple certs as one command, base64 encode the binary DER cert key:

```
base64 -w 0 /cert.key.der | vault kv put /secret/rauthy_certs cert.pem=@/cert.pem cert.key=@/cert.key cert.key.der=-
```

---

Then set env vars for rauthy (or in future using a config.vault containing this info as alternative way, could be used with injecting or  docker secrets ...):
(this is windows cmd ...)

To enable loading from Vault
```
set USE_VAULT_CONFIG=true
```

if using unsafe connection it is required to set dev mode:
```
set DEV_MODE=true
```

either using vault.toml file:

```
[vault]
addr='http://127.0.0.1:8202'
token='unsafe'
mount='secret'
path='rauthy_config'
config_key='config.toml'
kv_version='2'

# You can provide a root certificate bundle, if you
# are running servers / clients Rauthy needs to connect
# to with self-signed certificates.
# The certificates need to be in PEM format.
#
# overwritten by: HTTP_CUST_ROOT_CA_BUNDLE
#root_ca_bundle = """
#-----BEGIN CERTIFICATE-----
#...
#-----END CERTIFICATE-----
#-----BEGIN CERTIFICATE-----
#....
#-----END CERTIFICATE-----
#"""

```

or from env vars (VAULT_PATH_CERTS currently not in use):

```
set VAULT_ADDR=http://127.0.0.1:8202
set VAULT_TOKEN=unsafe
set VAULT_MOUNT=secret
set VAULT_PATH=rauthy_config
set VAULT_CONFIG_KEY=config.toml
set VAULT_KV_VERSION=2
```


cocoon:
Added now some basic setup for a test env and code should compile and work.

Still have some leftovers like debug logging lines etc.
Currently it uses an own new instance of request client, maybe could use the default one etc.

cocoon:
I removed loading certs for now, focus on simply loading the config.

cocoon:
Now it is cleaned up.
Moved nearly everything into central vault_config.rs

As connection is before a config exists, it is not possible to use the central http_client.
So I had to use client = Client::new();

A review would be nice.

sebadob (changes requested):
The general idea looks nice. Only the code needs restructuring.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,349 @@
...
+    ///
+    /// let source = VaultSource::new(
+    ///     "http://127.0.0.1:8200".to_string(),
+    ///     "hvs.EXAMPLE_TOKEN".to_string(),
+    ///     "secret".to_string(),
+    ///     "rauthy_config".to_string(),
+    ///     "rauthy_certs".to_string(),
+    ///     KvVersion::V2,
+    /// );
+    /// ```
```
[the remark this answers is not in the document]
cocoon: done

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// Builds the URL for Vault's KV1/KV2 engine read API.
+    ///
+    /// This function takes the base address of Vault and builds the complete URL
+    /// to access the read API of the KV1 engine with the specified path.
+    ///
+    /// # Returns
+    ///
+    /// * `Result<Url, ConfigError>` - The constructed URL or an error if the address is invalid
+    //fn build_kv_read_url(&self) -> Result<Url, Box<dyn std::error::Error>> {
+    fn build_kv_read_url(&self, path: &str) -> Result<String, Box<dyn std::error::Error>> {
```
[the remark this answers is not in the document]
cocoon: done

src/models/src/rauthy_config.rs
```diff
@@ -40,13 +42,33 @@ pub struct RauthyConfig {
...
+
+    pub async fn build_from_vault(
+        vault_source: &VaultSource,
+        tx_email: mpsc::Sender<EMail>,
+        tx_events: flume::Sender<Event>,
+        tx_events_router: flume::Sender<EventRouterMsg>,
+    ) -> Result<(Self, hiqlite::NodeConfig), Box<dyn Error>> {
+        let (vars, node_config) = Vars::load_from_vault(vault_source).await;
+        Self::build_from_vars(vars,node_config,tx_email,tx_events,tx_events_router).await
+    }
```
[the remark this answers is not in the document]
cocoon: I moved everything out and only added he loading into "Vars load".
Only other changes in rauthy_config.ts are the new public "pub fn t_str" ...

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,349 @@
...
+use reqwest::Client;
+
+static CONFIG: OnceLock<VaultConfig> = OnceLock::new();
+
+#[derive(Debug)]
+pub struct VaultConfig {
+    pub vault_source: VaultSource,
+}
+
+impl VaultConfig {
```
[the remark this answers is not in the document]
cocoon: This gives me the problems.
I only can load a string from the Vault, so I load the whole config toml file as string from the vault.

If I want to parse this in VaultConfig I would need to duplicate the code or like I started split the parsing into separate function.

Currently I have it that it loads the whole config string and either uses the one from the vault OR use the one from the config file, but I don't know how to merge them.
I would expect that the start always would be a let slf = Self::default() in both cases and then maybe process in order and overwrite with configs that exist.

In my case I even don't want to have a local config.toml at all and load everything from the vault.

But maybe you can help here and adjust the code as needed to get parsed vars instead of the string I use?

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+        if let Ok(v) = env::var("VAULT_MOUNT") {
+            existing_config.vault_source.mount = v;
+        }
+        if let Ok(v) = env::var("VAULT_PATH") {
+            existing_config.vault_source.path = v;
+        }
+        if let Ok(v) = env::var("VAULT_PATH_CERTS") {
+            existing_config.vault_source.path_certs = v;
+        }
```
sebadob, line 89: I would change all of these. Remove all the `if let ...`. In case of a vault config, all of these values are mandatory anyway, so there is no need to do optional checks. Instead, add `.expect()` in case of any missing value.

Removing all of this makes cleaner and less code again.

Also, for the KV version parsing, be more explicit. Parse `1` and `2` and `panic` if it's anything else to avoid unexpected behavior if a user configured something different, because currently, you could say `VAULT_KV_VERSION=Batman` and it would still be considered ok.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    async fn load_from_string(config: &String) -> Self {
+        let vault_source = VaultSource::default();
+        let mut slf = Self::new(vault_source);
+
+        // Note: these inner parsers are very verbose, but they allow the upfront memory allocation
+        // and memory fragmentation, after the quite big toml has been freed and the config stays
+        // in static memory.
+        // It will also be easier in the future to maybe add other config sources into the mix
+        // and use something else as default very easily, like e.g. a config fetched from a Vault.
```
sebadob, line 108: Get rid of this comment. The inner part here should not be that verbose. All vault config values are mandatory, so it's the same as above: Don't do optional parsing or check for `Ok(_)`, but instead `.expect()` the values. Everything is mandatory to successfully connect to the vault, so don't even allow any options. Makes less and cleaner code again.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+    ) -> Self {
+        Self {
+            addr,
+            token,
+            mount,
+            path,
+            path_certs,
+            kv_version,
+        }        
+    }
```
sebadob, line 236: Instead of even having a `new()` function, I would probably build the `VaultSource` struct directly. Much more clear, avoids the possibility of gettings arguments in the wrong order, and is again less code to maintain.

Any `new()` function in my eyes is only then helpful, when it does more than just mapping the same values to itself, like e.g. inisitalizing private ones or doing some convertions of the values before initializing `Self`. In any other scenario, building it directly is much less error prone and less code to maintain.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+
+    fn build_kv_read_url(&self, path: &str) -> Result<String, Box<dyn std::error::Error>> {
+        let api_path = self.kv_version.get_api_path(&self.mount,path);
+        let url = format!("{}{}",remove_quotes(&self.addr),api_path);
```
sebadob, line 240: Why are there even quotes in the first place? If I have a `self.addr`, I would never expect it to have quotes. It would be much cleaner to parse the values correctly from the beginning.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+        self.get_secrets(&self.path).await
+    }
+
+    async fn get_certs(&self) -> Result<HashMap<String, serde_json::Value>, Box<dyn std::error::Error>>  {
+        self.get_secrets(&self.path_certs).await
+    }
+
+    async fn get_secrets(&self, path: &str) -> Result<HashMap<String, serde_json::Value>, Box<dyn std::error::Error>>  {
+        let url = self.build_kv_read_url(path)?;
+        let client = Client::new();
```
sebadob, line 257: I know that the global HTTP client does not exist yet, but take a look at how it's being initialized. We want to build this client with a bit more things set, like HTTPS only, set a proper User-Agent for debugging, min TLS version, and so on, just to provide more debug capabilities and security out of the box.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        if response.status().is_success() {
+            let raw_text = &response.text().await.unwrap();        
+            let raw: serde_json::Value = serde_json::from_str(&raw_text)?;
+
+            let json_obj = raw
+            .get("data")
+            .and_then(|x| { if self.kv_version == KvVersion::V2 { x.get("data")} else { Some(x) } } )
+            .and_then(|x| x.as_object())
+            .unwrap();
```
sebadob, line 271: Indenting here is broken. Please use `cargo fmt` or `just fmt` to have consistent formatting.

Something else that throws me off here: Does this mean, that you expect the `RauthyConfig` to be saved a JSON inside the vault?

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+
+            let mut secret = HashMap::new();
+            for (k, v) in json_obj {
+                secret.insert(k.clone(), serde_json::Value::from(v.as_str().unwrap()));
+            }
+
+            return Ok(secret)
+        }
+        
+        Err(Box::from(format!("Failed to fetch secret from Vault (wrong kv version?): {}", response.status())))
```
sebadob, line 281: It's probably very helpful in that case if you add the raw text response body as well. The vault probably tells you way more information.

cocoon: I tried to add the body, but I am not sure if it is safe to always expect a body to exist.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    #[inline(always)]
+    fn get() -> &'static Self {
+        CONFIG.get().unwrap()
+    }
+
+    async fn load(path_config: &str) -> Self {
+        let Ok(config) = fs::read_to_string(path_config).await else {
+            panic!("Cannot read config file from {}", path_config);
```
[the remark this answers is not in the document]
cocoon: removed

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,283 @@
...
+    pub fn new(vault_source: VaultSource) -> Self {
+        Self { vault_source }
+    }
+
+    pub async fn load_vars() -> Result<String, Box<dyn Error>> {
+        let vault_config_file = "vault.toml";
+        let config_key= "config"; // maybe make configurable by env vars
+        let vault_conf = VaultConfig::build(vault_config_file).await?;
+        vault_conf.init_static();
+        let config_from_vault = Self::get().vault_source.get_config().await?;
```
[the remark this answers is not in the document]
cocoon: changed it and returning an error now

src/models/src/vault_config.rs
cocoon: Agree. Added a custom client with the values copied like it currently is the default for rauthy_config.

You could even further lock it down if you like.
What I do in another private project (AWS Lambda function) is the following:

```
let client = reqwest::Client::builder()
    .use_rustls_tls()
    .tls_built_in_root_certs(false)
    .add_root_certificate(cert)

    // Enable hickory dns for dns caching and async dns resolve.
    //Requires feature "hickory-dns" //reqwest = { version = "0.12.15", default-features = false, features = ["blocking", "rustls-tls", "hickory-dns"] }
    //.hickory_dns(true)
    
    // Custom DNS Resolver
    .dns_resolver(GLOBAL_HICKORY_RESOLVER.clone())

    .timeout(Duration::from_secs(5))
    .build()?;
```

I only allow to connect to specific root certs, a bit like cert pinning.
But that would requires to know what certs the http client would need when it connects to an endpoint.

Additionally you could maybe use a custom DNS resolver and then maybe force to only use specific secure DNS.

cocoon: something broke, now it fails. Will have to track down ...
`Cannot read config from Vault. builder error for url (http://127.0.0.1:8202/v1/secret/data/rauthy_config)`

src/models/src/rauthy_config.rs
```diff
@@ -628,12 +630,28 @@ impl Default for Vars {
 
 impl Vars {
     async fn load(path_config: &str) -> (Self, hiqlite::NodeConfig) {
-        let mut slf = Self::default();
-
-        let Ok(config) = fs::read_to_string(path_config).await else {
-            panic!("Cannot read config file from {}", path_config);
+       let (slf, config) = match env::var("VAULT_CONFIG") {
+            Ok(_value) => {
```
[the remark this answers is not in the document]
cocoon: done

src/models/src/vault_config.rs
cocoon: ok works now with either DEV_MODE env var set to true or correct root ca certs added to config.

src/models/src/vault_config.rs
sebadob: They could easily be removed if you would not split into 2 functions, but use the original `t_str()` from `rauthy_config`, which combines toml + env var, but if you like it better that way, its fine for me.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,339 @@
...
+}
+
+impl Default for VaultConfig {
+    fn default() -> Self {
+        VaultConfig {
+            vault_source: VaultSource::default(),
+        }
+    }
+}
+#[derive(Debug, Clone)]
```
sebadob, line 152: Does it need to impl `Clone`? If not, please remove it, the same for `Debug`, if not necessary. Each `derive` macro increases the compilation time.
The same would be the case for the other `struct`s here.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,339 @@
...
+    fn http_client(&self) -> Result<Client, Box<dyn std::error::Error>> {
+        let mut dev_mode = false;
+        if let Ok(v) = env::var("DEV_MODE") {
+            if v == "true" {
+                dev_mode = true;
+            };
+        }
+
+        //if Vars::default() would be public we could use something like:
+        //let http_client_default = RauthyConfig::Vars::default().http_client;
```
sebadob, line 274: Yeah you could use `RauthyConfig::Vars::default().http_client`, but with a very huge overhead, because the `Vars` are really big, so it makes a lot more sense to build a new client. 

You could also remove the min TLS version and hardcode it to `1.2` in this case. You should never ever use any vaul with any version below that anyway. Makes this section a bit smaller. the same is true for all other values. It only makes sense to have them configurable if you do really custom stuff with external clients, but a vault connection should always be secure and not slow at all. So have timeouts hardcoded to 10 seconds will be just fine, and you don't need to care about pool timeout at all, because this Client will be used once for fetching and then dropped anyway.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,339 @@
...
+
+#[derive(Debug)]
+pub struct VaultConfig {
+    vault_source: VaultSource,
+}
+
+impl VaultConfig {
+    pub async fn load_vars() -> Result<String, Box<dyn Error>> {
+        let vault_config_file = "vault.toml";
+        let config_key = "config"; // maybe make configurable by env vars
```
sebadob, line 21: I don't think we need this to be configurable. But I would either change it to `[vault]` to avoid conflicts better, if the values live in a bigger config with other values, or remove the table completely and expect all the values at the root level, which would make everything even simpler.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,339 @@
...
+    fn get_api_path(&self, mount: &str, path: &str) -> String {
+        match self {
+            KvVersion::V1 => format!("/v1/{}/{}", &mount, &path),
+            _ => format!("/v1/{}/data/{}", &mount, &path),
+        }
+    }
+}
+
+// remove only the outer quotes
+fn remove_quotes(s: &str) -> String {
```
sebadob, line 193: If I did not miss it, you removed the usage of this function, so you can remove the function itself as well. It should not be necessary anyway.

src/models/src/vault_config.rs
cocoon: you mean this comment regarding config_key = "config" ?
This key is the key in the vault containing the config. 

image

So we can either force that it must be named config or make it configurable.
Currently I left it fixed, but could easily be made configurable in future if needed.

src/models/src/vault_config.rs
```diff
@@ -0,0 +1,339 @@
...
+        let mut table = config
+            .parse::<toml::Table>()
+            .expect("Cannot parse TOML file");
+
+        slf.parse_vault(&mut table);
+
+        Ok(slf)
+    }
+
+    fn parse_vault(&mut self, table: &mut toml::Table) {
```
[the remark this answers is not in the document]
cocoon: Ok renamed, also renamed the load_vars() to load_config() as I am not loading Vars?

src/models/src/vault_config.rs
cocoon: not anymore, gone

src/models/src/rauthy_config.rs
```diff
@@ -2222,12 +2242,6 @@ impl Vars {
         if self.server.pub_url.is_empty() {
             panic!("Empty `server.pub_url`");
         }
-        if self.server.pub_url.contains("://") {
-            panic!(
-                "The `server.pub_url` must not contain the Scheme. Rauthy builds it automatically \
-                depending on a few other values."
-            );
-        }
```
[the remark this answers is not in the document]
cocoon: yes I used the git button here from time to time to update the changes

src/models/src/vault_config.rs
sebadob: > Additionally I fixed parsing of toml, the value was adding outer quotes with to_string().

That sounds like a bug in the `toml` crate to me. The quotes are usually added, when you display a `String` in `Debug` mode, but `to_string()` is derived from the `Display` impl, which should never add quotes.

I am using the destructuring syntax in the `rauthy_config.rs` like so:

```rust
let Value::String(s) = map.remove(key)? else {...};
```

But your way is fine as well. No need to change it.

sebadob (commented):
Looks good now.

sebadob:
@cocoon would you please do me a favor and update the `CHANGELOG.md` and add an example for the `vault.toml` file?

I added the vault feature in general already in the `UNRELEASED` section -> `Load config from Vault`. There you can find the `TODO` row and it would be nice to have an example for the `toml` just like I added it for the `geolocation` above.
And it would be very nice to create a `vault.toml` file in the root, with empty values of course, and also check this into version control for future reference.
It should have the same format as the current `config.toml`, mentioning default values, if they exist, and overwriting ENV vars.

Even nicer would also be a section in the `book`, but that is mostly copy & paste then if you add an example `vault.toml` and I can add this when I rebuild the docs with the next release.

cocoon:
yes I can do it.
I will need write and test ist with linux commands.

PR #1053, opened 2025-06-18
Feat: OIDC-backed forward auth

### LINKED ISSUE #1052: Forward auth and login
I'm trying to use Rauthy to lock down the Syncthing UI, which doesn't natively support oauth2. My only options are a static credential, or LDAP.

I've got it working with Caddy's `forward_auth` directive, except that I always get an error from Rauthy back:

```json
{"timestamp":1750195474,"error":"Unauthorized","message":"Authorization header missing"}
```

I suppose this makes sense, since my browser doesn't know to add an authorisation header. In most other cases where I've seen this implemented, there's a way to tell the proxy to redirect to the login portal and handle this. I'm not sure if Rauthy can do this, or if I'm holding it wrong?

Configuration wise in Rauthy I have:

```toml
[auth_headers]
enable = true
```

And for Caddy, it looks like:
```caddyfile
sync.example.com {
    forward_auth https://id.example.com {
        header_up Host {upstream_hostport}
        uri /auth/v1/oidc/forward_auth
        copy_headers X-Forwarded-User
    }

    reverse_proxy 127.0.0.1:8384
    tls <redacted>
    encode
```

sebadob:
fixes #1052

This PR will bring an improved version of the already existing `/forward_auth` endpoint.

The current one expects a JWT token to exist in the `Authorization` header. It should only be used as a last resort, because it comes with many sharp edges and issues you can run into, if you don't configure everything very carefully.

This new version of the `forward_auth` will be an addon and not replace the existing one. It works in combination with a 2nd `/callback` endpoint, that you can inject into client apps via your reverse proxy. It should be possible to have a fully secured OIDC login flow with this setup, even with CSRF checks, as long as your proxy is configured correctly.


sebadob:
The nginx example config is working nicely at this point. Traefik is still giving me a hard time with its local configuration.

PR #1101, opened 2025-07-09
Feat: Rauthy PAM module

sebadob:
This PR is in a very early and experimental stage. On it's own, it does not really do that much. It will be used in combination with another project in the future, which has not been open-sourced yet. In combination, both of them should make it possible to have a dedicated PAM module for Rauthy.

After first local testing, Login to a Linux machine with a user managed by Rauthy was successful. Not only via password, but also with a Yubikey. The other PAM module project will be made public as a separate project in the future to avoid licensing issues. But there is still quite a bit of work to do and even way more testing.


sebadob:
Very first successful login without the user pre-existing on the system. Dynamic `useradd` during login (Password + Yubikey) . This test was also done on a RHEL based SELinux system and a module definition to allow the additional tasks for `local_login_t` has been created.

I don't have any AppArmor system available right now, but it's probably even easier to set up than SELinux.

Next step: Find a good, UX friendly way without the need for additional software on the client, to make SSH logins work with MFA.

sebadob:
I created [rauthy-pam-nss](https://github.com/sebadob/rauthy-pam-nss) to supplement this PR.

sebadob:
Some small quality of life things are still missing, but the basic setup works as expected. Ther other small addons will come with dedicated PRs.

Issue #1123, opened 2025-07-20
RS256 requires key modulusLength to be 2048 bits or larger


valpackett:
**Describe the bug**
Trying to get some legacy clients that haven't heard of EdDSA to work with Rauthy…

After inserting extra log statements to work around [the real error not being logged](https://github.com/panva/openid-client/discussions/567), turns out the error was (just as listed in that link) [due to the modulus size being smol](https://github.com/panva/jose/discussions/516).

```
[2025-07-20T21:52:14.436] [INFO] settings - GOT KEY {"jwk":{"kty":"RSA","alg":"RS256","kid":"uWu(..len:24)","n":"OwO..(len:339)","e":"XXXX"},"algorithms":{}}
```

The [spec does say](https://www.rfc-editor.org/rfc/rfc7518#section-3.3) that

> A key of size 2048 bits or larger MUST be used with these algorithms.

but… Rauthy isn't using 1024 bits, so?? Well, decoding JWK and then PEM shows that it's-

<img width="619" height="448" alt="Image" src="https://github.com/user-attachments/assets/ca867801-03f5-42d6-8645-d97ec7f251d4" />

…20**2**8 bits??

**Additional context**

Container: `ghcr.io/sebadob/rauthy:0.31.2`

sebadob:
Oh yeah, you are right. That was actually a typo when creating keys:

https://github.com/sebadob/rauthy/blob/8dd90965f16adaf54aa6530e43303c510e1e778e/src/models/src/entity/jwk.rs#L181

It's super weird, that even building the key worked, because it never should have. I will fix that, thanks!

PR #1124, opened 2025-07-20
Fix: RSA256 JWK key length, 2048 not 2028 (#1123)

valpackett:


sebadob:
Oh, sorry, closing it was absolutely unintentional. Didn't see it.

PR #1131, opened 2025-07-22
Add an MVP Helm chart for using Rauthy in k8s.

wt:
This chart is meant to be a starting point. This chart will fully fire up an instance of Rauthy in a k8s cluster. However, you will need some form of smtp since rauthy crashes without it. This chart assumes the use of mailcrab, which is a test smtp server that receives mail and present a them in a web UI for testing.

The chart does not require a custom values yaml to launch and is configured for testing. Any serious deployment would need to persist some secrets that are generated in this chart.

Also, this chart runs rauthy on http inside the cluster. I am not sure of a generic way to setup TLS for communication between the ingress proxy and the services inside the cluster. If there is a good generic solution that works be default with all ingress setups, I think that would be a obvious next step.

### DIFF
--- a/charts/rauthy/Chart.yaml
+++ b/charts/rauthy/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: rauthy
+description: A Helm chart for Kubernetes
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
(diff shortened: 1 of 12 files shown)


wt:
References #131

sebadob:
Looks quite nice for testing, but a lot of stuff is missing to make it actually useful of course.
However, one very important thing is that any HPA must not exist because of the Raft and how it's built and maintained. You can scale instances up (at a cost, and if configured properly), but not scale them down afterwards again. The autoscaling should absolutely be removed at least.

I would also not care about adding default variables like for password hashing, when they are hardcoded anyway.

sebadob:
Because Raft is complicated: https://raft.github.io. 

Hiqlite can auto-join and create Raft clusters, but it's impossible (in a reasonable way)  to auto-leave clusters because of autoscaling. Fully removing a node from a cluster needs manual interaction. You will run Rauthy either as a single instance, or in a cluster of 3, or maybe 5 is you want even more safety and give up write throughput. But changing the structure needs a configuration change. A Rauthy deployment is stateful and not ephemeral.

PR #1180, opened 2025-08-28
feat(i18n): Adds translations for Norwegian Bokmål

DuncanConroy:
Replaces switch-case with Map-lookup.


DuncanConroy:
Moin!

I had some compilation errors and thus couldn't test it locally. It was when running 
```cargo run
   Compiling rauthy-data v0.32.1 (/home/danielbunte/development/rauthy/src/data)
error: template "html/index.html" not found in directories ["/home/danielbunte/development/rauthy/src/data/../../templates"]
   --> src/data/src/html/templates.rs:266:19
    |
266 | #[template(path = "html/index.html")]
    |                   ^^^^^^^^^^^^^^^^^
```
    
And there were 36 errors like these, mentioning different files in the `src/data/src/html/templates.rs` file.
I'm not familiar with the askama crate. Also, I couldn't find the files mentioned in the derive in this repository.

Hope this works nonetheless :) Svelte did compile without errors.

sebadob:
> I had some compilation errors and thus couldn't test it locally.

see github.com/sebadob/rauthy/blob/main/CONTRIBUTING.md#initial-development-setup

Don't have much time today, but will look at it when I do. Would be nice if you do full testing on your side as well after the initial setup, and also `just pre-pr-checks`.

DuncanConroy:
Thanks for the quick feedback! I overlooked some step or mixed up the order. But now got it running locally.

It needs some more adjustments in the backend, which I wasn't aware of. Will update later today.

sebadob:
I just took a look at your branch, but something is not working right yet. I cannot change the language in the UI via the [LangSelector](https://github.com/sebadob/rauthy/blob/main/frontend/src/lib/LangSelector.svelte). It resets immediately, while the other languages are fine. It does the reload of the page like expected, but does not overwrite the language with `nb`.

Maybe I should add a section for the contributing guide and explain, what you need to do to add a new language after all the latest changes and reworks.

DuncanConroy:
Let me check that again. I tried it and seemed just fine.
Adding new languages involves quite a few changes, true.

sebadob:
> Let me check that again. I tried it and seemed just fine.

I did a fresh checkout of your branch, `just build-ui` and used the static HTML (served by the Rust backend) for testing. Was unable to switch to `nb`, which is a bit weird, because you should not need to adjust the `LangSelector`. I did not have enough time to look into it myself though.

> Adding new languages involves quite a few changes, true.

Yeah in the backend, it's a no brainer - just update the `enum` and the compiler tells you all the places you need to take a look at. In the frontend, it should be pretty straight forward as well. Maybe I do have some more time later on to check why it does not let me change to `nb`.

sebadob:
Sorry, my bad. I guess it was an issue with my browser beforehand. I tested again and it's fine.

PR #1206, opened 2025-09-26
Feature: Helm Chart

Reversing0148:
Hello!

I have been hard at work creating a helm chart to get new folks up and running with rauthy, while also implementing support for advanced features.

It supports easy setup for both standalone and high availability deployments.
You can also configure usage of an external secret with, for production use.

A helm release github action is also available [here](https://github.com/Reversing0148/rauthy-helm), although I excluded it from this pr.

Please have a look and let me know your thoughts.

### DIFF
--- a/charts/rauthy/.helmignore
+++ b/charts/rauthy/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/charts/rauthy/Chart.yaml
+++ b/charts/rauthy/Chart.yaml
@@ -1,0 +1,36 @@
+apiVersion: v2
+name: rauthy
+description: A Helm chart for Rauthy
+type: application
+version: 0.32.3
+appVersion: "0.32.3"
+
+# Chart metadata
+home: https://github.com/sebadob/rauthy
+icon: https://sebadob.github.io/rauthy/rauthy_grey_small.png
+sources:
+  - https://github.com/sebadob/rauthy
+
+# Keywords for chart discovery
+keywords:
+  - authentication
+  - authorization
+  - oidc
+  - openid-connect
+  - sso
+  - single-sign-on
+  - identity
+  - rust
+
+# Maintainer information
+maintainers:
+  - name: Reversing0148
+    url: https://github.com/Reversing0148
+
+# Minimum Kubernetes version requirement
+kubeVersion: ">=1.19.0-0"
+
+# Annotations for additional metadata
+annotations:
+  category: Authentication
+  licenses: Apache-2.0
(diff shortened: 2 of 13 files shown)


Reversing0148:
NOTES.TXT is still a todo

sebadob:
Hey, thanks for your work. I am sick right now, but will take a deeper look as soon as I am better.

Reversing0148:
> Hey, thanks for your work. I am sick right now, but will take a deeper look as soon as I am better.

Get well soon! No rush! Health should be the absolute priority. Thanks for the notice!

sebadob:
I have not tested it on a live cluster yet, but after a review here, it looks really good so far.

A few things that could be improved maybe:

- The `server.trusted_proxies` should maybe be changed to a mandatory value you must set, instead of specifying all private networks as allowed. Rauthy always tries to have as safe as possible defaults and this allows quite a big range. I would prefer the user needing to set this explicitly. An automatic extraction from the cluster during install is probably not really feasible because of the many different possibilities where / how such a range could be specified.
- I would probably choose shorter ENC KEY IDs, something more in the range of 6 or 8 characters is easier to read and identify. Or even use yyyy-mm-dd. I would also rename the value in the chart from `randAlphaNum16` to something like `encKeyId`.
- `webauthn.webauthn` does not really need to be hardcoded to `Rauthy`, as it has such a default already.
- I only created very simple Helm charts for testing in the past. Is there a way to force an odd `replicaCount` directly inside the chart? It is not recommended to use odd replicas, but more like a hard requirement, as even numbers don't make any sense at all and can give a false sense of safety if someone does not know the details of Raft. This should probably be updated in the doc for the value as well.
- A pod affinity should absolutely be added, as this is crucial for proper HA. I currently use:
```yaml
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
                - key: app
                  operator: In
                  values:
                    - rauthy
            topologyKey: "kubernetes.io/hostname"
```
- `resources` are super important, especially because of higher defaults for password  hashing. We at least need to set a minimum that matches Rauthys default config. It then uses ~100MB (with filled caches and so on) when `hiqlite` is enabled and without memory tuning. In addition, it will use `hashing.argon2_m_cost` of 128MB during password hashing. And that even times 2 (`hashing.max_hash_threads`) by default. So, to make sure it can never crash with default settings, I would require at least 256MiB for memory resources. A user can lower this value, but only after manual hasing config and memory tuning via `MALLOC_CONF` env var. There is no way that a default config will run on minimal environments that cannot provide this amount of memory, because it would be OOM killed during password hashing.
Maybe even add an option for the `Values.yaml` to specify the instance size and then set proper values here would be a nice addition. These sizes could kind of be copied from https://sebadob.github.io/rauthy/config/tuning.html#memory-allocator. For small instances, I would only allow a single hash thread and `MALLOC_CONF=abort_conf:true,narenas:1,tcache_max:1024,dirty_decay_ms:1000,muzzy_decay_ms:1000`. This would lower the required minimum memory to ~164MiB without sacrificing security by lowering `hashing.argon2_m_cost`.
- I would change the `PodDisruptionBudget` and enable it by default. Also remove the `minAvailable` and instead always use `maxUnavailable: 1` which is a way safer setting. `minAvailable` of 2 for instance would kill the cluster during restarts if you use 5 replicas.
- We could maybe add a template for providing a manual config, that could be copied from the repo.

Really nice work with the chart so far. Thanks a lot!

Reversing0148:
Thank you for the detailed feedback. Given your feedback I think there is much more room to improve, so I created this discussion #1220, lets continue working together, and discuss a potential future pull request there. 

Edit: sentence reorganized

PR #1221, opened 2025-10-13
Feat: ToS

### LINKED ISSUE #1196: Terms of Service Checkbox for public services
When enabling open registration for public services, there is currently no way to require users to explicitly accept Terms of Service (TOS) before completing registration. This makes it difficult to enforce legal or policy requirements during self-enrollment.

Add support for a mandatory "Accept TOS" checkbox in the registration form.  
- The checkbox must be checked before a user can successfully register.  
- The label text should be configurable by the Rauthy admin.  
- Optionally, allow linking the text to an external Terms of Service page (instead of embedding the full text).  
This would allow public-facing services to meet compliance requirements without additional customization.

- Manually modifying the registration page to include a TOS checkbox (not maintainable and would break with updates).  
- Relying on external documentation instead of enforcing acceptance (not legally sufficient in many cases).  

**Additional context**
This feature would be especially useful for services that allow open, public registration and need to comply with legal requirements by collecting explicit user consent. It should be optional and configurable so that private/internal deployments are not affected.

sebadob:
fixes #1196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,58 @@
 # Changelog
 
+## UNRELEASED
+
+### Changes
+
+#### Terms of Service
+
+It is now possible to add Terms of Service (ToS) to Rauthy. These can be found in the Admin UI under `Config` -> `ToS`.
+I am not a lawyer, but I would say the implementation is legally "safe". If you have an open registration, the latest
+existing ToS is being shown to the user and the registration can only be completed after an accept. The other situation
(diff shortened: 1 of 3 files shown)


sebadob:
Turns out to be way more complex when we want to provide a good UX because of different steps, locations and cache TTLs with password or webauthn. Not making the auth logic much more complicated takes some time.

The main issue are ToS updates when we need to show it directly after a successful login before proceeding.

PR #1223, opened 2025-10-16
Example Config use Default Values

JuliDi:
The default config in the book sometimes does not use the claimed default value in the commented-out line right below the description.

I think it is better to be able to explicitly configure the default value and I usually expect the default values to be used in such an example config.

And I fixed a few typos I spotted.


sebadob:
thanks!

Issue #1224, opened 2025-10-17
Navigation in Admin UI can overflow


sebadob:
The Navigation in the Admin UI can overflow in windows with a very small height.

sebadob:
fixes in #1230

Issue #1229, opened 2025-10-23
Cloudflare Zero Trust is incompatible with Rauthy


joelsa:
**Is your feature request related to a problem? Please describe.**
I've stated it that way around on purpose, because I am certain the issue is on Cloudflare's side, but: Cloudflare Zero Trust is incompatible with Rauthy.

So for me this issue mainly serves as a heads-up/information for other uses, feel free to close it if there is no viable path to mitigate this in Rauthy.

---
Some details: Cloudflare Zero Trust seems unable to handle encrypted tokens and expects unencrypted ones.
When connecting it to Rauthy, the error it gives is:
`Failed to verify oidc token with fresh keys `

Other Users using Authentik (and possibly Keycloak) could mitigate this by disabling encryption: https://github.com/goauthentik/authentik/issues/12228
https://community.cloudflare.com/t/oidc-sso-failed-to-verify-oidc-token-with-fresh-keys/720407

I am not really sure what that option does in Authentik and what tokens are even encrypted vs. signed in JWTs, but it solved the problem for them.

**Describe the solution you'd like**
Ideally, if the issue is indeed on Cloudflare's side, they should implement this correctly and to spec. Otherwise, if the encryption (is it actually encryption?) is just an additional security layer, adding an option to disable it in Rauthy could be a viable path.

**Describe alternatives you've considered**
Like stated above, duplicating the functionality found in Authentik, i.e. disabling something that Cloudflare doesn't support on a per-Client basis.

**Additional context**

<img width="876" height="352" alt="Image" src="https://github.com/user-attachments/assets/8ed05ed9-db72-49d1-9ffd-d8e7e640945a" />

sebadob:
Rauthy does not encrypt tokens at all, because I think that that's a complete waste of resources these days and just a left-over from a time when we used unencrypted connections. These days, you even need to switch some browsers to specifically allow plain HTTP. 

What kind of token algorithm did you use? I suspect, that Rauthys edfault EdDSA is just not compatible with CF. If you did, please have a look with RS256. The token verification error indicates such an issue.

There a lots of clients and apps that only work with RS256 btw.

joelsa:
> What kind of token algorithm did you use? I suspect, that Rauthys edfault EdDSA is just not compatible with CF. If you did, please have a look with RS256. The token verification error indicates such an issue.

Their docs (https://developers.cloudflare.com/cloudflare-one/integrations/identity-providers/generic-oidc/) state this:

[...]

PR #1230, opened 2025-10-27
UI improvements

JuliDi:
As discussed here https://github.com/sebadob/rauthy/discussions/1210 I have drafted a few changes to the UI.
I tried to be very conservative about adding more padding etc. because I really like the rather minimalist design of Rauthy.
So here are also a few screenshots of what it looks like now and I'd like to hear what you think before proceeding with this.

Most notably, I switched the order of label and input because then the validation errors do not "collide" visually with the input description label (also most other websites have this order so I think it is more what users expect).
And I added a small border around the input to make it visible more clearly that it is an input, which I hope improves accessibility for people with vision impairments.

Screenshot_20251027_100048
<br>
Screenshot_20251027_101410
<br>
Screenshot_20251027_104000

***

On the user page I have added an `overflow: auto` to the info div to avoid clipping:

Before
grafik

Now
grafik

### DIFF
--- a/frontend/src/lib/admin/config/PasswordPolicy.svelte
+++ b/frontend/src/lib/admin/config/PasswordPolicy.svelte
@@ -173,11 +173,6 @@
             />
         </div>
 
-        <p>
-            {ta.passwordPolicy.validityNew}<br />
-            {ta.passwordPolicy.resetSet0}
-        </p>
-
         <div class="row">
             <Input
                 typ="number"
@@ -190,6 +185,12 @@
                 width={inputWidth}
             />
         </div>
+
+        <p>
+            {ta.passwordPolicy.validityNew}<br />
+            {ta.passwordPolicy.resetSet0}
+        </p>
+
         <div class="row">
             <Input
                 typ="number"


sebadob:
Thanks!

I am not sure if I like the switched order for the label. Before, it made just more sense to me when label + validation error are actually placed together instead of spaced apart, divided by the input. The red border can stay on errors for sure. The padding is out of place for the label in the suggestion as well. It at least would need to start at the same distance as the error message, more likely at the end of the border radius, and I would like it more towards the actual input, because if you have multiple inputs, you might get confused about where the label belongs otherwise. To make this work in all situations, you would need more margin between them in general, or push the label a tiny bit more toawrds the input, I guess. Especially the "Password" label in the screenshot looks out of place and seems to have more spacing than "E-Mail", which looks weird. The "Token Lifetime" in the 2nd screenshot looks better again, but I would add a `padding-left` of probably something like `.5rem` to have the same indentation as the error message.

All of this is the original reason why I had label + error message below it. To improve readability, I would probably only add a bit more spacing between label and error message in the original layout, but not too much, so they clearly belong together.

In terms of accessibility, the inputs were fine in the independent review, because they are also marked as invalid and screen readers can pick this up, but yeah the bigger, red full border is easier to see for sure.

In terms of spacing between inputs and label -> input -> error, adding a custom attribute is probably a good idea. This is a place where they are stacked on top and it should be clearly visibile where everything belongs, even without the context like "when you look at the top, you see the label is above the input":

grafik

This is also the reason why I currently have some negative marings for labels and error messages.

If we're going with the label on top and full border approach, I would reduce the inner padding for options selectors slighty to match the inputs. They are a bit bigger in your screenshot.

---

While you are on this, could you also add an `overflow-y: auto` for the Nav in the Admin UI to fix #1224 ?

sebadob:
Something else I just noticed, it that label, actual input and error message should have the same indentation from the left to actually look nice. This looks a bit off as well in the screenshot. Either have all of them start at the same indentation, or maybe try to find a good middleground that the input value starts at the 2nd char of the label / error message, but this will probably not work with non-monospace fonts anyway.

JuliDi:
Thank you for your feedback @sebadob 
I will look into all your points and see how I can fix them – and then get back to you with more screenshots.

Regarding the fix for #1224 I can integrate that for sure, thanks for the hint.

JuliDi:
> While you are on this, could you also add an `overflow-y: auto` for the Nav in the Admin UI to fix #1224 ?

grafik

Fixed in 3f642352

I had to make a few more changes to prevent having two scrollbars in the navigation (one for the navigation bar and one for the menu elements). It seemed to me that having the bottom controls fixed in place and not be hidden would be more practical than having to scroll down in order to logout.

JuliDi:
> Something else I just noticed, it that label, actual input and error message should have the same indentation from the left to actually look nice.



Changed in dacd36e8
I changed it so that it _roughly_ starts at the end of the border, and label and error align. It's a bit difficult to get perfect with different fonts and font sizes, though :/

JuliDi:
> Especially the "Password" label in the screenshot looks out of place and seems to have more spacing than "E-Mail", which looks weird.



and for the password



Addressed this in 50473b5 by aligning the InputPassword's label with the one in Input and 92b936a by adding the missing invalid class to make the password input have a red border, as well.

JuliDi:
> In terms of spacing between inputs and label -> input -> error, adding a custom attribute is probably a good idea. This is a place where they are stacked on top and it should be clearly visibile where everything belongs, even without the context like "when you look at the top, you see the label is above the input":



This may be a bit excessive padding between the inputs, but I think it's quite clear where everything belongs. Should I reduce the padding a little?

sebadob:
> This may be a bit excessive padding between the inputs, but I think it's quite clear where everything belongs. Should I reduce the padding a little?

Yeah I like how it looks, but with a bit less margin / padding would be nicer. As long as the label text is clearly closer to the input it belongs to than the one above, I would reduce the distance between them.

sebadob:
Just one thing I noticed as well: Could you please get rid of the auto-formatting and don't change files between spaces and tabs (I think that's what happened here), because this procuded a huge amount of changes and it's super hard to see what actually changed and what was switching from spaces to tabs and vice versa.

JuliDi:
Yeah that really is a problem :/
Do you use any formatter that I could install to get the same formatting as you have?
My IDE does auto formatting on save, which does not align well with your formatter.

JuliDi:
I am working on fixing a few other layout issues that I have introduced with these changes, e.g. for the datepicker.

Also I am thinking about how to best unify the CSS. Do you have preferences?
Right now I am mostly copy and pasting styles between input module files. Maybe move all the CSS to global.css? Or somehow reorganize how the input modules are used?

sebadob:
I am using defaults from RustRover / IntelliJ and yes it does auto-format everything on save as well. I never created a specific formatter file for a JS project, would need to look that up later.

sebadob:
> e.g. for the datepicker.

Oh yeah regarding that one, I actually pushed a lot of fixes for the calendar input, because In found some issues in behavior with it. They were mostly in [this push](https://github.com/sebadob/rauthy/pull/1237/commits/c9f8853d6953f0322b28826922ef080b6173e5f1)

> Also I am thinking about how to best unify the CSS. Do you have preferences?

For CSS that should affect all inputs, check out the [global.css](https://github.com/sebadob/rauthy/blob/main/frontend/src/css/global.css), yes. There is already global stuff inside it and the other modules should have had more unique stuff applied to them, because when you start separating them in the global file by type, you can mostly edit in the specific components as well, which is a bit cleaner imho.

JuliDi:
> I am using defaults from RustRover / IntelliJ and yes it does auto-format everything on save as well.

Apparently you can export a `.editorconfig` file in IntelliJ https://www.jetbrains.com/help/idea/editorconfig.html
That _might_ work with my IDE (Zed) then. Or at least I could try running another formatter before committing.

sebadob:
Regarding the formatter, I just saw that my IDE is doing something to the indentation of some older files I used in the current PR. Either they were in a bad format before, or they changed the default formatter, not sure about it. I guess a dedicated one would be an improvement for sure.

sebadob:
> Apparently you can export a .editorconfig file in IntelliJ

I will have a look at this. Tried Zed before and really liked it, but it's still missing too many things for me unfortunately.

JuliDi:
And it does have some weird bugs sometimes, but they are mostly fixed relatively fast. But I have no issue switching to VSCode or an IntelliJ-based IDE for working on this project if that makes things easier.

sebadob:
fyi, I just added a `.editorconfig` file in #1240. This should fix all formatting issues.

JuliDi:
I added it manually to my working copy, but it seems that especially the tabs vs spaces doesn't want to work. This is rather inconsistent across files and I don't know why. I will try to get something to work with the editorconfig, Or just go with Rustrover for the time being...

JuliDi:
Hmm, it Rustrover itself doesn't do the formatting. It is the Svelte Plugin. And the Svelte Plugin for Rustrover does different formatting than the Svelte Plugins for Zed and VSCode. So the .editorconfig probably just doesn't contain the settings from the Rustrover/Intellij Svelte Plugin.

Maybe and external formatter would be good for the project. That could be run locally and enforced via CI.

sebadob:
> And the Svelte Plugin for Rustrover does different formatting than the Svelte Plugins for Zed and VSCode.

Oh yeah, wasn't even aware of this. The main issue is, that I cannot configure the Svelte plugin directly, so yeah, probably prettier is the way to go here.

Edit:

I guess that the Svelte plugin for IDEA could even be different as well, because I saw some re-formatting on my side for old files, which have been created when RustRover didn't even exist. That's super annoying.

Edit 2:

Yeah `prettier` is the way to go. I just need to figure out settings that I like, because the default makes things less readable (which is probably the reason why I disabled it ^^).

sebadob:
Started to work on switching to `prettier` in #1241.

JuliDi:
Great, thank you!

sebadob:
I just merged #1241. The only annoying thing now it that you have quite a few file conflicts. -.-

But, there is the prettier end editor config. And you can either use `npm run format` in the `frontend` dir, or `just fmt`.

sebadob:
Btw can it be the case that you need to do a merge from master to sync your `git`, or why is it still showing a diff here when there is none compared to main after your `npm run format`?

JuliDi:
Yeah I will need to rebase but there is lots of conflicts :/

JuliDi:
Gave rebase a try, now there are fewer conflicts – that's something.

JuliDi:
I will fix these on the weekend so maybe we can get this into the next release along with the workaround from #1229

JuliDi:
There is an issue now with the side-navigation. When I click the "burger menu" button, the sidenav closes and the menu button is gone until I reload the page. I presume this is something that broke during rebase, there was quite a lot of stuff to rebase and I think I will need to check the exact diff now again once all conflicts are resolved.

sebadob:
> There is an issue now with the side-navigation. When I click the "burger menu" button, the sidenav closes and the menu button is gone until I reload the page. I presume this is something that broke during rebase, there was quite a lot of stuff to rebase and I think I will need to check the exact diff now again once all conflicts are resolved.

I just had a quick look over the latest changes and yeah, a few things are gone that I have added / fixed in the last days. That's one of the reasons why I never rebase anymore. I only merge from main and then cherry pick on files with conflicts. I have lost too much code and time over the years by rebasing or force-pushing, but in the end everyone has a preferred method.

JuliDi:
Yeah, not a big fan of rebasing anymore after the last couple of rebase nightmares... Would have been easier here, as well.
But I got it fixed, was just a very small thing. Now it's working again.
I need to check everything once more, but it looks like we don't have any more formatting/whitespace related issues.

sebadob:
I did not check every change, but I saw some fixed with color being not-fixed again. In a few places, the `hsl()` was missing around the `var()`for instance. And The `ToSAccept` was gone in one place I think.

JuliDi:
Argh, sorry about that. I will go over the changes again tomorrow. And maybe just make a new branch off of main and copy all my changes there.

JuliDi:
One thing I noticed is that the `just` commands for the dev setup don't work anymore. The postgres container is created but not removed. Not sure what the issue is.

I was able to fix the `dev-env-rm` command with this, but I am unsure what the issue is. I can create a GH issue for that if you want me to.

```diff
diff --git a/justfile b/justfile
index cb3f6fda..38afbd1f 100644
--- a/justfile
+++ b/justfile
@@ -56,9 +56,10 @@ setup:
 # rm mailcrab and postgres docker containers
 @dev-env-rm: dev-env-stop
     just mailcrab-rm
+    just postgres-rm

     echo "Trying to cleanup orphaned containers"
-    {{ docker }} rm container rauthy || echo ">>> No orphaned 'rauthy' container found"
+    {{ docker }} rm rauthy || echo ">>> No orphaned 'rauthy' container found"
```

Apart from that I did a soft reset to main and manually re-committed all the changes. It does look way better now, and the PR no longer deletes the ToS code.

I think some of the paddings/margins could still be improved, but that might be something to do after merging the layout changes.

sebadob:
> One thing I noticed is that the just commands for the dev setup don't work anymore.

Oh that might need some double checking from my side as well. I updated these and made most of them compatible with apples native `container` runtime, since I partly moved to macOS a few weeks ago.

sebadob:
Yeah there are some smaller issues with the `just` recipes. I will fix them in another PR, no worries. Thanks for the hint!

sebadob:
I checked out your branch to take a look, and these are the issues I found so far:

1. The padding for inputs is a bit less on the bottom than it is on the top. Seems a bit too much on the top.

grafik

2. Something is broken with the pattern validation after input. For instance, the `!` entered in the screenshot does not trigger an "Invalid Input" error. Only after I type the next (allowed) character. It is delayed by 1 input. The behavior is inconsistent, because after an error was triggered once (entering invalid chars, deleting them, and then re-adding them), it seems to catch them just fine directly after the input. 

grafik

  The behavior changes and it behaves differently after mount / for the very first error input.

grafik

3. The delete button for the date input is out of place, It should be inline:

grafik

4. The TZ input does have an underline it should not have:

grafik

5. When I expand the TZ input, the overflow / scrollable option that was there before does not exist anymore and it breaks the whole layout:

grafik

6. The margin for inputs it asymmetrical. It has a lot more on the top than it has on the bottom. This leads to werid spacing e.g. when we have a `<Button>` following some inputs, like for the login page or password change for instance:

grafik

  This is pretty easy to see when we just put a border around the whole input incl label and error to visualize it:

grafik

  This looks werid in quite a few places. I guess it's pretty easily fixed when the margin from the top is cut in half and added to the bottom to center the whole thing.

7. The options selector in the `<SearchBar>` now has a weird background that expands beyond its border:

grafik

8. Something got broken for `<LabeledValue>` with the background not beeing transparent anymore for the value:

grafik

9. When collapsing the nav with the burger icon, it stil leaves like ~3rem of empty space:

grafik

  When it should actually disappear completely to make it work on mobile, like it is now:

grafik

  That's even a bigger problem, when the Nav is collapsed automatically because of media queries while reducing screen width:

grafik

10. The border of the languages selector clashes with the Theme switch icon when the Nav is in the narrow view mode. I would probably just remove the border for the input here (the same for the `<SearchBar>` mentioned above).

grafik

11. The new `2px` border width looks a bit disconnected from the whole rest of the UI, because borders are just `1px` everywhere else. When I reduce it to `1px` for the inputs as well, it looks more uniform.

12. The last thing I noticed, was that the `pre` in `glocal.css` was removed, probably because of the rebase?

JuliDi:
Thanks for the review! I will fix the issues asap. Some seem to stem from clashes with my changes and what has been added on main since I made them. Should be easy to fix.

sebadob:
Sorry, forget about the underline for the TZ input. It should have it. I was confused because it now has a border, which looks very weird in that situation. I would remove it and keep the underline to make it look like a "normal" input. It looks way better without it.

JuliDi:
> 1. The padding for inputs is a bit less on the bottom than it is on the top. Seems a bit too much on the top.

grafik

It is exactly the same (0.25rem = 4px on my screen) for top and bottom. I can decrease the top one if you want.



> 2\. Something is broken with the pattern validation after input.

```diff
--- a/frontend/src/lib/form/Input.svelte
+++ b/frontend/src/lib/form/Input.svelte
@@ -91,9 +91,8 @@
     }

     function oninput(event: Event & { currentTarget: EventTarget & HTMLInputElement }) {
-        if (touched) {
-            isValid();
-        }
+        touched = true;
+        isValid();
         onInput?.();
     }
```

Fixed in 60fb2f82b

> The delete button for the date input is out of place, It should be inline:

Argh, fixed this at least twice already, keeps breaking. Tried again in 08190cf7, hope it works now.

Also I noticed the label is below the Date Input. Fixed the order in ecbb1593.


> 4\. The TZ input does have an underline it should not have:

No idea why that border was in there. Removed in d5a1cdc81

> 5\. When I expand the TZ input, the overflow / scrollable option that was there before does not exist anymore and it breaks the whole layout:

This was due to the default max-height being removed. Re-added in 91046503


Will continue with more fixes in a new comment..

sebadob:
> It is exactly the same (0.25rem = 4px on my screen) for top and bottom. I can decrease the top one if you want.

Oh, no then it's fine, and looked odd probably because of the font or something.

JuliDi:
> 6\. The margin for inputs it asymmetrical.

This does look way less cramped in the current state:
grafik

But it is difficult to fix because part of the margin stems from the label and part is from the input. It might be better to make the container an explicit class and add the margin there. What do you prefer?

> 7\. The options selector in the `<SearchBar>` now has a weird background that expands beyond its border:

I can't seem to get it to look exactly like before, but I attempted a fix in 894e6e88 now it looks like this:

grafik



> 8\. Something got broken for `<LabeledValue>` with the background not beeing transparent anymore for the value:

Fixed in 217f67f5 looks very similar to before now.

grafik


> 10\. The border of the languages selector clashes with the Theme switch icon when the Nav is in the narrow view mode.

This was because I had added the border in the wrong class. It should have gone into the `.btn[data-border='true']` selector in Options.svelte. But I had added it in `.btn`. Then the `borderless` property became useless. Fixed it in 4d8bab03


9 is tough to fix, I will look into that later. Gotta figure out where exactly that broke...
11 and 12 should be easier, I will do that later, as well.

Will let you know when I'm done.

sebadob:
> But it is difficult to fix because part of the margin stems from the label and part is from the input. It might be better to make the container an explicit class and add the margin there. What do you prefer?

That would probably be my go to solution, but it does not need to be 100% exact here, as long as it does not look too out of place by default, it should be fine. When you combine them with other components, you usually need small adjustments between them anyway. It would just be nice to have a good looking `<Button>` below it by default.

> I can't seem to get it to look exactly like before, but I attempted a fix in https://github.com/sebadob/rauthy/commit/894e6e882b4a472d2049255513c9f251e1081206 now it looks like this:

Yeah before, not only the option background was there, but also one for the container it is rendered in within the SearchBar. But it you cannot fix it, no worries, I will update dependencies and recompile the UI anyway before the release.

JuliDi:
> 11\. The new `2px` border width looks a bit disconnected from the whole rest of the UI, because borders are just `1px` everywhere else. When I reduce it to `1px` for the inputs as well, it looks more uniform.

Fixed in abdb3f3


> 12\. The last thing I noticed, was that the `pre` in `glocal.css` was removed, probably because of the rebase?

I guess so, too. Re-added it in fea1568

JuliDi:
> 9\. When collapsing the nav with the burger icon, it stil leaves like ~3rem of empty space:

Fixed in b170de8a

Mobile:
grafik
Hidden:
grafik

JuliDi:
> as long as it does not look too out of place by default, it should be fine.

Then I'd rather leave it as is and only fix CSS where necessary. This PR already makes a lot of changes, so I think even more refactoring would be inappropriate here.

It now looks like this:
grafik

Is that alright?

> But it you cannot fix it, no worries, I will update dependencies and recompile the UI anyway before the release.

That would be great, thank you! I don't quite understand how it was working before, so I cannot seem to make it look like before :/

sebadob:
> Then I'd rather leave it as is and only fix CSS where necessary. This PR already makes a lot of changes, so I think even more refactoring would be inappropriate here.

Yeah sure. If you push your latest changes, I will take another look.

I just stumbled about a bug with the `<TZ>`, because the popup does not follow when the window is scrolled, but I am pretty sure that bug existed before already and I will fix that in another PR.

Edit:

The not following popover may even be a browser bug, or it's intended to work like that, because we are not doing anything custom with JS here.

JuliDi:
Yeah I noticed that as well, not sure why exactly that happens. Might be something that can be fixed with some CSS tweaks, though.

I have pushed all changes so far. Let me know if there is anything else that I need to change or that needs fixing.

There is one more usability thing that I probably should open a new issue for: The multi-item text edits that are used e.g. for the redirect-URLs are a bit annoying sometimes because you can't edit the text after entering it and iirc you cannot select and copy the text. So when you have a typo or need to add a wildcard `*`, you need to enter the entire URL again.
I think something like a table would be more practical, where every entry gets its own line or textedit.

sebadob:
> I have pushed all changes so far. Let me know if there is anything else that I need to change or that needs fixing.

But then I guess something is missing, because the button example with the input margins looks very much like before to me:

grafik

Apart from that, the rest looks good now.

sebadob:
> There is one more usability thing that I probably should open a new issue for: The multi-item text edits that are used e.g. for the redirect-URLs are a bit annoying sometimes because you can't edit the text after entering it and iirc you cannot select and copy the text. 

Yeah that's true, that could be improved. The reason why it behaves like this now is keyboard nagivation. It's not that straight forward to make the edit possible and have a proper keyboard navigation at the same time.

JuliDi:
> > I have pushed all changes so far. Let me know if there is anything else that I need to change or that needs fixing.
> 
> But then I guess something is missing, because the button example with the input margins looks very much like before to me:
> grafik
> 
> Apart from that, the rest looks good now.

That is very odd... Could you give it a try in a different browser? `git status` says everything is pushed and I have checked with both Firefox and Chrome. Not sure where this might come from.
Is there a way to clean the svelte stuff with a just command?

sebadob:
It's browser independent. Can also be nicely seen during login:

grafik

JuliDi:
The login is the same on my side. But the password change form looks as in the screenshot I provided... That is weird. I will fix the login screen and check if I can find anything else.

sebadob:
> Is there a way to clean the svelte stuff with a just command?

What exactly to clean? Caches or the build? I do have caching disabled in all my browsers and I am using the DEV ui on port `5173`. Otherwise, if you are testing via the Rust backend, you need to `just build-ui` of course after changes.

sebadob:
Just saw something else after the label change, but that's easily fixed by having these not side-by-side and instead on top of each other:

grafik

JuliDi:
grafik

Increased the margin in 141db12798
I could add more, but its a thin line between "not cramped" and wasted space here... Let me know what you think.

JuliDi:
Fixed the password policy in 404d63044

grafik

JuliDi:
But I am wondering whether the "valid for days" should be above the "Not one of last recent passwords" because the text above refers to the "Valid for Days".

sebadob:
Yeah looks better now. When the WF finished, I can merge it, if you are fine with it.

sebadob:
> But I am wondering whether the "valid for days" should be above the "Not one of last recent passwords" because the text above refers to the "Valid for Days".

I just saw the same thing. :D Yes

JuliDi:
Fixed in e65dc616bc

grafik

JuliDi:
> Yeah looks better now. When the WF finished, I can merge it, if you are fine with it.

Sure, that would be great!

sebadob:
Thanks for all your work! Looks definitely better now, I like it.

JuliDi:
Glad you like it! Thanks for your assistance!

Issue #1222, opened 2025-10-16
Send E-Mail to all users


sebadob:
It is sometimes very helpful to be able to easily send an E-Mail to all existing users (or maybe just a specific group), for instance when maintenance is scheduled or something like that.

Since Rauthy has all the functionality included already to send mails, it would be a very nice addition to the UI to have the possibility to send out an email to all users from a simple editor.

PR #1236, opened 2025-11-18
Feat: HTML / MD editor

sebadob:
prerequisite for #1222 and addon for #1196

For some different reasons, we want an HTML / MD editor in the Admin UI. It should have preview options, as well as being able to do HTML sanitization (even in the UI).

### DIFF
--- a/frontend/src/lib/admin/api_keys/ApiKeyAddNew.svelte
+++ b/frontend/src/lib/admin/api_keys/ApiKeyAddNew.svelte
@@ -62,8 +62,6 @@
         }
 
         if (doesExpire) {
-            console.log()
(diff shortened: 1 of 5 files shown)


sebadob:
The Markdown Editor is working pretty nicely. Some minor quirks may exist with the helper buttons, if you don't want to type in the formatting directly, but in general it works fine.

There is an additional HTML / Rich-Text editor prepared as well. Most things with it are working. However, it lacks some tools and tweaks to make it really nicely useable and actially better than the MD one. This is a lot of work and for now, it is disabled, and it can be tweaked over time. The basics are working fine though.

PR #1241, opened 2025-11-26
switch UI code style to `prettier`

sebadob:
For a consistent code style for all frontend files between editors and systems, the switch to `prettier` makes sense. 

I will not add ESLint, because I hate it from the bottom of my heart. It slowed me down so badly in the past in projects where I was forced to use it, because of such deep automatic integration, that it even rejected compilation if it was not entirely happy. But especially when testing stuff, you temporarily have code that is not perfectly pretty, which is okay and totally normal during development.

### DIFF
--- a/frontend/.prettierrc
+++ b/frontend/.prettierrc
@@ -18,7 +18,7 @@
     ],
     "quoteProps": "as-needed",
     "semi": true,
-    "singleAttributePerLine": true,
+    "singleAttributePerLine": false,
     "singleQuote": true,
     "svelteAllowShorthand": true,
     "svelteSortOrder": "none",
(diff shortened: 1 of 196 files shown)


sebadob:
So, the prettier plugin for svelte has some bugs and issues I found. There are also quite a few open issues in the repo, some open PRs, but not activity since 10 months. It kind of seems unmaintained, but for now, it at least gives a consistent style between editors.

~~The only issue I was not able to fix was with `src/lib/Template.svelte`. For whatever reason, `prettier` will wipe the whole file on format and leave it blank. So far, this was the only file with such an issue I could find. However, observing the UI for the next time is a good idea. Maybe it produced other issues I missed.~~

Edit: 

Found the reason why `prettier` wiped the `Template` file: it had no HTML content, because it doesn't need any. Somehow this screwed up `prettier`s internals it seems. An empty HTML comment fixed this issue.

PR #1244, opened 2025-11-27
Docs: Improved docker compose config

01e9:
* Don't see the reason for manually creating a network - docker compose already creates a new network and puts all containers in it
* Removed manually set container name - instead use service name/key via `docker compose ... {service}` 
  * Container name prefix can be [customized via `.env`](https://docs.docker.com/compose/how-tos/environment-variables/envvars/#compose_project_name)

----

P.S. I was not able to build the docs (and will not do it - too much time spent, I am not a Rust dev)

> I tried to build but without success:
> 
> * `mdbook` fails:
> 
> ```
> $ mdbook serve -n 0.0.0.0
>  INFO Book building has started
> [2025-11-27T11:34:40Z ERROR mdbook_admonish] Fatal error: Unable to parse the input
> [2025-11-27T11:34:40Z ERROR mdbook_admonish]   - Unable to parse the input
> [2025-11-27T11:34:40Z ERROR mdbook_admonish]   - invalid type: null, expected any valid TOML value at line 1 column 129
>  WARN Error writing the RenderContext to the backend, Broken pipe (os error 32)
> ERROR The "admonish" preprocessor exited unsuccessfully with exit status: 1 status
> ```
> 
> * `just` command doesn't exist

### DIFF
--- a/book/src/getting_started/docker.md
+++ b/book/src/getting_started/docker.md
@@ -47,6 +47,8 @@ volumes:
   data:
 ```
 
+_Optional: Create `.env` file with `COMPOSE_PROJECT_NAME=myapp` for custom container/volume name prefix._
+
 Save this into `docker-compose.yaml` and start with:
 
 ```


sebadob:
> Don't see the reason for manually creating a network - docker compose already creates a new network and puts all containers in it

Was probably there because someone else provided the compose file and it was copy & paste from local dev setup, which works with individual containers.

> P.S. I was not able to build the docs (and will not do it - too much time spent, I am not a Rust dev)

No need to be a Rust dev for this. This is just a default mdbook. There is a job to do this via `just build-docs`. No idea why `admonish` complains in your case, but maybe because of the `_` around the Optional, not sure. Fyi you could also take advantage of the `admonish` preprocessor by doing something like:

<pre><code>
```admonish hint
some optional hint ...
```
</code></pre>

But no worries, I can rebuild them.  Thanks. for the PR.

01e9:
I think you can reproduce the build issues - I did run the commands via rust docker image (exact version):

```bash
docker run --rm -it -u $UID -v $(pwd):/code --workdir=/code rust:1.91 bash
```

sebadob:
> I think you can reproduce the build issues

I will check that, thanks!

Edit:

Oh yeah, the basic `rust` image does not have the proper setup most probably. You would need to `mdbook` CLI locally to do this. I need to do some changes to `ghcr.io/sebadob/rauthy-builder` anyway and can include all the necessary tooling in the next one and maybe even set up a CI job. Thanks!

PR #1278, opened 2025-12-18
feat: add zhhans i18n for admin page

evrins:
add zhhans i18n translate for admin page


sebadob:
Sorry missed this PR.

PR #1328, opened 2026-01-16
Feat: Token Revocation

### LINKED ISSUE #1327: Token Revocation (RFC 7009)
Even though JWT token revocation is a very big mistake, very badly misleading, and does not even exist, because JWT tokens cannot be revoked by design, it will come up as an annoyance in the future probably quite a few more times, that Rauthy does not provide a token revocation endpoint.

Because of this, even though it's pretty stupid, adding a token revocation endpoint is sadly necessary for compatibility for some applications like e.g. Matrix next-gen auth.

sebadob:
fixes #1327

### DIFF
--- a/migrations/hiqlite/24_token_revocation.sql
+++ b/migrations/hiqlite/24_token_revocation.sql
@@ -13,6 +13,12 @@ CREATE TABLE issued_tokens
 CREATE INDEX issued_tokens_jti_index
     ON issued_tokens (jti);
 
+CREATE INDEX issued_tokens_user_id_index
+    ON issued_tokens (user_id);
+
+CREATE INDEX issued_tokens_sid_index
+    ON issued_tokens (sid);
+
 CREATE INDEX issued_tokens_exp_index
     ON issued_tokens (exp);
 
--- a/migrations/postgres/V19__revoked_tokens.sql
+++ b/migrations/postgres/V19__revoked_tokens.sql
@@ -10,6 +10,12 @@ CREATE TABLE issued_tokens
     revoked BOOL
 );
 
+CREATE INDEX issued_tokens_user_id_index
+    ON issued_tokens (user_id);
+
+CREATE INDEX issued_tokens_sid_index
+    ON issued_tokens (sid);
+
 CREATE INDEX issued_tokens_exp_index
     ON issued_tokens (exp);
 


sebadob:
@chbgdn I have not added any tests yet, and the revocation currently only happens on the endpoint directly and not anywhere else like OIDC logout, forced session deletion, and so on, but if you have the time, you could compile a version from this branch and give it a try.

Just be careful to not run it on any production DB yet, because if anything is still missing or not working properly, and the DB migrations need an update, it would error after changing them (without cleaning up the whole DB).

Edit:

What's a bit unclear from the RFC is how to handle public clients. Confidential ones must send `Basic` auth and usually on the `/token` endpoint, public clients are expected to send at least their client id with basic auth even if the password is empty. You said Element sends the bearer token in this header, which is definitely not what the RFC says. However, if this can happen, this needs to be checked somehow as well.

chbgdn:
> I have not added any tests yet, and the revocation currently only happens on the endpoint directly and not anywhere else like OIDC logout, forced session deletion, and so on, but if you have the time, you could compile a version from this branch and give it a try.

Sure, I'll try to setup complitely new test environment for this in few days, right now a huge electricity problems in Ukraine, so sorry for delay.

> What's a bit unclear from the RFC is how to handle public clients. Confidential ones must send `Basic` auth and usually on the `/token` endpoint, public clients are expected to send at least their client id with basic auth even if the password is empty. You said Element sends the bearer token in this header, which is definitely not what the RFC says. However, if this can happen, this needs to be checked somehow as well.

Maybe I misunderstood or explained something wrong, but I meant that Element sends the `access_token` in `Authorization` header to the Matrix server after requesting the `/token` endpoint.
`/token` endpoint payload look like this:
```
grant_type=authorization_code&redirect_uri=https://MY_ELEMENT_DOMAIN/?no_universal_links=true&code=*REDACTED*&code_verifier=*REDACTED*&client_id=dyn$CucGdI7QV8RHqueA
```

sebadob:
> Sure, I'll try to setup complitely new test environment for this in few days, right now a huge electricity problems in Ukrane, so sorry for delay.

No worries, I totally understand. -.-

> Maybe I misunderstood or explained something wrong, but I meant that Element sends the access_token in Authorization header to the Matrix server after requesting the /token endpoint.

Ah, got it. No I meant what it actually sends to the revocation endpoint. But I took the information out of the RFC now. As I said, I am just no sure, how exactly it should behave for public clients, as it's not described at all, or I didn't see it.

I will add some tests as the next step and as soon as they are fine, I "revoke" tokens in situations where it would be expected. Then it needs some real world testing.

chbgdn:
> Ah, got it. No I meant what it actually sends to the revocation endpoint. But I took the information out of the RFC now. As I said, I am just no sure, how exactly it should behave for public clients, as it's not described at all, or I didn't see it.

Ah, I get it. I'll try to catch this from Element Web behavior.

chbgdn:
Here is 2 similar requests to `revocation endpoint`, for access and refresh tokens accordingly.
Payload example:
```
token=*HERE_IS_ACCESS_OR_REFRESH_TOKEN*&token_type_hint=access_token&client_id=*CLIENT_ID*
```
For refresh token: `token_type_hint=refresh_token`

sebadob:
> Here is 2 similar requests to revocation endpoint, for access and refresh tokens accordingly.

Oh the `client_id` as a form param, that's good to know. Also not what the RFC says, but I need to catch it then. 
Thanks!

chbgdn:
There are some problems with compiling from main branch.
```
error[E0716]: temporary value dropped while borrowed
   --> src/service/src/token_set.rs:408:32
    |
408 |                     jti: Some(&secure_random_alnum(8)),
    |                                ^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
...
423 |             };
    |              - temporary value is freed at the end of this statement
...
426 |             JwtToken::build(&kp, &claims)?
    |                                  ------- borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value

For more information about this error, try `rustc --explain E0716`.
error: could not compile `rauthy-service` (lib) due to 1 previous error
```

sebadob:
> There are some problems with compiling from main branch.

That is super weird, because I don't get any issues at all, and the `clippy` CI is fine all the time as well. Are you maybe using an older Rust version?

chbgdn:
> That is super weird, because I don't get any issues at all, and the `clippy` CI is fine all the time as well. Are you maybe using an older Rust version?
> 
> Edit:
> 
> Even more weird, because the code you are referring to has not changed in quite some time.

I just synced my fork and tried to recompile with same rust as before. Maybe I just need to make some sort of `make clean`

sebadob:
Anyway, I just pushed #1342 which should fix it, even though it is super weird.

What rust version are you on?

sebadob:
> Maybe I just need to make some sort of make clean

Btw cargo provides something like this with `cargo clean`.

chbgdn:
> Anyway, I just pushed #1342 which should fix it, even though it is super weird.
> 
> What rust version are you on?

Thank you. I use latest rust from manjaro-testing, but before this PR it was ok for Rauthy. I'll try your fix today later.

chbgdn:
@sebadob Testing this right now and discovered that CORS headers is missing at revocation_endpoint, so Element Web can't revoke token. Also have some problems with refresh_token revocation.
```
{"timestamp":1771465836,"error":"JwtToken","message":"Token is not valid yet"}
```

sebadob:
> {"timestamp":1771465836,"error":"JwtToken","message":"Token is not valid yet"}

That's Element using a refresh token long time before the access token expires. Very bad and wasteful behavior, but this rejection can be disabled:

```toml
[access]
# By default, `refresh_token`s will have an `nbf` claim, making them
# valid at `access_token_lifetime - 60 seconds`. Any usage before
# this time will result in invalidation of not only the token itself,
# but also all other linked sessions and tokens for this user to
# prevent damage in case a client leaked the token by accident.
# However, there are bad / lazy client implementations that do not
# respect either `nbf` in the `refresh_token`, or the `exp` claim in
# `access_token` and will refresh early while the current
# access_token is still valid. This does not only waste resources
# and time, but also makes it possible to have multiple valid
# `access_token`s at the same time for the same session. You should
# only disable the `nbf` claim if you have a good reason to do so.
# If disabled, the `nbf` claim will still exist, but always set to
# *now*.
#
# default: false
# overwritten by: DISABLE_REFRESH_TOKEN_NBF
disable_refresh_token_nbf = false
```

chbgdn:
`disable_refresh_token_nbf = true` helped with refresh_token, endpoint return 200 now, only CORS issue left.

sebadob:
I am super busy this week, but you can add it easily. It's basically a copy & paste. Just take a look at `src/api/src/cors_preflight.rs`  -> `options_token()` for instance.

chbgdn:
I tried to do this yesterday, but I missing something in `src/service/src/oidc/token_revocation.rs`
https://github.com/chbgdn/rauthy/commit/699bfe77f4b8fe6998c6e222c660b6fe19af2b6e

sebadob:
If it's still something about CORS, you probably need to add the CORS headers to the actual response in `src/api/src/oidc.rs:1011`  -> `post_token_revoke()` as well.

PR #1330, opened 2026-01-19
fix: Allow @ character in redirect_uri pattern

FernandezBenjamin:
# Title : fix: Allow @ character in redirect_uri pattern

## Change

Add @ to regex PATTERN_URI in frontend/src/utils/patterns.ts

## Reason

Support URIs with path segments like /@warpgate/api/sso/return

Redirect URI for some apps contains @ like the following example :
https://warpgate.domain.com/@warpgate#/login?next=%2F%40warpgate%2Fadmin

The redirect URI for warpgate is https://warpgate.domain.com/@warpgate/api/sso/return

In Authelia this usage is documented (https://www.authelia.com/integration/openid-connect/clients/warpgate/)

## Other notes

The Cargo.lock file is not modified by me but by the command "just fmt" (just in case you ask yourself why, this is not on purpose 🙂 )

Have a good day !

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4765,6 +4765,7 @@ dependencies = [
  "reqwest",
  "ring",
  "rustls",
+ "serde",
  "serde_json",
  "spow",
  "tikv-jemallocator",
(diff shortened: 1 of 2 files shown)


sebadob:
Thanks, but it needs a lot more changes than only this one. The `RE_URI` in `src/common/src/regex.rs` also needs to change, and then all `code` and `/// Validation: *` comments on request types, where it's used.

Frontend validation is only for better UX. The "real one" happens in the backend.

Zit0X:
I will take a look later then

sebadob:
I will do the missing parts quickly, thanks.

Zit0X:
You are really fast thanks a lot :tada:

Issue #1371, opened 2026-01-28
Misspelled Github (called GitHub)


kerkmann:
I know that's really, reaaaaally picky, but it's called `GitHub` and not `Github` (misspelled in the SSO Provider Template and in the readme/documentation). 😂 

![Image](https://github.com/user-attachments/assets/b86c5548-88f4-4a24-ad5a-965707d62735)

PR #1373, opened 2026-01-28
fix: change misspelled Github to GitHub

kerkmann:
closes: #1371


kerkmann:
I've just used `sed` and `find`, this should be double checked. I'll test it as well when the pipeline went through. :)

kerkmann:
I've tested it on my cluster, the things I've tested:
- The entire SSO Provider flow (register, delete, unlink, link)
- Adding the SSO Provider Template GitHub

From my side it's ready to be merged. :)

PR #1395, opened 2026-02-05
Change start_index to be an optional field

cocoon:
VMware vCenter returns an emtpy list without an start index:
```
{
  "schemas": [
    "urn:com.vmware.vidm.usergroup.scim2.UserListResponse"
  ],
  "totalResults": 0,
  "Resources": [],
  "itemsPerPage": 20
}
```


sebadob:
Thanks. Could you also create a PR and remove the default `maxLength` for the `<InputPassword>`? That was a left-over from a very early time and it should not have a default. That should fix the cutoff issue.

cocoon:
Yes thanks will try to create a pr for it, too 👍

PR #1397, opened 2026-02-05
Fix: Remove default maxLength attribute from InputPassword

### LINKED ISSUE #1394: [v0.34.3] SCIM Bearer token cut off
**Describe the bug**
It looks like the bearer token is cut off after 128 chars in UI and in the final request (I logged the output of self.auth_header()).
Maybe the UI field or db table is not configured to take more?
Mine is ~1280 chars long (vCenter).

**To Reproduce**

1. Go to clients
2. Enable SCIM
3. Set a long token
4. Click Save / or maybe just try to copy out what is in the token field now
5. verify length in clipboard

**Expected behavior**
Full token gets saved / used

**Screenshots**
-

**Additional context**
-

cocoon:
Currently untested, will need to be compiled and checked.
I might not be able to today.

fixes #1394

### DIFF
--- a/frontend/src/lib/form/InputPassword.svelte
+++ b/frontend/src/lib/form/InputPassword.svelte
@@ -18,6 +18,7 @@
         autocomplete = 'current-password',
         placeholder = 'Password',
         disabled = false,
+        maxLength,
         required,
         pattern,
         errMsg,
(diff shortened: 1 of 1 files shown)


sebadob:
That change is a bad idea. Don't remove the whole var, just the default `= 128`, but not the rest.

cocoon:
Seems to be OK, no limit appears for the input field now.

PR #1402, opened 2026-02-06
Include remote ID in update payload

### LINKED ISSUE #1400: [v0.34.3] SCIM update user requires id (vCenter)
**Describe the bug**
The problem is, that vCenter requires the id in the json payload, too.
(not, that the url endpoint already would point to it, but ...)

The qustion is, would the change to add the id break other SCIM updates or not?
make id an option to enable/disable?

**To Reproduce**

**Expected behavior**
SCIM user update to succeed.

**Screenshots**
-

**Additional context**
current payload from rauthy is:

```
{
    "schemas": ["urn:ietf:params:scim:schemas:core:2.0:User"],
    "externalId": "ABcdE456...",
    "userName": "myuser@mydomain.com",
    "name": {
        "familyName": "something",
        "givenName": "something"
    },
    "displayName": "myuser@mydomain.com",
    "preferredLanguage": "de",
    "locale": "de",
    "active": true,
    "emails": [{
            "value": "myuser@mydomain.com",
            "primary": true
        }
    ],
    "roles": [{
            "value": "user"
        }
    ]
}
```

Error then:

```
{
  "schemas": [
    "urn:ietf:params:scim:api:messages:2.0:Error"
  ],
  "status": "400",
  "scimType": "invalidValue",
  "detail": "Errors:\nUser ID mismatch\n"
}
```

vCenter wants:

```
{
    "schemas": ["urn:ietf:params:scim:schemas:core:2.0:User"],
    "id": "123456ab-cdef-....-....-............",
    "externalId": "ABcdE456...",
    "userName": "myuser@mydomain.com",
    "name": {
        "familyName": "something",
        "givenName": "something"
    },
    "displayName": "myuser@mydomain.com",
    "preferredLanguage": "de",
    "locale": "de",
    "active": true,
    "emails": [{
            "value": "myuser@mydomain.com",
            "primary": true
        }
    ],
    "roles": [{
            "value": "user"
        }
    ]
}
```

cocoon:
Add remote ID to update payload for vCenter requirements.

fixes #1400

### DIFF
--- a/src/data/src/entity/clients_scim.rs
+++ b/src/data/src/entity/clients_scim.rs
@@ -1032,6 +1032,10 @@ impl ClientScim {
             update_payload.external_id = None;
         }
 
+        // add remote id, required i.e. by vCenter
+        let remote_id = user_remote.id.clone().unwrap_or_default();
+        update_payload.id = Some(remote_id.to_owned());
+
         let json = serde_json::to_string(&update_payload)?;
         let res = http_client()
             .put(format!(


sebadob:
~~For it to work, you would also need to remove the `#[serde(skip_serializing_if = "Option::is_none")]` for the `ScimUser.id`.~~

Edit:

Ah, forget about it, makes no sense. I did not have a coffee yet.

PR #1403, opened 2026-02-06
SCIM only create user if !unlink_ext_id set by group prefix match result

### LINKED ISSUE #1396: [v0.34.3] SCIM group prefix ignored
**Describe the bug**
I could sync users now, but it synced ALL users, even if I specified a group that is only mapped to 2 user accounts.

**To Reproduce**
1. create a group "vcenter:admins"
2. got to clients and enable scim group sync
3. specify a prefix like: vcenter:
4. click save
5. see all users are synced / in logs it tells how many accounts are to be synced

**Expected behavior**
It should filter users by group mapping to the ones containing the prefix

**Screenshots**
-

**Additional context**
-

cocoon:
Done.
Works like expected.

fixes #1396

### DIFF
--- a/src/data/src/entity/clients_scim.rs
+++ b/src/data/src/entity/clients_scim.rs
@@ -955,7 +955,7 @@ impl ClientScim {
             None => {
                 if !unlink_ext_id {
                     self.create_user(expected_groups, payload, groups_local, groups_remote)
-                    .await?;
+                        .await?;
                 }
             }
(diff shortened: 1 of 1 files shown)


sebadob:
When you fix the code style, I can merge it.

sebadob:
Thanks

cocoon:
I thank you for the great support all the time! 
Currently out, did it with my mobile.

PR #1404, opened 2026-02-07
include "use":"sig" in jwks response

bkw:
While trying to get an [oidc middleware for traefik](https://github.com/sevensolutions/traefik-oidc-auth/) working with rauthy, I noticed it [filters](https://github.com/sevensolutions/traefik-oidc-auth/blob/5bdadd0d213a6b1102de86a4e5d8b15ef7b00636/src/oidc/jwks.go#L186-L200) for keys with `"use": "sig"`, which I've seen [other providers return ](https://www.googleapis.com/oauth2/v3/certs). 

[RFC 7517](https://datatracker.ietf.org/doc/html/rfc7517#section-4.2) says the field is optional, but since I think rauthy is only ever using these keys for signing and at least this one implementation is filtering for it, there should not be much harm in statically adding the value, do you agree?

### DIFF
--- a/src/api_types/src/oidc.rs
+++ b/src/api_types/src/oidc.rs
@@ -359,7 +359,7 @@ pub enum JwkKeyPairType {
 pub struct JWKSPublicKeyCerts {
     pub kty: JwkKeyPairType,
     #[serde(rename = "use")]
-    pub _use: String,
+    pub _use: &'static str,
     pub alg: JwkKeyPairAlg,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub crv: Option<String>, // Ed25519
--- a/src/data/src/entity/jwk.rs
+++ b/src/data/src/entity/jwk.rs
@@ -608,7 +608,7 @@ impl From<JWKSPublicKey> for JWKSPublicKeyCerts {
 
         Self {
             kty,
-            _use: "sig".to_string(),
+            _use: "sig",
             alg,
             crv: pk.crv,
             kid: pk.kid,


sebadob:
I don't get it what other purpose this should solve other than increasing payload size and resource usage for no reason. It's the exact same, static value for everything and therefore completely useless for any filtering.

Does this middleware depend on this value? It's kind of "harmful" in the sense that it means more resources and CPU time without any advantage at all, because all of Rauthys keys will have the exact same purpose.

bkw:
I agree it should not be necessary. I just thought in the spirit of "be conservative in what you send and be liberal in what you receive" it would make sense to label the usage of the keys. The mistake is on the side of the clients, for sure.
But of course it's your choice. Never mind, I opened a PR on the client as well.

sebadob:
If the client depends on it, I mean, yeah we could add this information, but it's actually pretty pointless in case of Rauthy.

In this case, it should be a `&'static str` though to prevent the `String` allocation each time.

sebadob:
For the PR on the traefik side, since it's an optional field, they should not filter for it to be present, but actually filter out any other usage of `sig` if it's present. This would be a much better implementation.

bkw:
> For the PR on the traefik side, since it's an optional field, they should not filter for it to be present, but actually filter out any other usage of `sig` if it's present. This would be a much better implementation.

Like so? [sevensolutions/traefik-oidc-auth@`75c3c19` (#254)](https://github.com/sevensolutions/traefik-oidc-auth/pull/254/commits/75c3c1977ac4f2090d2a840f39c226e911a570c5)

sebadob:
> Like so? [sevensolutions/traefik-oidc-auth@75c3c19 (#254)](https://github.com/sevensolutions/traefik-oidc-auth/pull/254/commits/75c3c1977ac4f2090d2a840f39c226e911a570c5)

Depending on how they deserialize, yes that would be a much better approach for an optional field.

bkw:
> For the PR on the traefik side, since it's an optional field, they should not filter for it to be present, but actually filter out any other usage of `sig` if it's present. This would be a much better implementation.

Apologies for my non-existing rust skills. Does dc3213169541390932e949727fb79ff3d0c9132b look better?

sebadob:
> Apologies for my non-existing rust skills. Does https://github.com/sebadob/rauthy/commit/dc3213169541390932e949727fb79ff3d0c9132b look better?

No worries, that's what reviews are for.

Yes that's much better. Any time you use `String::new(_)`, `.to_string()`, ...,  on something, it does a heap allocation. When values never change and are static, you can avoid this syscall completely by using `&'static str` instead.

sebadob:
Thanks!

bkw:
Thanks for your quick response and explanation! Keep up your most excellent work!

PR #1406, opened 2026-02-09
Update paragraphing on index.html

todrobbins:
I'm not trying to be pedantic, but making the text a little more readable.


sebadob:
Everything in `docs/` is auto-generated and your change will be overwritten with the next docs build. You would need to update `book/src/intro.md`, and probably the `README.md` as well, because the content is basically a copy & paste.

PR #1415, opened 2026-02-16
feat: hierarchical scope matching for URNs

chbgdn:
Adds support for flexible scope matching. This is required for Matrix (MSC3861) to handle dynamic URNs, like `urn:matrix:device:some_id`. It now correctly matches base URNs (e.g., `urn:matrix:device`) with their hierarchical extensions.

This logic is opt-in via the new `[matrix]` config section (`MATRIX_SUPPORT_ENABLE`).

Also improved validation by splitting the CSV scope string, which avoids accidental substring matches (e.g., `api` matching `matrix:api`).

### DIFF
--- a/src/data/src/entity/scopes.rs
+++ b/src/data/src/entity/scopes.rs
@@ -435,6 +435,23 @@ impl Scope {
             && scope != "phone"
             && scope != "profile"
     }
+
+    /// Returns `true` if the `requested` scope matches the `allowed` scope.
+    /// Supports hierarchical matching for URNs if Matrix support is enabled
+    /// (e.g. `...:device` matches `...:device:ID`), as required by Matrix (MSC3861).
+    /// Reference: https://github.com/matrix-org/matrix-spec-proposals/pull/3861
+    #[inline]
+    pub fn matches(allowed: &str, requested: &str, matrix_enabled: bool) -> bool {
+        if allowed == requested {
+            return true;
+        }
+
+        if matrix_enabled && let Some(suffix) = requested.strip_prefix(allowed) {
+            return suffix.starts_with(':');
+        }
+
+        false
+    }
 }
 
 impl From<Scope> for ScopeResponse {
--- a/book/src/config/config.md
+++ b/book/src/config/config.md
@@ -1934,6 +1934,15 @@ level_access = 'modifying'
 # overwritten by: LOG_FMT
 #log_fmt = 'text'
 
+[matrix]
+# Enables specific compatibility support for Matrix MSC3861 (Matrix 2.0 native OIDC).
+# This enables hierarchical scope matching (e.g. 'device' matches 'device:ID').
+# Note: Dynamic Client Registration (DCR) should be enabled for full support.
+#
+# default: false
+# overwritten by: MATRIX_SUPPORT_ENABLE
+#msc3861_enable = false
+
 [mfa]
 # If 'true', MFA for an account must be enabled to access the
 # rauthy admin UI.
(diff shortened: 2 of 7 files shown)


src/data/src/entity/scopes.rs
```diff
@@ -435,6 +435,20 @@ impl Scope {
...
+    /// Returns `true` if the `requested` scope matches the `allowed` scope.
+    /// Supports exact matching and hierarchical matching (prefix followed by a colon).
+    pub fn matches(allowed: &str, requested: &str) -> bool {
+        if allowed == requested {
+            return true;
+        }
+
+        if let Some(suffix) = requested.strip_prefix(allowed) {
+            return suffix.starts_with(':');
+        }
```
sebadob: I would not generally allow this, because I've never seen any app even doing it like Matrix. I don't have a PoC in mind yet, but it might be able to abuse this in some situations.

I would rather prefer a config option for opt-in, and only check it if necessary. The reason is not only potential abuse, but also performance. This is a pretty quick comparison for sure, but it's not only called in a look in the `any()` above, it also does a strip prefix each iteration, which almost all deployments will never need.

When it looks for the `:` as the separator, it is already very specific to matrix. So I would maybe add a Matrix specific config option for opt-in, which could then have a hardcoded check for the prefix and therefore improve the UX. The other solution to make it more generally usable would be something like to allow `*` only at the end of scopes, which will then be seen as a wildcard, like it is the case for e.g. redirect URIs.

Another tiny thing here is that I would add the `#[inline]` annotation. The compiler probably does it anyway already, but just to be sure. This function will be called a few times in a tiny loop during the `any()` check, most often probably somewhere at the end, because the OIDC default scopes will come first. Usually, performance optimizations make sense after you measure them, but this is an easy one. It's also called only at a single place in the code, so an inline would not even increase the binary size. Maybe even reduce it a tiny bit.
The thing about performance is pretty picky, I know, but it's only a single annotation.

sebadob:
> (...) which avoids accidental substring matches.

Thanks. I had that on my TODO already.

When the Matrix flow is fully working it would be very nice if you could add it to the docs about it. Just shortly mention the setup and necessary config options to set. I would create a "Guides" or "Example Configs", section or something like that. I wanted to start this since a long time already, I just did not have the time. Matrix however requires a bit more setup and this would be very helpful for others I guess.

src/data/src/entity/scopes.rs
sebadob: > I actually started with the explicit * wildcard approach, but discarded it because it polluted the Discovery document with non-standard scope names (clients seeing device:* instead of device).

Yeah I see. The issue with wildcards is that they would need to be sanitized, and e.g. must not be added to defaults and things like that, right. Then I would probably prefer a dedicated Matrix option in the config here. The user would not even worry about setting the matrix scopes manually, as the prefix could be checked depending on a `bool`. 

Is their implementation final or is it a draft?

chbgdn: Updated. Added the ENABLE_MATRIX_SUPPORT flag and #[inline] as suggested. Hierarchical matching is now opt-in and won't affect performance for others.

chbgdn:
> When the Matrix flow is fully working it would be very nice if you could add it to the docs about it. Just shortly mention the setup and necessary config options to set. I would create a "Guides" or "Example Configs", section or something like that. I wanted to start this since a long time already, I just did not have the time. Matrix however requires a bit more setup and this would be very helpful for others I guess.

I actually managed to get a successful login with one more small tweak on top of this PR, so the proof of concept is working. I'll start with the guide once I've polished the remaining parts and made sure the whole flow is 100% solid.

PR #1416, opened 2026-02-17
fix: infinite IP ban in dynamic client registration

chbgdn:
The `rate_limit_ip` function was using `device_grant` config instead of `dynamic_clients`.

Since `device_grant` limit is optional, this resulted in `None` being passed as TTL to the cache, causing permanent IP bans if device flows weren't configured.

Switched to `dynamic_clients.rate_limit_sec` and fixed the error message to show the actual retry timestamp.


sebadob:
Thanks!

Issue #1410, opened 2026-02-14
Possibility to have upper case letters for client ID


Johan3F:
**Is your feature request related to a problem? Please describe.**
Client IDs can only be created having lowercase letters. 
In projects like OpenCloud, we need to create clients with some specific ClientID that it's a Camel Cased name. 
It would be great to support these kinds of scenarios in Rauthy

**Describe the solution you'd like**
Client ID to be able to be defined also with upper cased letters

**Describe alternatives you've considered**
Not using OpenCloud? :( 

**Additional context**
It seems to be a quite conscious decision, since the UI is telling me that this is an invalid client ID. If this is the case, could you please elaborate why?

Here's the evidence of the UI:
<img width="826" height="516" alt="Image" src="https://github.com/user-attachments/assets/56d13529-5102-403b-9434-50d55f669d9f" />

sebadob:
> It seems to be a quite conscious decision, since the UI is telling me that this is an invalid client ID. If this is the case, could you please elaborate why?

Yeah, because I never imagined someone would actually need this. There is no big reason why they are not allowed. I just usually avoid uppercase letters when not required for things like IDs, URL params, and so on, because of situations like mistaking an O for a 0, I for an l, and so on. 

I can allow uppercase here as well, no problem.

Johan3F:
I think your way of doing things is generally the way I would prefer as well. I just encountered this issue, cause all my other IDs are actually lowercase 😊 

I'm not sure what to say. I would generally say, keep it constraint, but I want to use OpenCloud... so up to you. I would understand if you prefer keeping it that way, I'll just have to find something else, or wait for Oxicloud to develop their oicd integration 🤷

sebadob:
It's not an issue allowing uppercase here. In the end, the admin is responsible for proper naming. This was necessary for groups and roles as well in the past because of weird client app behavior.

However, the fact that Open Cloud requires a hardcoded client id is concerning. This is a pretty bad design decision. The worst thing I have ever seen in that regard was owncloud, which not only required a hardcoded client id, but also a hardcoded client secret, which was publicly known and the same for everydeployment. Things like these are reason enough for me to never use such applications, even if they would fix it, because at some point they thought that such things would be a good idea or "safe".

I will allow it with the next patch.

Johan3F:
Hmmm I'm not very knowledgeable on this, so I'm now curious what would be a nicer approach for this?

[...]

PR #1418, opened 2026-02-17
allows for upper cased letters on client ID

Johan3F:
Addresses the issue #1410

I've made the unilateral decision to merge ephemeral clients and non ephemeral clients regex. Let me know what you think of it, and I'll re-create it with them being two again.

- Allows clientID to follow the same rules as ephemeral clients with upper cased letters
- Removes the now duplicate client-id-ephemeral regex

### DIFF
--- a/frontend/src/api/types/auth_provider.ts
+++ b/frontend/src/api/types/auth_provider.ts
@@ -58,7 +58,7 @@ export interface ProviderLoginRequest {
     // values for the downstream client
     /// Validation: `email`
     email?: string;
-    /// Validation: PATTERN_CLIENT_ID_EPHEMERAL
+    /// Validation: PATTERN_CLIENT_ID
     client_id: string;
     /// Validation: PATTERN_URI
     redirect_uri: string;
--- a/frontend/src/api/types/authorize.ts
+++ b/frontend/src/api/types/authorize.ts
@@ -7,7 +7,7 @@ export interface LoginRequest {
     /// Validation: max 256
     password?: string;
     pow: string;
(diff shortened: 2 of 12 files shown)


sebadob:
Thanks, looks pretty good so far. If I would not be super busy right now, I would have added it myself already.

While looking at the diff here, it's probably a good idea to remove some special characters from the allowed client ID in the future to not get possible conflicts when it's used as a URL param, but that's not too important.

Johan3F:
Ah nice! :) 
I still have some extra time to deal with it this week. Should we tackle it now? 
I did re-used the old ephemeral regex. If we change that, wouldn't that be a breaking change? 
If yes, then I'll suggest to revert this PR quickly and then I'll create a not unified version of it with the non ephemerals to have a more restricted regex

sebadob:
Making it more restrictive would be a breaking change, yes. That's why I would do it with the next minor version. This change now from you allows additional input, which is not a breaking change, so that's fine for the next patch.

Johan3F:
Quick question. When would we be able to have a patch with this released? Are we talking months, weeks or days?

sebadob:
I guess latest next week. I don't know if I have the time this week. Until then, you can use the current nightly: `ghcr.io/sebadob/rauthy:0.35.0-20260220`

It has a breaking change for the issuer though, if you should run into issues. It changed from `iam.example.com/auth/v1` to `iam.example.com/auth/v1/`

Johan3F:
Perfect! 
Thanks for letting me know. 
I guess I'll be testing with the nightly.

Thanks for your quick answer! :)

PR #1419, opened 2026-02-17
feat: opt-in cleanup for inactive dynamic clients

chbgdn:
As discussed previously (see [here](https://github.com/sebadob/rauthy/pull/1316#issuecomment-3759177068)), this PR implements an opt-in scheduler for cleaning up inactive dynamic client registrations.

Added `DYN_CLIENT_CLEANUP_INACTIVE_DAYS` config (default: `0`, disabled).
When enabled, the `dyn_client_cleanup` scheduler extends its scope to include clients that have `last_used` older than the configured threshold.
The scheduler still cleans up never-used clients as before.

This addresses the concern of database bloat from dynamic clients that are no longer in use, while allowing administrators to explicitly enable and configure this behavior to avoid unexpected deletions.

### DIFF
--- a/book/src/config/config.md
+++ b/book/src/config/config.md
@@ -998,6 +998,16 @@ pg_password = '123SuperSafe'
 # overwritten by: DYN_CLIENT_CLEANUP_MINUTES
 #cleanup_minutes = 60
 
+# Defines the number of days after which an inactive dynamic client will be
+# cleaned up. This applies to clients that have been used at least once but
+# have not been active since the configured amount of days.
+#
+# WARNING: This will permanently delete client registrations.
+#
+# default: 0 (disabled)
+# overwritten by: DYN_CLIENT_CLEANUP_INACTIVE_DAYS
+#cleanup_inactive_days = 0
+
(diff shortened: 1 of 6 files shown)


sebadob:
Thanks!

chbgdn:
The `dyn_client_cleanup` fails because of a `FOREIGN KEY constraint` in the database. It happens when we try to delete a client that still has lingering entries in `issued_tokens` or `refresh_tokens`. 

We should probably either:
1. Add `ON DELETE CASCADE` to the `client_id` foreign keys in the database schema.
2. Or explicitly delete all associated tokens before removing the client row.

sebadob:
This cannot be anything with the tokens tables, because they don't have any FK. I am pretty sure it's the `user_login_states` table. This has the on delete set to `RESTRICT` on purpose. If any login states for a client exist, they must be handled beforehand, and a backchannel logout must be triggered for consistency if it's configured.

When the backchannel logouts were added I forgot to add them to this cleanup scheduler. Check `src/api/src/clients.rs` lines 532 - 546.

PR #1420, opened 2026-02-19
fix: enable CORS for revocation_endpoint

chbgdn:
Add CORS support to the `/oidc/token/revoke` endpoint. This is required for web-based apps like Element Web to properly handle token revocation from the browser.


sebadob:
So the missing piece were the headers on the actual `POST `endpoint?

chbgdn:
Yes, I tested this change before making PR, Element Web now successfully revoke tokens.

PR #1422, opened 2026-02-20
fix: add trailing slash to issuer for better OIDC discovery compatibility

chbgdn:
Rauthy default issuer format (`/auth/v1`) currently breaks OIDC discovery in clients that use standard URL resolution (RFC 3986), such as [`matrix-js-sdk`](https://github.com/matrix-org/matrix-js-sdk/blob/8073f27d98acd385a143a12f1d754d301a2f82f1/src/oidc/discovery.ts#L36).

When the issuer URL lacks a trailing slash, the standard `new URL()` constructor treats the last segment (`v1`) as a filename and strips it when joining the discovery path.

**Example of the issue:**
```javascript
// Current Rauthy default
const issuer = "https://example.com/auth/v1";
new URL(".well-known/openid-configuration", issuer).href;
// Returns: https://example.com/auth/.well-known/openid-configuration
// The 'v1' segment is lost, leading to 404s.
```

By adding a trailing slash to the base issuer, we ensure that the `/v1/` segment is preserved as part of the path. I've also updated the endpoint formatting in `well_known.rs` to prevent double slashes in the generated metadata.

### DIFF
--- a/src/data/src/entity/well_known.rs
+++ b/src/data/src/entity/well_known.rs
@@ -85,19 +85,19 @@ impl WellKnown {
     pub fn new(scopes_supported: Vec<String>) -> Self {
         let issuer = &RauthyConfig::get().issuer;
 
-        let authorization_endpoint = format!("{issuer}/oidc/authorize");
-        let device_authorization_endpoint = format!("{issuer}/oidc/device");
-        let token_endpoint = format!("{issuer}/oidc/token");
-        let introspection_endpoint = format!("{issuer}/oidc/introspect");
-        let revocation_endpoint = format!("{issuer}/oidc/token/revoke");
-        let userinfo_endpoint = format!("{issuer}/oidc/userinfo");
+        let authorization_endpoint = format!("{issuer}oidc/authorize");
+        let device_authorization_endpoint = format!("{issuer}oidc/device");
+        let token_endpoint = format!("{issuer}oidc/token");
+        let introspection_endpoint = format!("{issuer}oidc/introspect");
+        let revocation_endpoint = format!("{issuer}oidc/token/revoke");
+        let userinfo_endpoint = format!("{issuer}oidc/userinfo");
         let registration_endpoint = RauthyConfig::get()
             .vars
             .dynamic_clients
             .enable
-            .then_some(format!("{issuer}/clients_dyn"));
-        let end_session_endpoint = format!("{issuer}/oidc/logout");
-        let jwks_uri = format!("{issuer}/oidc/certs");
+            .then_some(format!("{issuer}clients_dyn"));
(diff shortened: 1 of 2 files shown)


sebadob:
Thanks. 

I will delay the merge a bit though, because this will be a breaking change and it forces me to create a new minor version release.

sebadob:
Had a 2nd thought about it, and yes, this should be normalized, but I'm just wondering in which situation exactly you run into an error here? When the issuer is taken from what kind of resource to build the final URL? Because it should also be possible to successfully look up the `openid-configuration` when you just provide the base URL of Rauthy, like in this case `example.com`, because the `.well-known` exists on the base path as well for compatibility.

chbgdn:
When a user logs in via OIDC, Element Web persists the `issuer` string in the browser's **localStorage** (specifically the `mx_oidc_token_issuer` key).

Later, when the client needs to refresh the session, fetch account settings, or perform logout, it pulls this string to rebuild the link to `openid-configuration`. Without a trailing slash in the stored issuer, the standard `new URL()` constructor strips the `v1` segment (treating it as a filename). This leads the client to look for metadata at `.../auth/.well-known/...` instead of `.../auth/v1/.well-known/...`, resulting in a 404.

sebadob:
> When a user logs in via OIDC, Element Web persists the issuer string in the browser's localStorage (specifically the mx_oidc_token_issuer key).

Thanks! I was afraid it would pull the info out of the Token claims, which no one must ever do to build URLs for security reasons.

> Later, when the client needs to refresh the session, fetch account settings, or perform logout, it pulls this string to rebuild the link to openid-configuration.

Actually, not really the best approach imo. It would be a lot better to store the response from the `openid-configuration` lookup instead, but it is, what it is. :)

chbgdn:
> Actually, not really the best approach imo. It would be a lot better to store the response from the `openid-configuration` lookup instead, but it is, what it is. :)

Exactly. We could definitely add a check in `matrix-js-sdk`, but the reality is that this `new URL` behavior is a very common pitfall in many libraries across different languages, not just JS. 

Fixing it on the server side makes Rauthy more robust and "plug-and-play" for all clients out in the wild, similar to how other major providers handle it. It basically prevents these kinds of downstream interop issues regardless of how the client handles its persistence logic.

sebadob:
The only annoying thing is, that I never had any issues with this so far, and this is potentially a huge breaking change. But a normalization is a good idea in any case.

sebadob:
Yeah, just discovered that this PR  broke tests and some checks in other places.

chbgdn:
> Yeah, just discovered that this PR broke tests and some checks in other places.

I noticed the draft PR, thanks for taking care of the cleanup.

PR #1437, opened 2026-02-20
follow-up for issuer trailing `/` changes

sebadob:
follow up for #1422


chbgdn:
Tested it, authorization works fine now.

sebadob:
> Tested it, authorization works fine now.

Thanks! I do have some things on my TODO to finalize testing before I will merge it. Also, the `rauthy-client` needs an update as well.

sebadob:
Everything seems fine so far. The `rauthy-client` needs an update though to make it possible to have a seamless migration for existing Rauthy instances. It needs to accept both issuer versions for at least 1 release.

PR #1446, opened 2026-02-28
Send HTTP status 302 instead of 401 on login_required

### LINKED ISSUE #1445: Rauthy returns HTTP status 401 instead of 302 on login_required
**Describe the bug**
When an OIDC client sends a request with ``prompt=none`` without an active session, Rauthy returns HTTP status 401.

**Expected behavior**
According to the specs, a 302 is required: https://openid.net/specs/openid-connect-core-1_0-final.html#AuthError

**Additional context**
A 401 will break some clients, as they parse this code as a regular error instead of an AuthError.

Forceu:
Fixes #1445

Please note: This PR was **AI generate**d with Claude AI and is **untested**, as I do not have a Rust compiler available at the moment.

### DIFF
--- a/src/api/src/oidc.rs
+++ b/src/api/src/oidc.rs
@@ -143,17 +143,29 @@ pub async fn get_authorize(
         }
     }
 
-    // check for `prompt=no-prompt`
+    // check for `prompt=none`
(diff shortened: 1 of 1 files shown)


sebadob:
Properly fixed it with #1484

PR #1460, opened 2026-03-12
feat(idp_hint): Adds support to pass idp_hint as a parameter

DuncanConroy:
This enables custom applications to offer their own "Login with `<Providername>`" button outside the standard rauthy UI. The rauthy UI checks the given `idp_hint` against the provider ids and, if it finds a match, automatically triggers the login.

I couldn't finalize the `just pre-pr-checks`, due to issues with `issuer` (I guess your work https://github.com/sebadob/rauthy/pull/1437 is related to this).

I was also not able to run `mdbook serve --open` due to an error about "Invalid TOML file" (which I didn't touch).

I needed this functionality for a website security monitoring tool I'm building ([vioro.io](https://vioro.io)). Longer term, I’d be happy to contribute a "bring‑your‑own frontend" section to the Rauthy docs, showing how to integrate custom UIs with this flow.

### DIFF
--- a/src/api_types/src/oidc.rs
+++ b/src/api_types/src/oidc.rs
@@ -62,9 +62,6 @@ pub struct AuthRequest {
     /// Validation: `[a-zA-Z0-9]`
     #[validate(regex(path = "*RE_ALNUM", code = "[a-zA-Z0-9]"))]
     pub prompt: Option<String>,
-    /// Validation: max length 128
-    #[validate(length(max = 128))]
-    pub idp_hint: Option<String>,
 }
 
 #[inline]


sebadob:
That's a pretty cool idea. Could even be extended to suport `rauthy` as well, and if set it could hide all external IdPs. I don't have a use case for this yet, but might be interesting if you know that some external IdPs will not provide enough values for a specific login or something like that.

sebadob:
Thanks!

sebadob:
Btw forgot to mention this. To build the book, you need to use mdbook `v0.4`. The `admonish` preprocessor was not updated for mdbook `v0.5` yet. They have an open issue about that.

PR #1477, opened 2026-03-25
remove "multilingual" k/v pair from book.toml

stensonb:
afaict, `multilingual = ...` isn't a supported option for mdbook.

### DIFF
--- a/book/book.toml
+++ b/book/book.toml
@@ -1,7 +1,6 @@
 [book]
 authors = ["sebastiandobe@mailbox.org"]
 language = "en"
-multilingual = false
 src = "src"
 title = "Rauthy Documentation"
 


stensonb:
Apologies for the noise...

sebadob:
No it's not, you're right.

To build the book you need to use `mdbook-0.4` right now. The problem is that the `admonish` preprocessor is not updated to mdbook `v0.5` yet. They have an open issue for that: https://github.com/tommilligan/mdbook-admonish/issues/233

stensonb:
I thought I was going crazy.

sebadob:
Yeah I had the same problem when I updated all CLI tools. I will add that to the docs if the update takes a bit longer.

PR #1480, opened 2026-03-26
Set UNIX domain socket permissions, add $CONFIG_FILE variable for custom config.toml path

ungeskriptet:
These changes make it a bit easier to use rauthy on NixOS.


sebadob:
I do have it on the TODO to integrate `clap` and make the binary a CLI. This would then accept a custom config path, which would be a lot cleaner imo compared to an env var. But it could be an addition for now.

I do plan an pushing a bunch of PRs in the near future and hopefully release `v0.35` btw.

Issue #1389, opened 2026-02-04
Adding a trailing slash to the path is return a 404 not found


kerkmann:
**Describe the bug**
I've tried the endpoint `/auth/v1/admin/` and saw that the urls are trailing slash sensitive. `/auth/v1/admin` is working, but `/auth/v1/admin/` isn't working

**To Reproduce**

Add a trailing slash to an known endpoint.

**Expected behavior**

Trailing slash should work and not return a 404. :)

sebadob:
> the urls are trailing slash sensitive

Yes, of course they are. For all the base URLs and routes, it catches all possibilities and redirects properly, but for sub routes, I don't really care too much. I get it for an index page e.g., when some browsers automatically add a trailing `/` and others don't, but for other routes, not so much. You get to them via links and buttons and usually never type them manually.
I mean, if you dial a phone number and add a `0` at the end, you will get someone else than you expected at the other end as well.

This can be added at some point, but I don't consider this a bug or really important. The thing is, because of the way how svelte static routes work, you always would need to redirect from a trailing `/` route, because otherwise the links to all the assets in the HTML would be invalid. You cannot simply serve the document on another layer.

kerkmann:
Yeah ... fair enough, get your point. Can be closed, not so important. :)

sebadob:
solved with #1389

Issue #1392, opened 2026-02-04
Show a configurable hint (allowed characters) for the preferred_username field instead of "invalid input"


kerkmann:
**Is your feature request related to a problem? Please describe.**

Currently the `preferred_username` has a frontend and backend regex pattern. This pattern is used for evaluation only, in the web ui it's only showing "Invalid Input". This is misleading for new people, because do not know the policy the administrator is setting up. For example, I have a policy that only lowercase, numbers, point and dash is allowed. Some users were unable to register because they tried solme uppercase characters. The error message "Invalid Input" is very misleading there. :)

**Describe the solution you'd like**

A custom error message like an "input hint" would be awesome. Something like: "valid characters are: a-z 0-9 _ . -" (of course this should be configurable)

**Describe alternatives you've considered**

Just showing the regex in the webui would be possible, but I think that would confuse some non technical people.

PR #1500, opened 2026-03-30
fix: allow space-delimited prompt parameter and handle 'consent'

mega123-art:
Relaxed prompt validation in AuthRequest to allow spaces and underscores (supporting standard values like select_account and lists).
Updated get_authorize logic to use .contains() for checking login, none, and consent values in the prompt string.
Implemented consent handling by preventing the automatic "immediate refresh" when specified, forcing the user to interact with the login flow.
Fixes issue where clients sending prompt=consent select_account received a BadRequest validation error.

### DIFF
--- a/src/api/src/oidc.rs
+++ b/src/api/src/oidc.rs
@@ -111,7 +111,7 @@ pub async fn get_authorize(
     let mut force_new_session = if params
         .prompt
         .as_ref()
-        .map(|p| p.as_str() == "login")
+        .map(|p| p.contains("login"))
         .unwrap_or(false)
     {
         true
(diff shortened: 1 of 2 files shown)


sebadob:
Thanks!

mega123-art:
@sebadob anything else i can work on like some new feat i would like to contribute

sebadob:
One thing I can think of right now, is that I sometimes saw false-positives for `SuspiciousRequest` events. It checks for all non-expected `*.js` files as well, but sometimes a users' browser may have some cached HTML and request chunks in `/_app/**` that belonged to the old UI after a rolling release. Basically, the `/_app/immutable/**` dir should be whitelisted in the suspicious requests checks, so this never happens. Otherwise, it would block regular users, which should never be a thing.

Or you could fix #1389. This only requires a redirect to `/auth/v1/admin`.

#1392 could also be something, although this is not too obvious if we want to handle it for different languages. An easy solution would be to only do this for english. Not sure if this makes too much sense to have it even translated. Usually, this should not even be that necessary, because user will get live feedback file typing their username, so it should be pretty obvious which characters are not allowed. But maybe you find another nice solution for this.

PR #1501, opened 2026-03-30
Fix 404 on trailing slashes for sub-routes #1389

mega123-art:
This PR fixes the 404 Not Found error when adding a trailing slash to sub-routes (e.g., /auth/v1/admin/). Added a default service to the /v1 scope that redirects trailing slashes to the non-slash version.


sebadob:
Please update the branch (and also your other PR) from main so it only shows the actualy changes and not already merged ones.

Apart from that, the way this route is added would kind of kill the current behavior. It would not work as expected. It should only be done very selectively (for the admin UI route). Otherwise, it would conflict with the current "get all" route fallback which handles static assets and the suspicious requests detection.

When done selectively, you can also remove the path len check, which is actually pretty hard to maintain when you don't know about it, because you would end up with unexpected behavior much likely.

src/bin/src/server.rs
```diff
@@ -681,7 +681,8 @@ fn api_services() -> actix_web::Scope {
                 .service(generic::get_ready)
                 .service(swagger_ui::get_openapi_doc)
                 .service(swagger_ui::get_swagger_ui)
-                .service(html::get_static_assets),
+                .service(html::get_static_assets)
+                .default_service(web::to(generic::redirect_subroutes)),
```
sebadob: As I mentioned already a few days ago, this must not be a default service. This would destroy the current logic, as the `html::get_static_assets` is already the default fallback, which serves static assets and if not found checks for suspicious requests.

The redirect from this PR must be applied very selectively to only the admin UI. In that case, as mentioned already, it also does not need the path length checks and so on.

sebadob (changes requested):

sebadob (approved):

PR #1502, opened 2026-03-30
Whitelist /_app/immutable/ in suspicious request block

mega123-art:
This PR whitelists /_app/immutable/ from the suspicious request check. This avoids false-positives when users have cached HTML that requests old chunks after a rolling release.


sebadob:
Actually, I had a 2nd look at this. The blocked requests to `/_app/**` I saw were not in a place where they should have been. They were to root, not to `/auth/v1/_app/**` where they should have been. This means that they were actually kind of suspicious and NOT false-positives. The UI will never load anything from root as it resolves paths dynamically. This means they were manual requests to some assets, even though I have no idea why. But, this also means we don't need to whitelist them and therefore we don't need this PR, sorry, my bad.

PR #1510, opened 2026-04-03
Podman and SELinux related fixes

kulak:
- fix for podman detection conditional,
- Fixed access denied caused by SELinux on podman with :Z option.  I am not sure of this option impact on docker environment, so it might require additional change.
- used missing docker variable instead of accidental direct docker usage

Most of this was detected on Fedora 43 Workstation with podman and no docker installed.

### DIFF
--- a/justfile
+++ b/justfile
@@ -8,7 +8,7 @@ export USER := `echo "$(id -u):$(id -g)"`
 arch := if arch() == "x86_64" { "amd64" } else { "arm64" }
 is_mac_container := `if test -f /usr/local/bin/container; then echo true; else echo false; fi`
 docker := `which docker || which podman || which container || echo 'no-container-runtime-found'`
-map_docker_user := if docker == "podman" { "" } else { "-u $USER" }
+map_docker_user := if file_name(docker) == "podman" { "" } else { "-u $USER" }
 npm := `echo ${NPM:-npm}`
 cargo_home := `echo ${CARGO_HOME:-$HOME/.cargo}`
 node_image := "node:22"
@@ -380,7 +380,7 @@ build image="ghcr.io/sebadob/rauthy" push="push": build-wasm build-ui
     set -euxo pipefail
 
     # make sure base image is up to date
-    docker pull gcr.io/distroless/cc-debian12:nonroot
+    {{ docker }} pull gcr.io/distroless/cc-debian12:nonroot
 
     mkdir -p out/empty
 
@@ -389,8 +389,8 @@ build image="ghcr.io/sebadob/rauthy" push="push": build-wasm build-ui
     # https://github.com/cross-rs/cross/security/advisories/GHSA-2r9g-5qvw-fgmf
     # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189
     {{ docker }} run \
-        -v {{ cargo_home }}/registry:{{ container_cargo_registry }} \
-        -v {{ invocation_directory() }}/:/work/ \
+        -v {{ cargo_home }}/registry:{{ container_cargo_registry }}:Z \
+        -v {{ invocation_directory() }}/:/work/:Z \
         -w /work \
         {{ map_docker_user }} \
         {{ builder_image }}:{{ builder_tag_date }} \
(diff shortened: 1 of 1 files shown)


sebadob:
Adding `:Z` for volume mounts should not be necessary and can potentially be pretty dangerous. Depending on where you run the dev setup, it could expose sys dirs, or even lock you out of, if they get relabeled badly. If you run from `~/**` you should never need relabeling. Otherwise, if you run from a sys dir or anything that has another label than `unconfined_u`, yes it might be necessary, but then you should not run code from a random repo there, and it would be a way better idea to properly label the project dir itself.

The only situation I can imagine right now, is that you are maybe using a system-wide cargo registry, which has some other sys labels on it. Relabeling this can be a huge issue and it would be better to either have a local registry (and don't use `rust` from the package manager, which is always more up to date), or don't mount it at all, if its on OS level.
Mounting the registry is only an efficiency improvement to speed up builds. It's not mandatory that it exists.

I can use `podman` without any issues on Fedora with SELinux. Maybe check the mounted dirs on your side and make sure they have the correct labels first. You can restore labels with

```bash
restorecon -vR /path/to/the/project
```

sebadob:
If you remove the `:Z` options for the volume mounts, I can merge it because of the `file_name()` fix for your case.

kulak:
:Z request makes sense.  Honestly, I did not have time to troubleshoot :Z option on my side.

justfile
```diff
@@ -375,12 +375,12 @@ build-wasm:
     wasm-pack build -d ../../frontend/src/wasm/md --no-pack --out-name md --features md
 
 # Build the final container image.
-build image="ghcr.io/sebadob/rauthy" push="push": build-wasm build-ui
+build image="zotreg.gnezdovi.com/rauthy/myrauthy" push="push": build-wasm build-ui
```
sebadob: Don't change the default image here. You can actually provide that value as an argument as well.

justfile
```diff
@@ -397,19 +397,7 @@ build image="ghcr.io/sebadob/rauthy" push="push": build-wasm build-ui
...
-    # the container.
-    {{ docker }} run \
-        -v {{ cargo_home }}/registry:{{ container_cargo_registry }} \
-        -v {{ invocation_directory() }}/:/work/ \
-        -w /work \
-        -e JEMALLOC_SYS_WITH_LG_PAGE=16 \
-        {{ map_docker_user }} \
-        {{ builder_image }}:{{ builder_tag_date }} \
-        cargo build --release --target aarch64-unknown-linux-gnu
-    cp target/aarch64-unknown-linux-gnu/release/rauthy out/rauthy_arm64
```
sebadob, line 412: Yeah, don't remove that of course. At least don't push that change.

justfile
```diff
@@ -418,15 +406,15 @@ build image="ghcr.io/sebadob/rauthy" push="push": build-wasm build-ui
...
+            --platform linux/amd64 \
             --push \
             .
     else
         {{ docker }} buildx build \
             --build-arg CREATED="$NOW" \
             --build-arg VERSION="$TAG" \
             -t {{ image }}:$TAG \
-            --platform linux/amd64,linux/arm64 \
+            --platform linux/amd64 \
```
sebadob: The same here, Don't remove arm builds.

sebadob (changes requested):

sebadob:
> :Z request makes sense.

Has potential to be pretty dangerous depending on your environment. It should not be necessary, and if it does, there is something wrong with the labels for your current dir most probably.

justfile
kulak: accident, will fix

justfile
kulak: oh, I used this branch for private build and forgot about it.  Yes, I need to restore the original.

kulak: fixed

justfile
kulak: fixed

kulak: fixed

kulak:
It appears my today's push included more changes that I did privately and forgot about.

I removed all these modifications and kept only the related change which is mostly around podman when docker is not installed and there is no docker alias in the shell.

sebadob:
thanks

PR #1520, opened 2026-04-12
Update access token lifetime validation range to 30 days

parsecgames:
It is inconvenient to login every day for apps that are occasionally used in a hurry. An example is a swimming lap recorder application which is used only once a day and adding a few more taps to login each day make the flow cumbersome. This pull request increases the maximum lifetime to 30 days


sebadob:
Definitely not a desired change. Even the currently allowed range is way to high if you want to be somewhat secure. It just limits the damage that can be done. This change would not be enough btw, because the UI has additional validation.

I consider an access token that is valid for more than 1 hour already as too insecure. They should never be issued for such a long time. That's exactly what refresh tokens are for. You should take a look at them.

PR #1523, opened 2026-04-13
feat: add French (fr) locale support

alisterd51:


src/data/src/language.rs
```diff
@@ -65,6 +67,7 @@ impl From<&str> for Language {
         match value {
             "en" | "en-US" => Self::En,
             "de" | "de-DE" => Self::De,
+            "fr" | "fr-FR" => Self::De,
```
sebadob: Looks like a copy&paste error, and it should match to `Self::Fr` instead.

alisterd51: Thank you, it's fixed.

sebadob:
Thanks!

PR #1524, opened 2026-04-14
Feat: hidden password input

### LINKED ISSUE #1506: Add hidden password field to improve autofill by password managers
Currently, the Bitwarden Android app does not detect the login form if the password input field is not shown yet. On my phone, I have to enter the full email address manually, submit, and only then can I use autofill. I understand that this two-step login is necessary, but it would be more convenient if I could autofill the email, submit, and then autofill the password.
Interestingly, this works as expected when using Chrome on Android, but not when using Firefox.

Make login form detection by autofill applications more reliable, even when the password field is not shown yet. This could be accomplished by adding a hidden password field to the login form. For example, Google and Amazon use this approach on their login forms.

None.

**Additional context**
N/A

sebadob:
should fix #1506


kiorul:
Thank you, it works!

PR #1549, opened 2026-05-07
feat: process nonce within device authorization flow

justincmoy:
This commit:
- Accepts a nonce within the device authorization flow
- Stores the nonce in the cache when the device code is created
- Returns the nonce with the token once the device code is authorized

A nonce is useful in this flow to prevent token replay attacks.

### DIFF
--- a/src/api/src/oidc.rs
+++ b/src/api/src/oidc.rs
@@ -616,7 +616,7 @@ pub async fn post_device_auth(
     }
 
     // we are good - create the code
-    let code = match DeviceAuthCode::new(scopes, client.id, payload.client_secret).await {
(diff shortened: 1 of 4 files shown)


sebadob:
A nonce is technically not part of the RFC in this case, but I agree that it makes sense to optionally accept it. Thanks.

justincmoy:
> A nonce is technically not part of the RFC in this case, but I agree that it makes sense to optionally accept it. Thanks.

Thanks! It makes Rauthy compatible [with openpubkey](https://github.com/openpubkey/openpubkey#how-does-openpubkey-work).

I fixed up the formatting issues.

PR #1559, opened 2026-05-13
fix(ui): URL-encode `client.id` in admin path segments

BorisTyshkevich:
## Summary

The admin UI interpolates `client.id` directly into the path segment of `/auth/v1/clients/<id>...` requests across multiple components. For opaque-string client_ids (`dyn$Abc123`, `rauthy`, etc.) this is fine. For URL-shaped client_ids — which `RE_CLIENT_ID` already allows (`[a-zA-Z0-9,.:/_\-&?=~#!()*+%]{2,256}`) — the slashes and colons leak into the path and produce 404s on every UI action: viewing config, generating/rotating secrets, deleting, uploading branding/logo, fetching theme.

URL-shaped client_ids are useful for OAuth resource servers that want the audience to byte-match the resource URL (`aud = client.id = https://api.example.com/`), or for the MCP ecosystem where the connector resource URL doubles as the client identity.

## Fix

Wrap every path-segment `${client.id}` (and `${clientId}` in `ClientLogo.svelte`) with `encodeURIComponent`. The interpolation as a body/form/query value is unchanged — that's already safe.

## Affected files

- `frontend/src/lib/ClientLogo.svelte`
- `frontend/src/lib/admin/clients/ClientConfig.svelte`
- `frontend/src/lib/admin/clients/ClientDelete.svelte`
- `frontend/src/lib/admin/clients/ClientSecret.svelte`
- `frontend/src/lib/admin/clients/branding/ClientBranding.svelte`

5 files, 11 lines, no behaviour change for any client_id that doesn't contain URL-unsafe characters.

## Test plan

- [x] Verified locally against a static client with `id = "https://otel-rauthy-mcp.demo.altinity.cloud/"` — before the patch every admin-UI action 404s, after the patch each works (Config tab loads, Secret generation succeeds, Branding saves, Delete works).
- [x] Verified opaque-string clients (`dyn$*`, `rauthy`) still work — `encodeURIComponent` is a no-op for them.
- [x] No backend changes; backend's actix-web extractor already URL-decodes the path segment.

### DIFF
--- a/frontend/src/lib/ClientLogo.svelte
+++ b/frontend/src/lib/ClientLogo.svelte
@@ -12,8 +12,8 @@
     let showDefault = $state(false);
     let src = $derived(
         updated
-            ? `/auth/v1/clients/${clientId}/logo?updated=${updated}`
-            : `/auth/v1/clients/${clientId}/logo`,
+            ? `/auth/v1/clients/${encodeURIComponent(clientId)}/logo?updated=${updated}`
+            : `/auth/v1/clients/${encodeURIComponent(clientId)}/logo`,
     );
 </script>
 
--- a/frontend/src/lib/admin/clients/ClientConfig.svelte
+++ b/frontend/src/lib/admin/clients/ClientConfig.svelte
@@ -173,7 +173,7 @@
     });
 
     async function fetchClientDetails() {
-        let res = await fetchGet<ClientResponse>(`/auth/v1/clients/${client.id}`);
+        let res = await fetchGet<ClientResponse>(`/auth/v1/clients/${encodeURIComponent(client.id)}`);
         if (res.body) {
             let resp = res.body;
             // The only real difference here can be the `client.scim`, as all the rest is already included in the
@@ -275,7 +275,7 @@
 </script>
 
 <div class="container">
-    <Form action={`/auth/v1/clients/${client.id}`} {onSubmit}>
+    <Form action={`/auth/v1/clients/${encodeURIComponent(client.id)}`} {onSubmit}>
         <h5>{ta.common.information}</h5>
 
         <LabeledValue label="ID" mono copyToClip={client.id}>
--- a/frontend/src/lib/admin/clients/ClientDelete.svelte
+++ b/frontend/src/lib/admin/clients/ClientDelete.svelte
@@ -23,7 +23,7 @@
         err = '';
         isLoading = true;
 
-        let res = await fetchDelete(`/auth/v1/clients/${client.id}`);
+        let res = await fetchDelete(`/auth/v1/clients/${encodeURIComponent(client.id)}`);
         isLoading = false;
 
         if (res.error) {
--- a/frontend/src/lib/admin/clients/ClientSecret.svelte
+++ b/frontend/src/lib/admin/clients/ClientSecret.svelte
@@ -47,7 +47,7 @@
     });
 
     async function fetchSecret() {
-        let res = await fetchPost<ClientSecretResponse>(`/auth/v1/clients/${client.id}/secret`);
+        let res = await fetchPost<ClientSecretResponse>(`/auth/v1/clients/${encodeURIComponent(client.id)}/secret`);
         if (res.body) {
             if (res.body.secret) {
                 secret = res.body.secret;
@@ -62,7 +62,7 @@
             cache_current_hours: cacheSecret ? Number.parseInt(cacheCurrentHours) : undefined,
         };
         let res = await fetchPut<ClientSecretResponse>(
-            `/auth/v1/clients/${client.id}/secret`,
+            `/auth/v1/clients/${encodeURIComponent(client.id)}/secret`,
             payload,
         );
         if (res.body) {
--- a/frontend/src/lib/admin/clients/branding/ClientBranding.svelte
+++ b/frontend/src/lib/admin/clients/branding/ClientBranding.svelte
@@ -30,8 +30,8 @@
     let theme: ThemeRequestResponse | undefined = $state();
 
     let logoKey = $state(genKey());
-    let logoUrl = $derived(`/auth/v1/clients/${client.id}/logo?${logoKey}`);
-    let url = $derived(`/auth/v1/theme/${client.id}`);
+    let logoUrl = $derived(`/auth/v1/clients/${encodeURIComponent(client.id)}/logo?${logoKey}`);
+    let url = $derived(`/auth/v1/theme/${encodeURIComponent(client.id)}`);
 
     $effect(() => {
         fetchTheme();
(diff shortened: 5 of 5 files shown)


sebadob:
It makes sense that client IDs would require encoding in this case, but I don't really get it why you would want to add such ones. The backend accepts IDs encoded as URLs to make ephemeral clients work. What would make more sense is to restrict the UI client IDs to alnum. 

If you are using the ephemeral clients feature to resolve all the data via a URL, you don't need to add them to Rauthy. You only need to enable the feature in the config. As I am thinking about it, this might even be a possible bug, that a client would be resolved via URL when you add a link as ID. The logic in the code in the `find` for getting client information is doing a lookup when the ID can be parsed as a URL, and checks the DB in all other cases.

I never tried it, but it could be possible that the UI allows adding such clients (when they are encoded like with this PR of course), but then all the information you configure is simply ignored.

For your test setup, did you enable ephemeral clients via config? And if so, is there a specific reason why you wanted to add URL-IDs to the DB?

sebadob:
#1572 restricted the validation pattern for the client ID a lot more again. This is only in the UI of course. It should never be necessary to add URI-based clients via UI or to the DB at all (it's actually very confusing). 

Please check that your case is working just fine with enabled ephemeral clients.

bartekus:
I tried to reproduce this independently on current `main` (`9ba3edda`), fresh prod-mode hiqlite DB, to help answer the open question whether this fix is necessary at all. Summary: URL-shaped client IDs can still get into the DB today, the admin UI is the only thing that breaks on them, and this PR's frontend-only fix is sufficient; but whether such clients *should* exist depends on one more thing I found (ephemeral shadowing, below).

## URL-shaped IDs can still enter the DB after #1572

#1572 restricts the *UI input pattern* only. The API-side `RE_CLIENT_ID` still accepts URIs, so this succeeds on main today:

```
POST /auth/v1/clients  (API key, Clients:create)
{"id": "https://connector.example.com/oauth/client.json", "name": "Claude Connector Static", "confidential": true, ...}
→ 200
```

Same goes for `bootstrap/clients.json`, and any rows created before #1572. So the admin UI can always be confronted with such a client, even if it refuses to create one.

## Route matrix (main @ `9ba3edda`)

Static client with id `https://connector.example.com/oauth/client.json` in the DB; `{id}` sent unencoded (what the UI does today) vs percent-encoded (what this PR makes the UI do):

| Route | unencoded `{id}` | `encodeURIComponent({id})` |
|---|---|---|
| `GET /auth/v1/clients/{id}` | 404 | **200** |
| `PUT /auth/v1/clients/{id}` | 404 | **200** |
| `POST /auth/v1/clients/{id}/secret` | 404 | **200** |
| `PUT /auth/v1/clients/{id}/secret` | 404 | 403 ¹ |
| `GET /auth/v1/clients/{id}/logo` | 404 | 404 ² |
| `DELETE /auth/v1/clients/{id}/logo` | 404 | **200** |
| `GET /auth/v1/theme/{id}` | 404 | 404 ² |
| `DELETE /auth/v1/clients/{id}` | 404 | **200** |

¹ tested with a read-only API key, 403 means the route matched and auth was evaluated, vs the unencoded 404 which never matches a route.
² identical 404 to an alnum-id client without a custom logo/theme, route matches, business-level not-found.

So: every admin route works with the encoded form, actix matches the encoded segment and `web::Path` decodes it back. No server-side change is needed; the breakage is purely the unencoded fetch URLs in the Svelte components, which is exactly what this PR touches. I also grepped `main`'s frontend for further unencoded interpolations of a client id into a URL path and found none beyond the 11 this PR fixes, coverage is complete.

## The "configured info is simply ignored" suspicion is real

@sebadob your hunch from the first comment is confirmed: every flow lookup (`authorize`, `authorization_code`, `refresh_token`, revocation, validation) goes through `Client::find_maybe_ephemeral()`, which, when `ephemeral_clients.enable = true` and the id parses as a URL, goes straight to cache/URL-fetch and never consults the DB:

- ephemeral **disabled**: `GET /oidc/authorize?client_id=https%3A%2F%2Fconnector...` → 200, the static DB row is used.
- ephemeral **enabled**, same request, same DB row present: → 400, `Cannot fetch ephemeral client data from https://connector.example.com/oauth/client.json: ... dns error`.

So with ephemeral clients enabled, a URL-id static client is manageable in the admin UI/API but silently dead in every OIDC flow. With ephemeral disabled, it works as a perfectly normal (e.g. confidential) client, which is the one configuration where statically registering a CIMD-style URL id makes sense, e.g. to pin a connector's client config (secret, scopes, redirect URIs) without letting Rauthy fetch anything remotely.

@BorisTyshkevich for the claude.ai "Advanced settings" setup in #1562, was `ephemeral_clients.enable` on in that deployment? If it was, the static URL-id client can't be what made the flow work (it was being shadowed); if it was off for the claude.ai instance, then the static-registration use-case is real and this PR is what makes such clients manageable.

Either way the UI fix seems correct on its own: the API can hold these IDs, so the UI shouldn't 404 on them. If you'd rather keep URL ids out of the DB entirely, the alternative would be restricting `NewClientRequest`/`UpdateClientRequest` server-side to the #1572 pattern, but that forecloses the pinning use-case above and needs an answer for existing rows, so it felt worth laying out the data before anyone writes more code.

sebadob:
> Either way the UI fix seems correct on its own: the API can hold these IDs, so the UI shouldn't 404 on them. If you'd rather keep URL ids out of the DB entirely, the alternative would be restricting NewClientRequest/UpdateClientRequest server-side to the https://github.com/sebadob/rauthy/pull/1572 pattern, but that forecloses the pinning use-case above and needs an answer for existing rows, so it felt worth laying out the data before anyone writes more code.

Exactly. The UI already restricts it after the other PR. The general `ClientRequest` needs to allow the full URI pattern of course of ephemeral clients, but you are right, for clients, there are the minimal `NewClientRequest` structs. They should limit way more (just like the UI does now). It could be made possible to use a URI as client ID, but it does not make much sense. Especially because it's super confusing and behaves completely different depending on a static config value. In addition, the bootstrapping needs to have the stricter validation as well.

If the original idea was using static clients anyway, there would not have been any need to actually specify a URL as ID I guess.

bartekus:
@sebadob following your note above, I put up the server-side restriction as #1605: a stricter `RE_CLIENT_ID_NEW` (`^[a-zA-Z0-9._\-]{2,256}$`, the same as the UI's `PATTERN_CLIENT_ID_NEW`) on `NewClientRequest`, `UpdateClientRequest` and bootstrapping, with `EphemeralClientRequest` and every flow/lookup path left on the current URI pattern. There is one decision in the PR description worth your eye: whether `UpdateClientRequest` should be restricted too, given existing rows.

@BorisTyshkevich thanks for surfacing this; with the validation tightened server-side, a fresh URI-shaped client can no longer reach the Admin UI, so the encoding change here would only matter for legacy rows.

sebadob:
Closing this now because of no response in a long time, and the PR should not even be necessary in the first place. May be re-opened if necessary.

Issue #1584, opened 2026-05-29
RFC: trusted-local ephemeral admin token for headless provisioning (à la kanidm#1747)


caniko:
### Problem

Declaratively provisioning Rauthy (groups/roles/users/clients — and API keys) needs admin API access. Today the only headless options are **(a)** a long-lived API key (minted manually in the Admin UI, or the single `BOOTSTRAP_API_KEY`) — a *stored long-lived secret* that also *cannot manage api_keys* (those endpoints are `validate_admin_session()`-only), or **(b)** scripting the password login, which mandates a solved proof-of-work plus session/CSRF handling. Both push operators toward storing and distributing long-lived admin secrets — exactly what a secure-by-default IdP wants to avoid.

### Prior art

kanidm worked through this exact tension in [kanidm#1747](https://github.com/kanidm/kanidm/issues/1747) and rejected both "let users set their own secrets" and "trust external (config-management) state," landing instead on: *"a way that the root user can trigger the client to grant it an admin token … when acting as a trusted local user."* `kanidm-provision` uses this today (it runs `kanidmd recover-account` over the local admin socket). The trust anchor is **local host access**, the output is an **ephemeral** credential, and nothing long-lived lives in config.

### Proposal

Add a **trusted-local path to a short-lived admin token** — preferably a CLI subcommand, e.g.:

```
rauthy admin-token --config-file <path> [--ttl 120] [-o json]
```

run as the `rauthy` user with the server's DB + `ENC_KEYS`. It mints a short-TTL admin session/token (audit-logged) that automation captures and uses as the admin credential for the API (including the api_keys endpoints), after which it expires. **No long-lived secret is stored, and no external state is trusted** — a caller that already has DB + `ENC_KEYS` access can do anything regardless, so the token grants *nothing new*; it's just an ergonomic, time-boxed, auditable handle on authority the local operator already holds.

Rauthy today has no such local subcommand (the CLI is `serve` / `generate-*` / `hash-password`) and its unix socket is `0o666` transport-only with no peer trust — so this would be a new, **purely additive** capability that keeps Rauthy the source of truth. An alternative/again-additive transport would be a `SO_PEERCRED`-gated admin UDS where local peers are treated as admin.

Trust & hardening: mintable only with local DB + `ENC_KEYS` access; short TTL; audited as an admin-token mint; local trust legitimately bypasses admin-force-MFA (local access already implies full control).

### Question

[...]

PR #1585, opened 2026-05-31
Add API keys advanced bootstrap JSON

caniko:
This follows the maintainer-accepted scope from #1584: add API keys to Advanced Bootstrapping JSON, without adding an admin-token or trusted-local auth primitive.

### Changes

- Adds `bootstrap/api_keys.json` support during empty production database bootstrap.
- Keeps the existing single-key `bootstrap.api_key` / `BOOTSTRAP_API_KEY_SECRET` compatibility path.
- Supports declared plain and encrypted API-key secrets, stores only the encrypted hash, and zeroizes generated / declared plaintext strings after use.
- Adds an `ApiKey` bootstrap type plus an example-parse validation test.
- Documents short-lived production bootstrap API-key guidance and notes that bootstrap JSON is not live reconciliation.

Refs #1584.

### DIFF
--- a/src/data/src/migration/bootstrap/types.rs
+++ b/src/data/src/migration/bootstrap/types.rs
@@ -19,10 +19,10 @@ use rauthy_api_types::oidc::JwkKeyPairAlg;
 use rauthy_api_types::users::{UserAttrValueRequest, UserValuesRequest};
 use rauthy_common::regex::*;
 use regex::Regex;
-use serde::{Deserialize, Serialize};
+use serde::Deserialize;
 use std::fmt::{Debug, Formatter};
 use std::sync::LazyLock;
-use validator::Validate;
+use validator::{Validate, ValidationError, ValidationErrors};
 
 pub static RE_DB_ID: LazyLock<Regex> = LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9]{24}$").unwrap());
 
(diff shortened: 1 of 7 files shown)


src/data/src/migration/bootstrap/types.rs
```diff
@@ -12,13 +12,15 @@
...
 use crate::language::Language;
+use rauthy_api_types::api_keys::{AccessGroup, AccessRights};
 use rauthy_api_types::clients::ScimClientRequestResponse;
 use rauthy_api_types::cust_validation::*;
 use rauthy_api_types::oidc::JwkKeyPairAlg;
 use rauthy_api_types::users::{UserAttrValueRequest, UserValuesRequest};
 use rauthy_common::regex::*;
 use regex::Regex;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
```
sebadob: `Serialize` should not be necessary for both the `ApiKey` and `ApiKeyAccess`.

sebadob (changes requested):

src/data/src/migration/bootstrap/types.rs
caniko: Good call, I removed the `Serialize` derive/import and switched `ApiKey` to manual validation so `ApiKeyAccess` can stay deserialize-only. `cargo check -p rauthy-data` and the focused bootstrap example test pass.

sebadob (approved):
Looks good now, thanks.

PR #1589, opened 2026-06-02
Feat: configurable residentKey requirement for passkey registration

totoCZ:
Adds WEBAUTHN_RESIDENT_KEY / webauthn.resident_key config option (discouraged | preferred | required, default: discouraged) so operators can enable discoverable credentials for Apple passkey compatibility.

### DIFF
--- a/src/data/src/entity/webauthn.rs
+++ b/src/data/src/entity/webauthn.rs
@@ -941,18 +941,42 @@ pub async fn reg_start(
             let cfg = &RauthyConfig::get().vars.webauthn;
             ccr.public_key.timeout = Some(cfg.req_exp as u32 * 1000);
 
+            let resident_key = match cfg.resident_key.as_str() {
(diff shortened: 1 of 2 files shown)


totoCZ:
I'm not a rust guy so here's a quick fix that should work today.

Next would be some type of option, if this is configured, then don't ask for email etc..

sebadob:
This will definitely not work. It's a huge amount of work and there is a lot more to do it properly. Passkey ID's need to be extracted from existing DB data, the DB needs a migration and additional indexes, the frontend needs to handle it properly, the config + parser needs updates, SSR of the login page needs updates, the account dashboard needs updates, also the "password forgotten" flow in the UI, and also the handling in the backend, and so on.

This is not something small you can just add quickly. I expect probably 1000+ LoC that are actually necessary.

Edit:

It's definitely nothing an AI can just do properly on it's own. There are lots of moving parts involved.

totoCZ:
bla bla bla

PR #1597, opened 2026-06-05
feat(scopes): emit custom attributes at token root per-scope (#1595)

### LINKED ISSUE #1595: Allow attributes to be added at the top level instead of under the "custom" key
I came across this when trying to use rauthy with [Immich](https://immich.app/). Immich documents the process [here](https://docs.immich.app/administration/oauth#enable-oauth), and they also mention a `Role Claim` and a `Storage Label Claim` to automatically assign a user a `user/admin` role and to limit how much they can upload. 

Rauthy only emits custom claims in an object under the "custom" key. Though Immich only checks at the top level. You can see that here:

```ts
// https://github.com/immich-app/immich/blob/9d4a6614b149f23f6a6edb0b78d11ccccfa531f7/server/src/services/auth.service.ts#L618-L621
private getClaim<T>(profile: OAuthProfile, options: ClaimOptions<T>): T {
  const value = profile[options.key as keyof OAuthProfile];
  return options.isValid(value) ? (value as T) : options.default;
}
```
The [OpenID Spec](https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims) doesn't mention any standard for OIDC about where custom keys should be placed. It points to [JWT RFC 7519 Section 4.3](https://datatracker.ietf.org/doc/html/rfc7519#section-4.3), which also makes no specification of where custom keys can go. Putting under "custom" isn't wrong, but apps might as only check the top level.

Allow attributes to be emitted at the top level instead of under custom. 

I believe there are no alternatives for this one. This issue can be resolved in one of 2 ways:
- Ask all software that only checks top-level keys to also allow configuration to look at sub-objects.
- rauthy allows keys to be emitted at top-level.

bartekus:
## Summary

Resolves #1595 (builds on the merged migration #1596).

Adds an opt-in, **per-custom-scope** `claims_at_root` flag that promotes a scope's mapped attributes to the JWT root (flattened) instead of nesting them under `custom`. This is the per-scope design from the #1595 thread, not a global/unconditional switch.

## Collision safety (the core of this PR)

`#[serde(flatten)]` does **not** reject duplicate keys at serialize time; it would emit both and silently shadow. So detection is proactive and runs *before* the token is built: if a root-promoted attribute name collides with a reserved/registered claim (RFC 7519 + OIDC core + Rauthy's own top-level fields), **issuance fails** with a clear, logged error rather than ever shadowing a claim. The reserved list (`RESERVED_ROOT_CLAIMS`) is a first line of defense and is explicitly non-exhaustive: no maintained blacklist.

## What's included

- **Claims**: `custom_flattened` on `JwtAccessClaims` / `JwtIdClaims`; `validate_no_reserved_collision()`.
- **Token build**: per-scope routing to root vs nested `custom`; fail-to-issue on collision. A single token can mix nested and root-level claims.
- **Scopes**: `claims_at_root` threaded through CRUD (hiqlite + postgres) and declarative bootstrap; normalized to `false` for non-custom scopes so a default OIDC scope can never request root emission.
- **API types**: `ScopeRequest` (serde default for back-compat) + `ScopeResponse`.
- **Admin UI**: per-scope toggle on custom-scope edit with a persistent warning (root emission may break on future protocol/feature changes; the admin owns collision correctness), linking the IANA JWT Claims Registry.

## Tests

- **Unit** (`jwt/src/claims.rs`): flat-vs-nested serialization; borrowed round-trip via `from_slice` (mirrors token introspection in `token_info.rs`) proving the greedy flatten map never absorbs reserved claims; collision then fail-to-issue.
- **E2E** (`zza_handler_cust_attrs.rs`): flip a mapped scope to `claims_at_root` and assert the attribute is emitted at the token root and absent from `custom`.

## Notes

- Cross-scope same-attribute promotion is intentionally *not* a collision: user attributes are global (keyed by name), so two scopes promoting the same attribute resolve to the identical value (idempotent); failing it would break legitimate configs. Only reserved-claim collisions fail issuance.
- Real-world motivation: OIDC clients that read custom claims only at the token root (e.g. Immich, the original thread motivation).
- Non-English i18n strings use an English fallback for the two new keys; translations welcome.

Local verification: `cargo check --workspace --tests`, clippy (no new warnings), `cargo fmt --check`, and the jwt unit tests all pass; `svelte-check` is clean for the touched frontend files.

### DIFF
--- a/src/data/src/migration/bootstrap/scopes.rs
+++ b/src/data/src/migration/bootstrap/scopes.rs
@@ -1,4 +1,5 @@
 use crate::database::DB;
+use crate::entity::scopes::Scope as ScopeEntity;
 use crate::entity::user_attr::UserAttrConfigEntity;
 use crate::migration::bootstrap::bootstrap_data;
 use crate::migration::bootstrap::types::Scope;
@@ -38,8 +39,8 @@ VALUES ($1, $2, $3, $4, $5)"#;
 
         // `claims_at_root` is only meaningful for custom scopes; normalize it so a
         // default OIDC scope can never be stored as requesting root emission.
-        let claims_at_root = scope.claims_at_root.unwrap_or(false)
-            && crate::entity::scopes::Scope::is_custom(&scope.name);
+        let claims_at_root =
+            scope.claims_at_root.unwrap_or(false) && ScopeEntity::is_custom(&scope.name);
 
         let id = scope.id.unwrap_or_else(new_store_id);
 
--- a/src/data/src/migration/bootstrap/scopes.rs
+++ b/src/data/src/migration/bootstrap/scopes.rs
@@ -19,8 +19,8 @@ pub async fn bootstrap() -> Result<(), ErrorResponse> {
         .collect::<Vec<_>>();
 
     let sql = r#"
-INSERT INTO scopes (id, name, attr_include_access, attr_include_id)
-VALUES ($1, $2, $3, $4)"#;
+INSERT INTO scopes (id, name, attr_include_access, attr_include_id, claims_at_root)
+VALUES ($1, $2, $3, $4, $5)"#;
 
     for scope in scopes {
         let attr_include_access = build_attrs(
@@ -36,19 +36,36 @@ VALUES ($1, $2, $3, $4)"#;
             "attr_include_id",
         );
 
+        // `claims_at_root` is only meaningful for custom scopes; normalize it so a
+        // default OIDC scope can never be stored as requesting root emission.
+        let claims_at_root = scope.claims_at_root.unwrap_or(false)
+            && crate::entity::scopes::Scope::is_custom(&scope.name);
+
         let id = scope.id.unwrap_or_else(new_store_id);
 
         if is_hiqlite() {
             DB::hql()
                 .execute(
                     sql,
-                    params!(id, scope.name, attr_include_access, attr_include_id),
+                    params!(
+                        id,
+                        scope.name,
+                        attr_include_access,
+                        attr_include_id,
+                        claims_at_root
+                    ),
                 )
                 .await?;
         } else {
             DB::pg_execute(
                 sql,
-                &[&id, &scope.name, &attr_include_access, &attr_include_id],
+                &[
+                    &id,
+                    &scope.name,
+                    &attr_include_access,
+                    &attr_include_id,
+                    &claims_at_root,
+                ],
             )
             .await?;
         }
--- a/src/jwt/src/claims.rs
+++ b/src/jwt/src/claims.rs
@@ -131,8 +131,12 @@ pub struct JwtAccessClaims<'a> {
     pub groups: Option<Vec<&'a str>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub custom: Option<HashMap<String, serde_json::Value>>,
-    // #[serde(flatten, skip_serializing_if = "Option::is_none")]
-    // pub custom_flattened: Option<HashMap<String, serde_json::Value>>,
+    /// Custom user attributes promoted to the token root, driven by the per-scope
+    /// `claims_at_root` flag. Flattened, so each entry becomes a top-level claim
+    /// instead of nesting under `custom`. Issuance MUST fail if any key here
+    /// collides with a reserved claim; see [`validate_no_reserved_collision`].
+    #[serde(flatten, skip_serializing_if = "Option::is_none")]
+    pub custom_flattened: Option<HashMap<String, serde_json::Value>>,
 }
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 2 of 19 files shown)


src/data/src/migration/bootstrap/scopes.rs
```diff
@@ -36,19 +36,36 @@ VALUES ($1, $2, $3, $4)"#;
             "attr_include_id",
         );
 
+        // `claims_at_root` is only meaningful for custom scopes; normalize it so a
+        // default OIDC scope can never be stored as requesting root emission.
+        let claims_at_root = scope.claims_at_root.unwrap_or(false)
+            && crate::entity::scopes::Scope::is_custom(&scope.name);
```
sebadob: We can get rid of the full prefix `crate::entity::scopes::` for `Scope` here.

src/jwt/src/claims.rs
```diff
@@ -174,12 +178,105 @@ pub struct JwtIdClaims<'a> {
...
+/// offending keys so the misconfigured scope can be fixed; the token is never
+/// issued with a shadowed claim.
+pub fn validate_no_reserved_collision(
+    flattened: &HashMap<String, serde_json::Value>,
+) -> Result<(), ErrorResponse> {
+    let mut collisions = flattened
+        .keys()
+        .filter(|k| RESERVED_ROOT_CLAIMS.contains(&k.as_str()))
+        .map(|k| k.as_str())
+        .collect::<Vec<_>>();
```
sebadob, line 273: Not really sure on this one. The current approach is more better for debugging, because you see the complete list of collisions. On the other hand, each single token creation will call this, and you almost never need to debug here.

We would skip the `Vec<_>` allocation when we don't `.collect()`, and exit early on the first collision, then log an error just for that collision. This makes debugging slightly less user friendly, but only during the very first setup, and only if there are already multiple collisions. These would be caught by the blacklist in almost all situations already.

Another idea would be that we only allocate the new `Vec<_>` if we found a collision. The fast-path can be made completely alloc-free here. This would be the best of both worlds I guess. This would also not break the tests you have already written.

sebadob (commented):
Thanks a lot. Apart from 2 tiny comments, the PR is looking really solid. I would have expected more lines necessary to do full testing and everything. That's awesome.

src/data/src/migration/bootstrap/scopes.rs
bartekus: Done. The bare `Scope` in this module is the bootstrap config type (`migration::bootstrap::types::Scope`), so I imported the entity type as `ScopeEntity` and call `ScopeEntity::is_custom(...)`; drops the full path without shadowing.

src/jwt/src/claims.rs
bartekus: Went with your "best of both worlds." The fast path is now a short-circuiting `.any()` with no allocation, since this runs on every token issuance; the `Vec` collect + sort happens only on the cold path when a collision actually exists, so the full sorted offender list in the error is unchanged and the existing tests stay green.

I also considered swapping the `&[&str]` reserved list for a `HashSet`, but kept the slice: with ~45 entries the per-key linear scan is negligible next to the allocation you flagged, and a static set adds init machinery for no real gain. Optional, your call: I can add a two-collision regression test asserting the error still lists both keys sorted, to lock the debuggable-message contract through future refactors.

src/data/src/migration/bootstrap/scopes.rs
sebadob: Oh, sorry my bad, I missed that. Then it would have been okay of course, but renaming it is just a fine.

src/jwt/src/claims.rs
sebadob: To find the faster data structure at this size, you would actually have to measure it. When it's just about finding a match, a Vec usually falls behind at ~20 entries. However, we might also have multiple flattened claims, so it's not that clear anymore. I guess we are fine for now.

But, I was just thinking that we could also move it to the modifying endpoint for the scope. Then we would only have to validate once during updates. Unfortunately, it's not that easy. The linked attribute might be renamed later on, which would not remove the link to the custom scope, but rename it in a way that it might produce a collision again. So if we decide to go that route, which would completely get rid of any checks during token creation, we need to validate
1. Any scope that is updated, and before updating, validate all attribute names
2. Any attribute that is updated. We would need to check if it's linked to any scope, and if so, then validate that it cannot create collisions.

This way of checking would be more performant when issuing tokens, because no further checks would be needed. But, let's say we update the Rauthy-managed claims in the future. In this case, the values are stored an assumed as being valid. The only solution in that case would be to include a check in the already existing `temp_migration()` function that handles manual DB migrations between versions, and then emit a warning if any new collisions pop up.

The current version will always work but might be a tiny bit slower, probably not even noticeable now without allocations. The 2nd version would be preferred in terms of performance, but can break a lot easier. 

I think the current version is just fine. We could think about optimizing this in the future though. That was just some brainstorming. :) Thanks!

PR #1598, opened 2026-06-05
feat(api-keys): allow API-key auth for upstream provider endpoints (#1554)

### LINKED ISSUE #1554: Feature request: API-key (non-session) authentication for /providers/* endpoints
## Problem

The 8 endpoints under `/auth/v1/providers/*` (`create`, `lookup`, `update`, `delete`, `link`,
etc.) all gate on `validate_admin_session()`. There is no `Authorization: API-Key …` code path,
so the BOOTSTRAP_API_KEY admins use to manage `/clients/*` and other resources cannot reach
provider configuration.

This means provider setup — e.g. registering Google as the upstream OAuth provider — is the
only Rauthy resource that **cannot be managed via gitops / IaC**. It stays a documented manual
UI step on every fresh install and on every disaster-recovery rebuild after a PVC wipe.

## Use case

I'm running Rauthy in a single-cluster k3s deployment. ArgoCD reconciles a `clients-sync` Job
that uses the Bootstrap API key to declaratively manage OAuth2 clients (ArgoCD, OpenBao, future
agents). The flow works well — declare the client in a values.yaml, ArgoCD syncs, the Job
reconciles. The only piece that doesn't fit is the upstream Google provider, which has to be
re-clicked through the UI any time we rebuild Rauthy's data volume.

The `providers.json` bootstrap-file pattern Rauthy already uses for `clients.json` / `users.json`
would be a natural fit — operators would drop a JSON file alongside the existing bootstrap and
it'd be reconciled on startup. Alternatively (or additionally), accepting `Authorization: API-Key`
on the `/providers/*` endpoints would let consumers manage providers the same way they already
manage clients.

## Proposed solutions, in order of preference

1. **`providers.json` bootstrap file**, parallel to `clients.json` / `users.json`. Schema would
   mirror the `NewProviderRequest` body, with one additional `lookup_only` mode that resolves
   `discovery_endpoint` → issuer/JWKS/etc. fields if not provided. Bootstrap happens on first
   start; subsequent edits go through the existing admin UI. Cleanest fit with the existing
   bootstrap pattern.
2. **`Authorization: API-Key …` accepted on `/providers/*`** alongside admin sessions. Gives
   operators a runtime reconcile path (mirrors how `/clients/*` works today).
3. **Both** — `providers.json` for first-boot bootstrap, API-Key for steady-state reconcile.

I'm happy to take a stab at a PR if a maintainer can confirm the direction. Asking first to
avoid wasted work.

## Workarounds

Today: documented manual UI step on bootstrap + DR. The other workaround we considered —
scripting a session-cookie login + CSRF capture — was rejected because it requires the admin
password to live in vault permanently and breaks on any internal admin-session refactor.

## Versions

Rauthy 0.35.1, single-node, SQLite-backed (Hiqlite), ENC_KEYS scheme.

Thanks for Rauthy — the rest of the gitops story is delightful. This is the one rough edge.

bartekus:
## Summary

Resolves #1554.

Upstream auth provider management (`/providers/*`) was restricted to an admin session, so it could not be driven by an API key or declarative bootstrapping. This adds an `AuthProviders` variant to `AccessGroup` and switches the admin-management provider routes to `validate_api_key_or_admin_session(...)`, mirroring how `groups` / `roles` / `users` already work.

## Changes

- **`AccessGroup`**: new `AuthProviders` variant on both the api-types enum and the data-entity enum, plus the two `From` conversions between them. The entity enum is **bincode-serialized by variant index** for storage, so the new variant is appended last (per the existing note above the enum) to avoid breaking already-stored API keys.
- **Provider handlers** (`auth_providers.rs`): the eight admin-management routes now use `validate_api_key_or_admin_session(AccessGroup::AuthProviders, AccessRights::…)`:
  - `Read`: list providers, config lookup, delete-safe check
  - `Create`: create provider
  - `Update`: update provider, upload logo
  - `Delete`: delete provider, delete logo

  The login-flow / public routes (`login`, `callback`, `link`, `minimal`, image GET) are intentionally left untouched.
- **Admin UI**: `AuthProviders` added to the API-key access matrix (`ApiKeyMatrix.svelte`) and the `AccessGroup` TS type.

## Tests

- **Unit** (`entity/api_keys.rs`): `AuthProviders` round-trips between the entity and api-types `AccessGroup`, and `validate_access` honors it (granted rights pass, ungranted/other-group fail).
- **E2E** (`handler_api_keys.rs`): an API key granted `AuthProviders: [read]` can call `POST /providers`, and the same call is `403` once that access is removed.

Local verification: `cargo check --workspace --tests`, clippy (no new warnings), `cargo fmt --check`, `prettier -c`, and the unit tests all pass; `svelte-check` is clean for the touched files.

### DIFF
--- a/src/api/src/auth_providers.rs
+++ b/src/api/src/auth_providers.rs
@@ -11,6 +11,7 @@ use rauthy_api_types::auth_providers::{ProviderLookupResponse, ProviderResponse}
 use rauthy_api_types::generic::LogoParams;
 use rauthy_api_types::users::{UserResponse, WebauthnLoginResponse};
 use rauthy_common::constants::{HEADER_JSON, PROVIDER_ATPROTO};
+use rauthy_data::entity::api_keys::{AccessGroup, AccessRights};
 use rauthy_data::entity::auth_providers::{
     AuthProvider, AuthProviderLinkCookie, AuthProviderTemplate,
 };
@@ -40,7 +41,7 @@ use validator::Validate;
 )]
 #[post("/providers")]
 pub async fn post_providers(principal: ReqPrincipal) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    principal.validate_api_key_or_admin_session(AccessGroup::AuthProviders, AccessRights::Read)?;
 
     let providers = AuthProvider::find_all().await?;
     let mut resp = Vec::with_capacity(providers.len());
@@ -71,7 +72,8 @@ pub async fn post_provider(
     Json(payload): Json<ProviderRequest>,
     principal: ReqPrincipal,
 ) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    principal
+        .validate_api_key_or_admin_session(AccessGroup::AuthProviders, AccessRights::Create)?;
     payload.validate()?;
 
     if payload.issuer == PROVIDER_ATPROTO {
@@ -118,7 +120,7 @@ pub async fn post_provider_lookup(
     Json(payload): Json<ProviderLookupRequest>,
     principal: ReqPrincipal,
 ) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    principal.validate_api_key_or_admin_session(AccessGroup::AuthProviders, AccessRights::Read)?;
     payload.validate()?;
 
     let resp = AuthProvider::lookup_config(&payload).await?;
@@ -291,7 +293,8 @@ pub async fn put_provider(
     Json(payload): Json<ProviderRequest>,
     principal: ReqPrincipal,
 ) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    principal
+        .validate_api_key_or_admin_session(AccessGroup::AuthProviders, AccessRights::Update)?;
     payload.validate()?;
 
     if !payload.use_pkce && payload.client_secret.is_none() {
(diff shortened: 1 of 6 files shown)


src/api/src/auth_providers.rs
```diff
@@ -486,7 +491,8 @@ pub async fn delete_provider_img(
     id: web::Path<String>,
     principal: ReqPrincipal,
 ) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    principal
+        .validate_api_key_or_admin_session(AccessGroup::AuthProviders, AccessRights::Delete)?;
```
sebadob: This is a DELETE endpoint, but it only deletes the provider image. So I guess we are fine with changing `AccessRights::Delete` to `AccessRights::Update`, since it's actually just an update to the logo.

sebadob (changes requested):

src/api/src/auth_providers.rs
bartekus: Good catch, fixed. `delete_provider_img` now maps to `AccessRights::Update`, matching `put_provider_img`, since clearing the logo mutates the provider rather than removing it; `Delete` stays reserved for removing the provider entity. Added a short comment at the call site so the intentional "DELETE route with Update rights" choice is not reverted later.

While there, I re-checked the rights on all eight management routes for the same semantic: list / lookup / delete-safe are `Read`, create is `Create`, update plus both logo ops are `Update`, provider deletion is `Delete`.

Optional, your call: I can add an e2e test in `handler_api_keys.rs` asserting an `AuthProviders: [update]` key can clear a logo while a key lacking `update` gets a 403, to pin this mapping against future drift. It needs a provider-with-logo fixture, so it adds a little setup; happy to include it, otherwise I will keep the diff lean.

sebadob: > Optional, your call: I can add an e2e test in handler_api_keys.rs asserting an AuthProviders: [update] key can clear a logo while a key lacking update gets a 403, to pin this mapping against future drift. It needs a provider-with-logo fixture, so it adds a little setup; happy to include it, otherwise I will keep the diff lean.

That's totally fine, we don't need that. In integrations tests I am more worried about that API keys work like they should in general. This would be a test for only a single-line-condition. It's super easy to review (and to catch if there might be issues), so it's not worth the additional code for testing just this I think. Also, it's not security critical at all, as it just provides a more flexible way of updating providers, nothing important.

sebadob (approved):
Thanks!

PR #1599, opened 2026-06-07
feat(bootstrap): add generated bootstrap secret extraction

caniko:
This follows the maintainer-accepted scope from #1584: add generated bootstrap secret extraction for Advanced Bootstrapping JSON, without adding an admin-token or trusted-local auth primitive.

### Changes

- Adds an encrypted generated-secret container for bootstrap-created secrets.
- Supports generated client/user secrets during empty production database bootstrap.
- Adds `rauthy bootstrap get` and `rauthy bootstrap purge` for retrieving or purging generated bootstrap secrets.
- Keeps generated secrets out of bootstrap JSON and stores them in the configured encrypted container instead.
- Zeroizes temporary plaintext buffers and generated-secret values after use, including error paths before DB insertion.
- Supports optional generated-secret TTL handling, including expiry rejection and purge behavior.
- Adds generated-secret roundtrip, expiry, purge, version/magic validation, and env-name sanitization tests.
- Documents generated bootstrap secret configuration and CLI usage.

Refs #1584.

### DIFF
--- a/book/src/config/cli.md
+++ b/book/src/config/cli.md
@@ -104,6 +104,77 @@ is very safe while still not being way over the top. If we wanted the same level
 256-bit random bytes key, in alphanumeric form it would be ~43 characters. We have 48 here, which
 even exceeds it. Anything bigger would be a waste of resources.
 
+## Bootstrap Generated Secrets
+
+When `clients.json` or `users.json` uses `"generate"`, Rauthy writes the plaintext values into an
+encrypted local container before inserting the matching database rows. The CLI can read this
+container without starting or contacting the Rauthy server.
+
+The CLI must have the same encryption keys that were active during bootstrap:
+
+```bash
+export ENC_KEY_ACTIVE='2026-06-06'
+export ENC_KEYS='2026-06-06/<base64-key>'
+```
+
+You can also pass them explicitly:
+
+```bash
+rauthy bootstrap get \
+  --file data/bootstrap.secrets.enc \
+  --enc-key-active '2026-06-06' \
+  --enc-keys '2026-06-06/<base64-key>' \
+  --format env
+```
+
+If your deployment uses `USE_VAULT_CONFIG=true`, provide `ENC_KEY_ACTIVE` and `ENC_KEYS` explicitly
+or via the local environment. `rauthy bootstrap get` is an offline decrypt command and intentionally
+does not fetch config from Vault.
+
+To print only one value, use `raw` with an exact selector:
+
+```bash
+rauthy bootstrap get \
+  --file data/bootstrap.secrets.enc \
+  --kind client \
+  --id my-service \
+  --field secret \
+  --format raw
+```
+
+`json` preserves every field:
+
+```bash
+rauthy bootstrap get \
+  --file data/bootstrap.secrets.enc \
+  --kind user \
+  --id admin@example.com \
+  --field password \
+  --format json
+```
+
+`env` emits collision-safe names containing the kind, sanitized id, and field:
+
+```bash
+rauthy bootstrap get --file data/bootstrap.secrets.enc --format env
+```
+
+Example output:
+
+```bash
+RAUTHY_BOOTSTRAP_CLIENT_MY_SERVICE_SECRET=...
+RAUTHY_BOOTSTRAP_USER_ADMIN_EXAMPLE_COM_PASSWORD=...
+```
+
+After you have stored the values somewhere safe, purge the container:
+
+```bash
+rauthy bootstrap purge --file data/bootstrap.secrets.enc
+```
+
+Purging is idempotent: if the file is already absent, the command reports that and exits
+successfully.
+
 ## Hash Passwords
 
 If you want to hash a password for bootstrapping, you can do it like so:
@@ -118,4 +189,4 @@ more if you want to be extra secure.
 
 ## Serve Rauthy
 
-The last option here is `rauthy serve`. This will actually launch Rauthy. 
+The last option here is `rauthy serve`. This will actually launch Rauthy.
--- a/book/src/config/cli.md
+++ b/book/src/config/cli.md
@@ -110,32 +110,23 @@ When `clients.json` or `users.json` uses `"generate"`, Rauthy writes the plainte
 encrypted local container before inserting the matching database rows. The CLI can read this
 container without starting or contacting the Rauthy server.
 
-The CLI must have the same encryption keys that were active during bootstrap:
+The CLI reads everything it needs from your existing Rauthy config file: it loads the same
+`ENC_KEYS` that were active during bootstrap and resolves the configured container location. You
+never pass encryption keys on the command line (which would leak them into your shell history and
+process list) or a separate file path. Just point it at the config:
 
 ```bash
-export ENC_KEY_ACTIVE='2026-06-06'
-export ENC_KEYS='2026-06-06/<base64-key>'
+rauthy bootstrap get --config-file ./config.toml --format env
 ```
 
-You can also pass them explicitly:
-
-```bash
-rauthy bootstrap get \
-  --file data/bootstrap.secrets.enc \
-  --enc-key-active '2026-06-06' \
-  --enc-keys '2026-06-06/<base64-key>' \
-  --format env
-```
-
-If your deployment uses `USE_VAULT_CONFIG=true`, provide `ENC_KEY_ACTIVE` and `ENC_KEYS` explicitly
-or via the local environment. `rauthy bootstrap get` is an offline decrypt command and intentionally
-does not fetch config from Vault.
+`--config-file` defaults to `./config.toml`, so inside the container or next to the config you can
+usually omit it. If your deployment uses `USE_VAULT_CONFIG=true`, the CLI loads the config the same
+way the server does.
 
 To print only one value, use `raw` with an exact selector:
 
 ```bash
 rauthy bootstrap get \
-  --file data/bootstrap.secrets.enc \
   --kind client \
   --id my-service \
   --field secret \
@@ -146,7 +137,6 @@ rauthy bootstrap get \
 
 ```bash
 rauthy bootstrap get \
-  --file data/bootstrap.secrets.enc \
   --kind user \
   --id admin@example.com \
   --field password \
@@ -156,7 +146,7 @@ rauthy bootstrap get \
 `env` emits collision-safe names containing the kind, sanitized id, and field:
 
 ```bash
-rauthy bootstrap get --file data/bootstrap.secrets.enc --format env
+rauthy bootstrap get --format env
 ```
 
 Example output:
@@ -169,7 +159,7 @@ RAUTHY_BOOTSTRAP_USER_ADMIN_EXAMPLE_COM_PASSWORD=...
 After you have stored the values somewhere safe, purge the container:
 
 ```bash
-rauthy bootstrap purge --file data/bootstrap.secrets.enc
+rauthy bootstrap purge
 ```
 
 Purging is idempotent: if the file is already absent, the command reports that and exits
--- a/src/bin/src/utils/bootstrap.rs
+++ b/src/bin/src/utils/bootstrap.rs
@@ -22,13 +22,14 @@ pub async fn run(args: ArgsBootstrap) -> Result<(), StdError> {
 /// Load the existing Rauthy config. This initializes `ENC_KEYS` (so the local
 /// container can be decrypted) and resolves the configured container path,
 /// exactly like the server does — no keys or file path are passed to the CLI.
-async fn load_config(config_file: String) -> Result<RauthyConfig, StdError> {
+async fn load_config(config_file: String) -> Result<&'static RauthyConfig, StdError> {
     let (tx_email, _) = mpsc::channel::<mailer::EMail>(16);
     let (tx_events, _) = flume::unbounded();
     let (tx_events_router, _) = flume::unbounded();
     let (config, _node) =
         RauthyConfig::build(config_file, tx_email, tx_events, tx_events_router).await?;
-    Ok(config)
+    config.init_static();
+    Ok(RauthyConfig::get())
 }
 
 async fn get(args: ArgsBootstrapGet) -> Result<(), StdError> {
--- a/src/bin/src/utils/bootstrap.rs
+++ b/src/bin/src/utils/bootstrap.rs
@@ -1,0 +1,248 @@
+use crate::cli_args::{
+    ArgsBootstrap, ArgsBootstrapCommand, ArgsBootstrapGet, ArgsBootstrapPurge,
+    BootstrapOutputFormat,
+};
+use crate::utils::StdError;
+use cryptr::EncKeys;
+use rauthy_data::migration::bootstrap::generated_secrets::{
+    GeneratedSecretEntry, GeneratedSecrets, read_container,
+};
+use std::env;
+use std::path::Path;
+
+pub async fn run(args: ArgsBootstrap) -> Result<(), StdError> {
+    match args.command {
+        ArgsBootstrapCommand::Get(args) => get(args).await,
+        ArgsBootstrapCommand::Purge(args) => purge(args).await,
+    }
+}
+
+async fn get(args: ArgsBootstrapGet) -> Result<(), StdError> {
+    init_enc_keys(args.enc_key_active.as_deref(), args.enc_keys.as_deref())?;
+
+    let container = read_container(&args.file)
+        .await
+        .map_err(|err| err.message)?;
+    let entries = filter_entries(container, &args);
+    if entries.is_empty() {
+        return Err(format!(
+            "no generated bootstrap secret matched kind={:?}, id={:?}, field={:?}",
+            args.kind, args.id, args.field
+        )
+        .into());
+    }
+
+    match args.format {
+        BootstrapOutputFormat::Raw => {
+            println!("{}", render_raw(&entries)?);
+            Ok(())
+        }
+        BootstrapOutputFormat::Json => {
+            println!("{}", render_json(&entries)?);
+            Ok(())
+        }
+        BootstrapOutputFormat::Env => {
+            print!("{}", render_env(&entries));
+            Ok(())
+        }
+    }
+}
+
+async fn purge(args: ArgsBootstrapPurge) -> Result<(), StdError> {
+    purge_path(Path::new(&args.file)).await
+}
+
+async fn purge_path(path: &Path) -> Result<(), StdError> {
+    match tokio::fs::remove_file(path).await {
+        Ok(()) => {
+            eprintln!(
+                "deleted generated bootstrap secret container {}",
+                path.display()
+            );
+            Ok(())
+        }
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            eprintln!(
+                "generated bootstrap secret container {} was already absent",
+                path.display()
+            );
+            Ok(())
+        }
+        Err(err) => Err(err.into()),
+    }
+}
+
+fn init_enc_keys(enc_key_active: Option<&str>, enc_keys: Option<&str>) -> Result<(), StdError> {
+    let Some(enc_key_active) = enc_key_active else {
+        return Err(missing_keys_error("ENC_KEY_ACTIVE"));
+    };
+    let Some(enc_keys) = enc_keys else {
+        return Err(missing_keys_error("ENC_KEYS"));
+    };
+    let enc_keys = enc_keys
+        .lines()
+        .map(str::trim)
+        .filter(|line| !line.is_empty())
+        .map(ToString::to_string)
+        .collect::<Vec<_>>();
+    if enc_keys.is_empty() {
+        return Err("ENC_KEYS did not contain any encryption keys".into());
+    }
+
+    EncKeys::try_parse(enc_key_active.to_string(), enc_keys)?.init()?;
+    Ok(())
+}
+
+fn missing_keys_error(name: &str) -> StdError {
+    let vault = env::var("USE_VAULT_CONFIG")
+        .unwrap_or_else(|_| "false".to_string())
+        .parse::<bool>()
+        .unwrap_or(false);
+    let prefix = if vault {
+        "USE_VAULT_CONFIG=true, but this offline command does not contact Vault. "
+    } else {
+        ""
+    };
+    format!(
+        "{prefix}missing {name}; pass --enc-key-active / --enc-keys or set \
+         ENC_KEY_ACTIVE / ENC_KEYS so the local bootstrap container can be decrypted"
+    )
+    .into()
+}
+
+fn filter_entries(
+    container: GeneratedSecrets,
+    args: &ArgsBootstrapGet,
+) -> Vec<GeneratedSecretEntry> {
+    container
+        .entries
+        .into_iter()
+        .filter(|entry| {
+            args.kind.as_ref().is_none_or(|kind| &entry.kind == kind)
+                && args.id.as_ref().is_none_or(|id| &entry.id == id)
+                && args
+                    .field
+                    .as_ref()
+                    .is_none_or(|field| &entry.field == field)
+        })
+        .collect()
+}
+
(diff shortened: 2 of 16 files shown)


book/src/config/cli.md
```diff
@@ -104,6 +104,77 @@ is very safe while still not being way over the top. If we wanted the same level
...
+export ENC_KEYS='2026-06-06/<base64-key>'
+```
+
+You can also pass them explicitly:
+
+```bash
+rauthy bootstrap get \
+  --file data/bootstrap.secrets.enc \
+  --enc-key-active '2026-06-06' \
+  --enc-keys '2026-06-06/<base64-key>' \
```
sebadob: That's a very bad thing to do, and we should not even allow this. You would have the enc keys in clear text in your shell history, and also visible in other places like when you inspect the process, and so on.

In fact, we don't even need to parse any enc keys at all. Neither via arg, nor via ENV export. This is way too complicated and has a bad UX. They are almost always auto-generated and you probably don't even know them. Make your life a lot easier: Just parse the existing config.

Apart from that, the active key is never needed when only decrypting.

book/src/config/cli.md
```diff
@@ -104,6 +104,77 @@ is very safe while still not being way over the top. If we wanted the same level
...
+
+If your deployment uses `USE_VAULT_CONFIG=true`, provide `ENC_KEY_ACTIVE` and `ENC_KEYS` explicitly
+or via the local environment. `rauthy bootstrap get` is an offline decrypt command and intentionally
+does not fetch config from Vault.
+
+To print only one value, use `raw` with an exact selector:
+
+```bash
+rauthy bootstrap get \
+  --file data/bootstrap.secrets.enc \
```
sebadob: I would not even take the `file` arg. The same as above: just parse the current config. If any custom file exists, it will be set inside it, and there is no need to provide the value again.

src/bin/src/utils/bootstrap.rs
```diff
@@ -0,0 +1,248 @@
...
+        Ok(serde_json::to_string_pretty(&entries[0])?)
+    } else {
+        Ok(serde_json::to_string_pretty(entries)?)
+    }
+}
+
+fn render_env(entries: &[GeneratedSecretEntry]) -> String {
+    let mut out = String::new();
+    for entry in entries {
+        out.push_str(&format!("{}={}\n", entry.env_name(), entry.value));
```
sebadob: Not a big deal for a CLI, but the `write!()` macro is a lot faster and does not do an additional `String` allocation because of `format!()`.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+    entry: GeneratedSecretEntry,
+) -> Result<(), ErrorResponse> {
+    let path = path.as_ref();
+    let mut container = if path.exists() {
+        read_container(path).await?
+    } else {
+        GeneratedSecrets::default()
+    };
+    container.upsert(entry);
+    let deadline = deadline_from_ttl(ttl_seconds, SystemTime::now());
```
sebadob: Probably not a huge issue here, but genereally always do everything in `UTC` time. This avoids many issues and headaches down the road, especially in distribubted systems. If you use `SystemTime` - what if the volume is an NFS volume mounted to multiple hosts with different times set? It might happen that another instance immediately deletes the container because it thinks it's expired for instance. There are lots of problematic situations.

In general: Always work in UTC until you reach a point where you need to display a timestamp to a user in a nicely readable format, then convert to the current local time on the current machine the user is working on.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+        internal(format!(
+            "bootstrap generated-secret path '{}' has no parent directory",
+            path.display()
+        ))
+    })?;
+    tokio::fs::create_dir_all(parent).await?;
+
+    let mut payload = serde_json::to_vec(container)?;
+    let encrypted = match EncValue::encrypt(&payload) {
+        Ok(value) => value.into_bytes().to_vec(),
```
sebadob: You should not need the `Vec<_>` allocation here. If you did it because of the `zeroize()`, you would not even need if you work with borrowed data. Only the `.to_vec()` creates a real copy. The same is true for the `decrypted` above. `Bytes` does (usually) not allocate under the hood.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+
+    let tmp = tmp_path(path)?;
+    {
+        let mut options = OpenOptions::new();
+        options.write(true).create_new(true);
+        #[cfg(unix)]
+        options.mode(0o600);
+        let mut file = options.open(&tmp)?;
+        file.write_all(&bytes)?;
+        file.sync_all()?;
```
sebadob: The sync / flush will be done immediately automatically, because the `file` is dropped after this line, but keeping the `sync_all()` would be fine as well.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+    }
+    bytes.zeroize();
+    std::fs::rename(&tmp, path)?;
```
sebadob: This is a blocking call in an async context. You should use the `tokio` version for correctness. The same is true for the `File::open()` below.  Not sure why you are opening it to do another fsync? This is done automatically on drop. Would be super weird if you actually need this.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+    let file_name = path.file_name().and_then(OsStr::to_str).ok_or_else(|| {
+        internal(format!(
+            "invalid generated-secret path '{}'",
+            path.display()
+        ))
+    })?;
+    let nanos = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_nanos();
```
sebadob: No one really needs nanos precision. Usually, seconds is good enough and a lot shorter. We also don't need to care about file conflicts here and use the process id. There can always only run a single Rauthy process at a time, and the bootstrapping is single threaded. 

When I write into files, I usually follow the same naming as most editors and use `<original name>~` to keep things simple.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+    Ok(path.with_file_name(format!(".{file_name}.{}.{nanos}.tmp", std::process::id())))
+}
+
+fn unix_seconds(now: SystemTime) -> u64 {
+    now.duration_since(UNIX_EPOCH).unwrap_or_default().as_secs()
+}
+
+fn internal(msg: impl Into<String>) -> ErrorResponse {
+    ErrorResponse::new(ErrorResponseType::Internal, msg.into())
```
sebadob: It probably makes a lot of sense to start adding these shorthands to the `ErrorResponse` `impl` directly. This would clean up the code in lots of places. I have that on the TODO anyway.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,482 @@
...
+        "_".to_string()
+    } else if out.as_bytes()[0].is_ascii_digit() {
+        format!("_{out}")
+    } else {
+        out
+    }
+}
+
+fn render_header(deadline: u64) -> Vec<u8> {
+    format!("{MAGIC}\n{VERSION}\n{deadline}\n").into_bytes()
```
sebadob: If you really want to write your own custom header instead of using a `struct` inside the encrypted data, always be very specific about everything. It's probably just fine in this case, because you separate by `\n` and then parse from a `String`, but you usually don't want to do this, because it's slow and wasteful. Instead you would want to work with bytes directly. Then there was no need any separators, because you know the padding. However, you then also need to be very specific about converting integers. Never just use the native format. Instead, lock-in to always use LE or BE (does not really matter in this case as long as it's the same). The native defaults do not always match when moving between architectures.

This is only a CLI and performance really does not matter at all, but you would usually not want to do it like that. You are also actually making your life harder than it needs to be here. You could just have a wrapping struct that has a field for the `exp` or `deadline`, and then `Vec<_>`s for the secrets. You could simply encrypt the whole thing and just write it to the file without the manual header handling. `cryptr` uses AEAD, more specific ChaCha20Poly1305. If the decryption succeeds, the data is valid, and you would not need a magic str or versioning.

But that's up to you. I am fine with both versions. That was just a small rant on manually working with situations like these.

sebadob (changes requested):

caniko:
All 10 review points are addressed in commit f46cea8d on `feat/bootstrap-generated-secrets-pr1-submit`.

**Structural highlights:**
- **Config-driven CLI** — `rauthy bootstrap get/purge` now loads `RauthyConfig::build()` (same as `validate-config`), which initializes `ENC_KEYS` from config and resolves the container path. No more `--enc-keys`/`--enc-key-active`/`--file` on the command line.
- **Single AEAD struct** — the on-disk format is one `cryptr` encrypted JSON value containing `{version, deadline, entries}`. The manual magic/version/deadline/header + payload assembly is gone. Decryption is the integrity check; expiry is read via a cheap decrypt.

**Per-point breakdown:**
1. CLI drops `--enc-keys`/`--enc-key-active`/`--file` — config-driven via `RauthyConfig::build()`.
2. No manual header — one AEAD value over `ContainerPayload { version, deadline, entries }`.
3. UTC time via `chrono::Utc::now().timestamp()` for deadlines.
4. Blocking I/O replaced with `tokio::fs` write/rename; redundant dir-fsync dropped.
5. Works on `cryptr Bytes` directly — no `.to_vec()` copies.
6. Editor-style temp file (`<name>~`), no pid/nanos.
7. `writeln!` into buffer instead of `format!` per entry.
8. ~**Deferred to your TODO** — kept the local `internal()` helper in `generated_secrets.rs` rather than moving it onto `ErrorResponse`.~
9. Docs (`cli.md`) updated for config-driven usage.
10. Updated tests: `garbage_file_fails_to_decrypt`, `expired_container_is_rejected`, `purge_removes_expired_file` pass with the new format.

Ready for re-review.

book/src/config/cli.md
caniko: > That's a very bad thing to do, and we should not even allow this. 

I agree, I was trying to stay true to the RFC, forgot this very important principle.

caniko:
Update on item 8: `ErrorResponse::internal()` is now available in ba0ad7b8 and used in `generated_secrets.rs` — the local `fn internal()` helper is gone.

```rust
// src/error/src/error_impls.rs
pub fn internal(message: impl Into<Cow<'static', str>>) -> Self {
    Self::new(ErrorResponseType::Internal, message)
}
```

caniko:
Fixed: startup purge is now best-effort (warns, never panics).

`server.rs` startup call changed from `.expect(...)` to `if let Err(err) = ... { warn!(...) }` — if the container is undecryptable (key rotation, corrupt/stray file), the server boots with a warning instead of panicking. The stale container stays on disk; the operator can recover it with the right key or `rauthy bootstrap purge`.

Also resolved item 8 from the previous round — `ErrorResponse::internal()` is now a real method on `ErrorResponse` (was a local helper in generated_secrets.rs). Both fixes in `ba0ad7b8..7eccb0d2`.

bartekus (commented):
I built and ran this branch (head `7eccb0d2`) as a headless-provisioning consumer would: fresh empty prod-mode DB (hiqlite), an Advanced Bootstrapping set with a confidential client using `"secret": "generate"` and a user with `"password": "generate"`, then the full extract/expire/purge/rotation lifecycle via the CLI. Here is what I found.

## Confirmed working

All of the below verified end-to-end (debug build, macOS arm64, single node, `data/` as hiqlite data dir):

- First boot on an empty prod DB writes `data/bootstrap.secrets.enc` with `0600` perms via the `~` temp file + rename. Neither plaintext ever shows up in logs (`grep <value> server.log` → 0 hits) or anywhere else on disk.
- The generated values actually work against the running server:
  - `POST /auth/v1/oidc/token` with `grant_type=client_credentials` + extracted client secret → 200 (wrong secret → 401);
  - `grant_type=password` with the extracted user password → 200.
- `rauthy bootstrap get` offline (server stopped), all three formats, with and without `--kind/--id/--field` selectors; no-match → clean error, exit 1. *(Caveat: only after a local one-line fix, see finding 1.)*
- `rauthy bootstrap purge` deletes; second run reports "already absent" and exits 0.
- TTL runtime auto-purge: with `generated_secrets_ttl = 5`, the container disappears ~1s after the deadline while the server runs, with the info log line.
- Expired container + server down: `get` → `Error: "bootstrap generated-secret container data/bootstrap.secrets.enc expired at 1781228586"`, file left in place; next server start purges it (info log). Both halves behave as documented.
- Key rotation: flipped `key_active` to a second key with the original still in `encryption.keys` → `get` decrypts fine (cryptr header key id); removed the original from the list → clean `Error: "CryptrError::Keys(EncKey ID does not exist)"`.
- Day-2 boot with a corrupt container on an initialized DB: warn + normal startup (`7eccb0d2` does what it says).
- Unit tests: all 16 across `generated_secrets`, `types`, and the CLI module pass.

## Findings

### 1. `rauthy bootstrap get` panics unconditionally — `ENC_KEYS` are never initialized

```
$ rauthy bootstrap get -c config.toml
thread 'main' panicked at .../cryptr-0.10.0/src/keys.rs:381:14:
`init()` to have been called on valid EncKeys once before
```

`load_config()` in `src/bin/src/utils/bootstrap.rs` calls `RauthyConfig::build()`, but `EncKeys::init()` only happens inside `RauthyConfig::init_static()`, which the server calls in `server.rs` and the CLI never does. So every `get` dies at the first decrypt. The module's unit tests don't catch it because they call `keys.init()` themselves.

This one-liner fixed it for me, and everything in the "confirmed working" list above was run on top of it:

```rust
async fn load_config(config_file: String) -> Result<&'static RauthyConfig, StdError> {
    let (tx_email, _) = mpsc::channel::<mailer::EMail>(16);
    let (tx_events, _) = flume::unbounded();
    let (tx_events_router, _) = flume::unbounded();
    let (config, _node) =
        RauthyConfig::build(config_file, tx_email, tx_events, tx_events_router).await?;
    config.init_static();
    Ok(RauthyConfig::get())
}
```

`purge` is unaffected (no decrypt).

### 2. A stale container file at *first* boot leaves the DB half-initialized, silently

Setup: empty prod DB, but a non-decryptable `bootstrap.secrets.enc` already exists at the target path (stale file from a previous install attempt, wrong key, partial copy — exactly the situations the startup purge was added for).

First boot:

```
WARN rauthy::server: Could not purge bootstrap generated-secret container: ... CryptrError::Deserialization(Invalid EncFileVersion)
thread 'main' panicked at src/bin/src/server.rs:105:25:
Database migration error: ErrorResponse { ... message: "CryptrError::Deserialization(Invalid EncFileVersion)" }
```

The best-effort startup purge warns and moves on (good), but `upsert_secret()` then trips over the same file inside `read_container()` and the error propagates out of `DB::migrate()`. The real problem is the second boot: `rauthy_admin` and `jwks` bootstrap already ran before `bootstrap_additional_data()`, so the JWKS table is non-empty, `migrate_init_prod()` returns early, and the server comes up healthy — with the declared clients and users never created and no error anywhere. For a headless pipeline that's the worst case: provisioning silently incomplete, and the only recovery is wiping the DB or creating everything manually.

Same shape via a different trigger: a leftover `bootstrap.secrets.enc~` (crash mid-write) makes `write_container()`'s `create_new(true)` fail with `IO Error: File exists (os error 17)` → same panic, same half-initialized DB on reboot.

Suggestion — since the container only ever holds first-boot secrets, an unreadable one at bootstrap time is garbage by definition: let `upsert_secret()` warn and start from `GeneratedSecrets::default()` when `read_container()` fails (same best-effort semantics you already applied to the startup purge), and open the temp file with `.create(true).truncate(true)` instead of `create_new` so a stale `~` file can't block the write. That would make first boot self-healing for both cases. Happy to test a revision.

### 3. `bootstrap.md` and `config.toml` still document the previous design

Both still describe the "cleartext magic/version/deadline header" (`book/src/config/bootstrap.md` and the `config.toml` comment: "An expired file can therefore be purged without decrypting it") — after the AEAD refactor the deadline lives inside the encrypted payload and the purge does decrypt. `bootstrap.md` also still shows the removed CLI interface (`--file`, `--enc-keys`, `--enc-key-active`, "pass these key values explicitly to the CLI" for Vault), contradicting the updated `cli.md` on the same points. The `config.toml` comment additionally mentions "API-key tokens", which the PR doesn't implement (see question below).

## Questions

- Is `"generate"` for `api_keys.json` secrets intended for this PR or a follow-up? The `config.toml` comment text suggests it, and headless flows would benefit (API keys can't currently be created via API-key auth), but clients+users alone already make this useful — just asking so the docs match the actual v1 scope.
- Minor: `get` on a missing/never-created container prints `Error: "IO Error: No such file or directory (os error 2)"` without the path. Given the per-case error quality elsewhere in the config code, a dedicated "no generated bootstrap secrets stored at <path>" message might be worth it.
- Minor: in `decrypt_container()`, the decrypted JSON buffer (which holds every plaintext value) is dropped without zeroizing; the entries themselves zeroize on `Drop`. Cheap to add if you care about it at this layer.

caniko:
Thank you so much for testing @bartekus, I really appreciate it 💪🏻 will work on these issues

caniko:
Thanks again for the detailed test pass. I pushed `b2ef7e0f` with fixes for the findings:

- `rauthy bootstrap get -c config.toml` now initializes the static config / `ENC_KEYS` path before decrypting, so it no longer panics.
- First boot now self-heals stale generated-secret artifacts: an unreadable existing container is warned about and replaced during first-boot secret upsert, and a stale `<container>~` temp file is truncated/reused instead of blocking bootstrap.
- Missing containers now return a path-specific error.
- The decrypted JSON buffer is zeroized when it can be reclaimed mutably from `Bytes`.
- `bootstrap.md` and `config.toml` now describe the current AEAD payload format and config-driven CLI, and no longer imply API-key generated secrets are included in this PR.

For the API-key generation question: I kept it out of this PR and corrected the docs/comments so v1 scope is clearly generated client secrets and user passwords.

bartekus:
Re-ran the same consumer pass against `b2ef7e0f`; all five are fixed, verified end-to-end (not just the unit tests):

- **CLI panic:** `rauthy bootstrap get -c config.toml --format env` now decrypts and prints; no more `EncKeys init()` panic. `raw`/`json`/selector variants all work.
- **Corrupt container at first boot (the one that worried me):** pre-staged an undecryptable `bootstrap.secrets.enc` on an empty prod DB. Now: `WARN ... starting with a fresh container`, then `Migrated 1 clients` / `Migrated 1 roles`, server boots healthy, the container is replaced with a valid `0600` file, and the extracted client secret authenticates (`client_credentials` → 200). No more silently-half-initialized DB.
- **Stale `<container>~` temp file at first boot:** pre-staged the temp file on an empty DB. Bootstrap no longer hits `File exists (os error 17)`; it truncates/reuses, boots clean, leaves a single valid `0600` container, secret retrievable.
- **Missing container:** `Error: "no generated bootstrap secrets stored at <path>"`; path-specific as hoped.
- **Zeroize + docs:** decrypted buffer zeroized via `try_into_mut` on both the ok and error paths; `bootstrap.md`/`config.toml` now describe the AEAD payload and config-driven CLI, and the API-key wording is gone. Agree with keeping api_key generation out of v1.

Tests: the full `generated_secrets` + CLI suite passes here (9 + 6), including the three new regression tests for these paths. Nothing else flagged on my end; LGTM from a consumer standpoint.

src/data/src/migration/bootstrap/generated_secrets.rs
```diff
@@ -0,0 +1,549 @@
...
+
+fn is_expired(deadline: i64, now: i64) -> bool {
+    deadline > 0 && now >= deadline
+}
+
+pub fn deadline_from_ttl(ttl_seconds: u64, now: i64) -> i64 {
+    if ttl_seconds == 0 {
+        0
+    } else {
+        now.saturating_add(ttl_seconds as i64)
```
sebadob: This is an unsafe downcast of `ttl_seconds`. To make this safe, it should be either a parsed `i64` from the beginning, or a `u32`. `i64` does not make much sense logically, since it should always positive. A `u32` can be cast to `i64` safely, as it always "fits in".

src/data/src/rauthy_config.rs
```diff
@@ -3430,6 +3457,8 @@ pub struct VarsBootstrap {
     pub api_key: Option<String>,
     pub api_key_secret: Option<String>,
     pub bootstrap_dir: Cow<'static, str>,
+    pub generated_secrets_file: Cow<'static, str>,
+    pub generated_secrets_ttl: u64,
```
sebadob: Mentioned above already, changing this to `u32` makes more sense to make the cast to `i64` always safe. A `u32` is way more than big enough here as well.

src/data/src/rauthy_config.rs
```diff
@@ -3934,6 +3963,16 @@ fn t_u32(map: &mut toml::Table, parent: &str, key: &str, env_overwrite: &str) ->
         None
     }
 }
+fn t_u64(map: &mut toml::Table, parent: &str, key: &str, env_overwrite: &str) -> Option<u64> {
```
sebadob: When `ttl_seconds` is made a `u32`, we don't need this function anymore.

sebadob (changes requested):
Apart from the type of `ttl_seconds`, looks good now.

The only thing left is that we could also make it possible to generate API Key secrets. The additional code would be not that much work, and then it's really solid.

caniko:
Addressing the remaining review points:

- Changed `generated_secrets_ttl` from `u64` to `u32` through config parsing and generated-secret helpers.
- Replaced the TTL deadline cast with safe widening conversions.
- Removed the now-unused `t_u64()` parser helper.
- Added `"secret": "generate"` support for `api_keys.json`.
- Generated API-key secrets are stored in the same generated-secret container as `kind=api-key`, `id=<key name>`, `field=secret`.
- Updated `bootstrap.md` and `config.toml` to document generated API-key secrets.

bartekus:
Re-ran the consumer pass against the API-key `generate` commit (`6f4382d2`) on a fresh prod hiqlite DB. The plaintext-exposure trade holds end to end:

- A bootstrapped `"secret": "generate"` key writes its 64-char plaintext only to the encrypted container. Grepping the minted secret afterwards: 0 hits in `api_keys.json` (only the literal `generate`), 0 in the logs, 0 anywhere else in the data dir. The stored column is `encrypt(sha256(secret))`, so the container is the only place the cleartext exists.
- Extraction works in all three formats; `--format env` sanitizes the kebab kind into a valid shell identifier (`RAUTHY_BOOTSTRAP_API_KEY_PROVISION_KEY_SECRET=...`). The key authenticates an admin call (`API-Key <name>$<secret>` → `GET /auth/v1/clients` 200; wrong/absent → 401).
- With a positive TTL the runtime purge removes the container while the server runs (`Purged expired bootstrap generated-secret container ...`), and the key keeps authenticating afterward since the DB copy is independent. Post-purge `bootstrap get` fails cleanly (`no generated bootstrap secrets stored at ...`) instead of panicking. `generated_secrets_ttl = 0` keeps the container.

One small thing: `sanitize_env_name` maps `-`, ` `, and `_` all to `_` (and folds case), so two differently-named entries can collide on the same `--format env` variable (e.g. `my-key` and `my_key` both become `RAUTHY_BOOTSTRAP_API_KEY_MY_KEY_SECRET`). The `raw`/`json`/`--kind/--id/--field` forms use the original id and stay unambiguous; only the env serialization is lossy, so it might be worth a doc note that the env form isn't guaranteed unique. Not a blocker.

Docs (`bootstrap.md`) extraction command and the `API-Key <name>$<secret>` header match what I ran.

sebadob (approved):
This looks really good now. Thank you!

PR #1604, opened 2026-06-14
feat(clients): custom claims on `client_credentials` tokens

### LINKED ISSUE #1601: Custom claims on `client_credentials` tokens
## Problem

Rauthy's custom-attribute pipeline is entirely user-bound. Scope-mapped
attributes are read from `UserAttrValueEntity` (`(user_id, key, value)`) in
`TokenSet::build_access_token()` (`src/service/src/token_set.rs`), so a token
minted through the `client_credentials` grant carries no custom claims at all:
`TokenSet::for_client_credentials()` (`token_set.rs:498`) calls
`build_access_token` with `scope_customs: None`. The only client-driven
identity knob today is `CLIENT_CREDENTIALS_MAP_SUB`, which copies `client.id`
into `sub`. There is no way to attach any deployment-defined claim to a machine
client's token.

The practical consequence is that every resource server protected by a
`client_credentials` client has to do an out-of-band lookup keyed on
`client_id` (or `azp`) to recover anything about the caller: which workload it
is, which tenant it belongs to, what it is allowed to do. That defeats the
point of a self-describing bearer token. Concrete cases where the claim wants
to ride the token directly: workload identity (a per-service or per-CI-runner
client whose token should say which service / pipeline it is), agent-style MCP
clients (the direction in #1562, where the connector is the OAuth client and
the resource server needs more than an opaque id), and multi-tenant SaaS where
the tenant id must be present on the access token for the RS to authorize
without a second round-trip.

## Design options

Smallest first. My recommendation is **(B)**.

### (B, recommended): a `claims` JSON object on the client

A nullable `claims` column on `clients`: a flat JSON object of scalar values
(string / number / bool), set by an admin or via the management API, emitted
verbatim into `client_credentials` access tokens. Minimal surface: no new
tables, no scope plumbing. It reuses the existing emission path
(`JwtAccessClaims.custom` / `custom_flattened`) and the collision guard added
in #1597.

- **Where they land:** nested under `custom` by default, exactly like the user
  path. A sibling `claims_at_root` bool on the client opts into root-level
  (flattened) emission, with the same warning and the same
  `validate_no_reserved_collision()` fail-to-issue guarantee that #1597 already
  enforces, so a client can never shadow a reserved claim.
- **Write-time validation:** object only, scalar values only (reject nested
  objects / arrays), and a size cap (I would suggest 4 KiB on the serialized
  object) with a dedicated per-field error message, in the same spirit as the
  hand-parsed config errors elsewhere.
- **Threading:** identical to the `claims_at_root` change (#1597 /
  `ff81b94d`): one column, both migrations (hiqlite + postgres), the field
  through `Client` CRUD on both DB paths, `ClientRequest` / `ClientResponse`
  with serde defaults for back-compat, declarative bootstrap, and the admin-UI
  client form. Absent column / null value means no claims, so existing clients
  are unaffected.

### (A): full parity with the user attribute pipeline

A client-attribute table plus per-scope `attr_include_*` mapping, mirroring
`UserAttrConfigEntity` + `UserAttrValueEntity` + `Scope`. More uniform with the
user side and reusable across grant types, but it is a lot more surface (new
table, new CRUD, scope plumbing, admin UI for the attribute catalog) for what
is, for machine clients, usually a small fixed set of claims known at
provisioning time. I do not think the generality earns its cost here, but
flagging it as the "consistent with users" option.

### (C): reuse the user-attr tables keyed by `client_id`

Least new code, worst semantics: store client claims as
`UserAttrValueEntity` rows with the `client_id` in the `user_id` slot. It
overloads a user-scoped table with non-user rows and would confuse every
existing query and the admin UI. Listed only to show it was considered and
rejected.

## Scope notes

- **Dynamic clients (`dyn$`)** can be confidential and can use the
  `client_credentials` flow today (`grant_type_credentials` handles
  `client.is_dynamic()` explicitly). For v1 I propose dynamic clients
  **cannot** set their own `claims` through the dynamic client registration
  endpoint (RFC 7591): a client registering itself and then asserting its own
  claims is a privilege-escalation foot-gun. Claims would be settable only via
  the admin UI / management API, i.e. by the operator, not by the
  self-registering client. The registration request would either ignore or
  reject the field.
- **Ephemeral clients** (`ephemeral_from_url`): out of scope for v1.

## Precedent

This reuses the same emission and collision machinery that #1597
(`claims_at_root`) added for the user path; the only new pieces are the
client-side column and the `client_credentials` emission point.

Happy to implement once you pick a direction (I am assuming B unless you would
rather go another way). Anything you would want done differently on the
root-vs-nested behaviour or the validation limits?

bartekus:
## Summary

Resolves #1601.

Adds admin-defined custom claims on a `Client`, emitted into the access token of
the `client_credentials` flow, nested under the `custom` claim. This lets machine
clients (which have no user) carry self-describing claims: workload identity,
tenant id, deployment profile, so a resource server does not have to look anything
up by `client_id`. Storage mirrors the groups/roles JSON `meta` work (#1491);
emission reuses the existing `custom` map.

## Design

- **Nested by default, opt-in root.** Claims are emitted as a single `custom`
  object, exactly like user custom attributes. A client-level `claims_at_root`
  flag promotes them to the token root (flattened) instead, guarded by the
  `validate_no_reserved_collision()` check from #1597 so a root-promoted client
  claim can never shadow a reserved claim (issuance fails instead). This mirrors
  the per-scope `claims_at_root` routing, without the scope plumbing.
- **Admin surface only.** Claims are set via the Admin UI / management API.
  Dynamic (`dyn$`) and ephemeral clients cannot set their own claims: their request
  types carry no `claims` field, so a self-registering client cannot assert its own
  claims (a privilege-escalation foot-gun).

## What's included

- **DB**: nullable `claims` column + `claims_at_root` bool on `clients`
  (`BLOB`/`BYTEA` + `INTEGER`/`bool`), both migrations. `claims` is stored as
  serialized JSON bytes (`Option<Vec<u8>>`), same as `Group.meta`, to avoid
  `Bincode: Serde(AnyNotSupported)`.
- **Entity**: both fields threaded through the client `save` / update on both DB
  paths; `into_response` decodes `claims` back to JSON for `ClientResponse`.
  `create` paths leave them at default, so they can only be set through the admin
  `PUT`.
- **API types**: `claims` + `claims_at_root` on `UpdateClientRequest` (serde
  `default` for back-compat) and `ClientResponse`.
- **Validation** (on client `PUT`): `validate_claims` requires a JSON object and
  caps the serialized client claims at 1024 characters. The cap applies to the
  client JSON only, not the whole token (which is often kept small / cookie-stored).
- **Emission**: in `build_access_token`, gated on `user.is_none()` so it applies
  only to the `client_credentials` flow; routed to the token root (flattened, with
  the collision guard) or nested under `custom` by `claims_at_root`.
- **Admin UI**: a `Custom Claims` JSON editor + a `claims_at_root` toggle (with the
  same reserved-claim warning + IANA registry link as the per-scope toggle) on the
  client config form, mirroring the groups `meta` editor; i18n keys added to all
  locales.

## Also fixes: postgres migration `v24` was silently skipped

While validating the new migration on postgres, a fresh postgres bootstrap panics
at the scopes step (`SqlState 42703`, undefined column `claims_at_root`). refinery
0.9.1 only recognizes migration files whose name matches `^([U|V])(\d+...)__(\w+)`,
i.e. an **uppercase** `V`/`U` prefix. `migrations/postgres/v24__cust_attrs_token_root.sql`
(added in #1596) is **lowercase**, so refinery skips it: `scopes.claims_at_root` is
never created on postgres, and the #1597 scopes bootstrap that INSERTs it then
fails. This is needed for this PR anyway (the new `V25` runs after it), so it is
renamed `v24` -> `V24` here. hiqlite is unaffected (numeric prefixes). I verified
the `client_credentials` E2E test passes on postgres only after this rename.

## Tests

- **Unit** (`api_types/cust_validation`): object-required, and the 1024-char cap
  (pass + fail boundaries).
- **E2E** (`zzd_handler_clients`): set claims on a confidential client, request a
  `client_credentials` token, assert they are nested under `custom` (not at root);
  flip `claims_at_root` and assert they are emitted flattened at the root (and not
  under `custom`); assert a root-promoted claim colliding with a reserved claim
  (`iss`) fails issuance; then clear and assert no `custom` (back-compat). Runs on
  both hiqlite and postgres.

## Notes

- Back-compat: the column is nullable; existing clients (absent / null) emit no
  `custom`. `UpdateClientRequest.claims` has a serde default, so older API clients
  that omit it are unaffected.
- Non-English i18n strings use an English fallback for the two new keys;
  translations welcome.

Local verification: `cargo check --workspace --tests`, clippy (no new warnings),
`cargo fmt --check`, the `api_types` unit tests, the `client_credentials` E2E test
on both hiqlite and postgres, and `svelte-check` (clean for the touched frontend
files) all pass.


frontend/src/i18n/admin/en.ts
```diff
@@ -46,6 +46,11 @@ export let I18nAdminEn: I18nAdmin = {
             descVariables: `Each following label is at the same time the name of the CSS variable. This means,
                 that you can reference these in the free inputs, e.g. with <code>hsla(var(--action) / .7)</code>.`,
         },
+        claimsAtRoot: 'Emit claims at the token root',
+        claimsAtRootWarning: `When enabled, this client's claims are written at the token root instead of being nested under 'custom'. You own collision-correctness: if a claim name collides with a reserved JWT claim, token issuance fails. Root-level custom claims may also break with future protocol or feature changes. Reference:`,
+        claims: 'Custom claims (client_credentials)',
+        claimsDesc:
+            'A JSON object emitted into client_credentials tokens, nested under the custom claim. Admin-set only; dynamic and ephemeral clients cannot set it. Max 1024 serialized characters.',
```
sebadob: For all the translations in the UI, please manually insert a line-break roughly at the same position as all the others in these files. Auto-formatting does not work here. We want to keep them at about the same length as everything else.

src/api_types/src/cust_validation.rs
```diff
@@ -24,6 +24,30 @@ pub fn validate_vec_attr(value: &[String]) -> Result<(), ValidationError> {
     Ok(())
 }
 
+/// Custom claims attached to a client and emitted into `client_credentials`
+/// tokens. Must be a JSON object (it is emitted nested under the `custom` claim)
+/// and is capped to keep tokens small, since they are frequently stored in
+/// cookies. The cap applies to the client's custom JSON only, not the full token.
+#[inline]
+pub fn validate_claims(value: &serde_json::Value) -> Result<(), ValidationError> {
+    const MAX_LEN: usize = 1024;
```
sebadob: I would move the `MAX_LEN` into `src/common/src/constants.rs`, so we have all constants in one place.

src/api_types/src/cust_validation.rs
```diff
@@ -24,6 +24,30 @@ pub fn validate_vec_attr(value: &[String]) -> Result<(), ValidationError> {
...
+    if !value.is_object() {
+        return Err(ValidationError::new("`claims` must be a JSON object"));
+    }
+
+    let len = serde_json::to_vec(value)
+        .map(|v| v.len())
+        .unwrap_or(usize::MAX);
+    if len > MAX_LEN {
+        return Err(ValidationError::new(
+            "`claims` must not exceed 1024 serialized characters",
```
sebadob: We can re-use the `MAX_LEN` here inside the err msg instead of hard coding, so it could be changed easily.

sebadob (changes requested):

bartekus:
Thanks for the review. Addressed all three:

- Moved the cap into `src/common/src/constants.rs` as `CLIENT_CLAIMS_MAX_LEN`.
- The validation error now interpolates that constant instead of hardcoding the
  number, so changing the cap changes the message too.
- Hand-wrapped the new admin i18n strings (`claimsAtRoot` / `claimsAtRootWarning` /
  `claims` / `claimsDesc`) across all locales to roughly match the surrounding line
  width, instead of relying on auto-format.

Code Style CI is green. Let me know if anything else.

sebadob (approved):
Thank you

Issue #1562, opened 2026-05-13
Make Rauthy work end-to-end as the AS for MCP custom connectors (ChatGPT + claude.ai)


BorisTyshkevich:
## Context

[MCP](https://modelcontextprotocol.io/) (Model Context Protocol) is the emerging standard for connecting LLM clients to remote tool/data servers. As of [MCP authz spec 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization), every MCP server acts as an OAuth 2.1 resource server and the LLM client (ChatGPT, claude.ai, others) acts as the OAuth 2.1 client. Servers tell clients where to authenticate via RFC 9728 protected-resource metadata; clients prefer **Client ID Metadata Documents** (CIMD, SEP-991 / [draft-ietf-oauth-client-id-metadata-document-00](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00)) for registration because there's typically no prior relationship between client and AS.

I spent the last few days running Rauthy as the AS for a small fleet of MCP servers, end-to-end with both ChatGPT and claude.ai as clients. Rauthy's `ephemeral_from_url` already implements the CIMD-style flow; the gaps were all in discovery, claim emission, and admin-UI URL handling. This issue documents the experiment, results, and the three PRs.

## TL;DR

| Want | Status |
|---|---|
| ChatGPT custom-connector "Add app" → CIMD option ungreyed | ✅ after PR #1561 |
| claude.ai custom connector via "Advanced settings" → user-defined OAuth client with URL-shaped `client_id` | ✅ after PR #1559 (admin UI to manage the static client) |
| Resource server enforces `email_verified=true` on the access token | ✅ after PR #1560 |
| Audience-bound tokens via RFC 8707 `resource=` (only relevant for strict gating mode) | ❌ separate work; not in these PRs |

End-to-end ✅: both ChatGPT and claude.ai connect to `https://otel-rauthy-mcp.demo.altinity.cloud/mcp`, log in via Google brokered through Rauthy, and execute read-only ClickHouse queries against the MCP-protected resource. Verbatim ChatGPT response to the prompt "test rauthy mcp server":

> rauthy-test-f0beab MCP server works.
> Authenticated as: btyshkevich@gmail.com
> Read-only query test: SELECT 1 AS ok → ok = 1

### #1559 — fix(ui): URL-encode `client.id` in admin path segments

[...]

PR #1605, opened 2026-06-14
fix(clients): restrict client id validation for non-ephemeral clients

bartekus:
Follows up on the discussion in #1559 / #1562. As discussed there (https://github.com/sebadob/rauthy/pull/1559#issuecomment-4689605960), the proper fix is server-side: the Admin UI already restricts client IDs to `^[a-zA-Z0-9._\-]{2,256}$` after #1572, but the API and bootstrapping still accept URI-shaped IDs via the minimal client requests. That lets a client into the DB that the Admin UI cannot manage and that behaves completely differently depending on whether ephemeral clients are enabled.

This restricts the manually managed (non-ephemeral) creation paths to the same pattern the UI uses, so a URI-shaped ID can no longer enter the DB through them.

### Changes

- New `RE_CLIENT_ID_STRICT` regex (`^[a-zA-Z0-9._\-]{2,256}$`), byte-identical to the UI's `PATTERN_CLIENT_ID_NEW`.
- Applied to `NewClientRequest.id` and the bootstrap `Client.id`.
- `EphemeralClientRequest` and every flow/lookup path keep using `RE_CLIENT_ID` (the full URI pattern), since ephemeral / CIMD / Solid clients legitimately use a URI as their ID.
- Removed the redundant `id` field from `UpdateClientRequest` (per review): the ID is always taken from the URL path on update and could never be changed, so the body field only fed a now-removed equality check. Existing clients with a legacy URI-shaped ID therefore stay manageable (updatable via their path) rather than being locked out.
- Test: creating a client with a URI-shaped ID returns `400`.
- CHANGELOG entry and a short note in the ephemeral-clients book page clarifying the ID distinction.

Thanks to @BorisTyshkevich for surfacing this in #1559; this takes the server-side route from that thread, which should make the Admin UI encoding change there unnecessary for new clients.

### DIFF
--- a/src/common/src/regex.rs
+++ b/src/common/src/regex.rs
@@ -21,6 +21,11 @@ pub static RE_CITY: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9À-ÿ-\s]{0,48}$").unwrap());
 pub static RE_CLIENT_ID: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9,.:/_\-&?=~#!$'()*+%]{2,256}$").unwrap());
+// Stricter pattern for manually managed (non-ephemeral) clients. Mirrors the admin UI's
+// `PATTERN_CLIENT_ID_NEW`: alphanumeric plus `.`, `_`, `-`, so a client id can never be a URI.
+// Ephemeral clients keep using `RE_CLIENT_ID`, which still allows full URI ids.
+pub static RE_CLIENT_ID_STRICT: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9._\-]{2,256}$").unwrap());
 pub static RE_CLIENT_NAME: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^[a-zA-Z0-9À-ɏ-\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{2,128}$").unwrap()
 });
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,25 @@ clients cannot set their own claims.
 
 [#1604](https://github.com/sebadob/rauthy/pull/1604)
 
+#### Stricter Client ID validation
+
+Client IDs for manually managed clients are now validated against the stricter pattern
+`^[a-zA-Z0-9._\-]{2,256}$`, which matches the existing Admin UI restriction. This applies when
+creating a client via the API and when bootstrapping clients from JSON files. Ephemeral clients are
+unaffected and may still use a full URI as their ID.
+
+Previously, the backend accepted URI-shaped IDs on these paths, even though such a client cannot be
+managed in the Admin UI and behaves completely differently depending on whether ephemeral clients
+are enabled. Existing clients with such an ID keep working and can still be updated, since the ID for
+an update is taken from the URL path; only creating or bootstrapping a new client with such an ID is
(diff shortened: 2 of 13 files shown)


bartekus:
Verified locally beyond the Code Style gate: the full `just test-hiqlite` suite passes, including `test_clients` with the new assertion that `POST /clients` with a URL-shaped id now returns `400`.

Worth noting the test config runs with `ephemeral_clients.enable = true`, so the rejection happens at request validation regardless of the ephemeral setting; it does not depend on the runtime toggle that made the original behaviour confusing.

src/common/src/regex.rs
```diff
@@ -21,6 +21,11 @@ pub static RE_CITY: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9À-ÿ-\s]{0,48}$").unwrap());
 pub static RE_CLIENT_ID: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9,.:/_\-&?=~#!$'()*+%]{2,256}$").unwrap());
+// Stricter pattern for manually managed (non-ephemeral) clients. Mirrors the admin UI's
+// `PATTERN_CLIENT_ID_NEW`: alphanumeric plus `.`, `_`, `-`, so a client id can never be a URI.
+// Ephemeral clients keep using `RE_CLIENT_ID`, which still allows full URI ids.
+pub static RE_CLIENT_ID_NEW: LazyLock<Regex> =
```
sebadob: I think the `_NEW` in the end should be changed. When you see it in code in other places it kind of seems like "I added a 'new' regex there for whatever reason", and it might be confusing. I would rather name it `_LIMITED` or `_STRICT`, which is probably a lot more clear for future maintainability.

sebadob (changes requested):

sebadob:
> UpdateClientRequest is included in the restriction. Since a client ID is immutable (update_client rejects a changed ID), this means an existing client that already has a URI-shaped (or otherwise special) ID can no longer be updated and would have to be deleted and recreated with a valid ID. If you would rather keep such legacy rows editable, UpdateClientRequest can stay on the looser RE_CLIENT_ID while create + bootstrap go strict; happy to adjust.

While thinking about it, we should probably fully remove the `id` from `UpdateClientRequest`. We don't even need that, and it's a bit of a left-over from a very old API version that worked differently. There is only a single check for that ID when updating a client, but in all situations, the ID is taken from the path anyway. We need to do some cleanup of such things in the near future on the road to a v1.0.0. I have a few other things in mind as well.

src/common/src/regex.rs
bartekus: Good point, renamed to `RE_CLIENT_ID_STRICT` and force-pushed.

bartekus:
Done. Removed `id` from `UpdateClientRequest` entirely; it only fed the old "id cannot be changed" check, and the id is taken from the path on every update anyway. So `RE_CLIENT_ID_STRICT` now applies only to create + bootstrap, and as a side effect an existing client with a legacy URL-shaped id stays manageable (updatable via its path) rather than being locked out. Updated the service, the request type, the admin UI, the tests, and the CHANGELOG; force-pushed.

Glad to help with the other v1.0.0 cleanups when you get to them.

sebadob:
> Glad to help with the other v1.0.0 cleanups when you get to them.

Thanks, but we need to collect them first. I know already that I will probably want to remove WebID and FedCM from the code. WebID never really got adopted, and Google basically killed FedCM when they decided to not make it possible to bring your own provider.

But I will probably open an issue about it, keep it open for a while, and just see if anyone is actually using it, and why. If there is some good rreason to keep it, that would be fine as well. I just want to go over everything once again and check what we can clean up before v1, so we don't have to maintain stuff we don't want or need anyway in the future.

---

After the merge of the other PR, the only thing left here is to resolve conflicts in the changelog.

sebadob:
~~Oh, one last thing I just saw: The `UpdateClientRequest` in `frontend/src/api/types/clients.ts` can have the `id` removed as well, and then the UI components using it.~~

Edit:

My bad, you did it already. Sorry didn't see it right away.

sebadob (approved):
Thanks!

PR #1606, opened 2026-06-14
feat(users): delegated group admins via rauthy_admin:<group> roles

### LINKED ISSUE #1538: User group management without `rauthy_admin`

The single `rauthy_admin` for multiple user groups doesn't particularly scale well. You can end up swimming in support requests. It would be good to delegate some user group management to a `user_admin` (vs. `rauthy_admin`). Some use cases:

* You disable registration but still want to allow manual account creation based on some out-of-band invite-only process. User-admins can self-organise this onboarding.
* Some users of a specific group pause their membership and the rest of the group want to stop them logging in. User-admins can flick the `enabled` switch.
* Less digitally literate users struggle to self-service their own profile and ask explicitly for someone to do this for them. User-admins can handle this.

Users can be elevated to an admin of a specific group. This gives access to the admin panel but not with full `rauthy_admin` role super powers. They have powers to adjust user profile information and create/delete user accounts for users which fall under the group they are admin for.

Single or multiple trusted admins. Hosting multiple rauthy instances.

**Additional context**
N/A.

bartekus:
## Summary

Implements delegated, group-scoped user administration as designed in #1538: a role
named `rauthy_admin:<group>` makes its holders a **group admin** who can manage only the
users of the matching group(s), without the full `rauthy_admin` role.

This follows the prefix scheme from your 2026-05-16 comment and the scope we agreed on in
the issue thread.

Closes #1538

## Role grammar

- `rauthy_admin` (no colon) is the full admin, unchanged.
- `rauthy_admin:<group>` matches the group `<group>` **exactly**.
- `rauthy_admin:<prefix>*` is a **prefix glob** (`rauthy_admin:eng*` matches `eng`,
  `engineering`, ...).
- `rauthy_admin:*` matches every group, i.e. a user-super-admin, with no special-case
  code (it falls straight out of the glob).

No new validation was added: the existing `RE_ROLES_SCOPES` regex already allows `:` and
`*`, so these are valid role names today.

## Non-breaking

This is non-breaking, which was the stated condition in the issue:

- `Principal::is_admin()` still does an exact `rauthy_admin` match; the full-admin and
  ApiKey code paths are byte-identical.
- The only risk is a pre-existing custom role literally matching `rauthy_admin:<...>`. For
  that, `temp_migrations.rs` now logs a `warn!` on each startup (for the whole `v0.36`
  cycle) listing any such role, so an operator gets a heads-up instead of a silent
  semantics change.
- Verified at the integration level: the existing `handler_users`, `handler_roles`,
  `handler_groups`, `handler_auth` and `handler_generic` suites all still pass unchanged.

## What a group admin can do / cannot do

All authorization lives in one place (`Principal`), never scattered string checks.

Can: see the (full, minified) user list; read any user; create users into a managed
group; edit profile / `enabled` / `email_verified`; add or remove memberships only for
managed groups; reset MFA; view and invalidate sessions of managed users. Gets a
read-only view of Sessions, Events and the Blacklist for debugging.

Cannot (enforced server-side, each covered by a test): manage any user that holds a
`rauthy_admin*` role (no escalation, group admins can't manage each other); change roles;
set passwords; delete users; touch groups outside its prefix; manage clients / scopes /
roles / groups / providers / API keys / config; write to the blacklist. Delegation is
session-only; ApiKeys are never group-scoped.

## See-all user list

As discussed, the user list is not filtered by group: the `users` table inlines
roles/groups for fast logins, so per-group filtering would be a slow regex scan. The
security boundary is enforced on every write (by the target user's groups), not on list
visibility. This also enables adding an existing user into a managed group.

## Backend

- `Principal` (`src/data/src/entity/principal.rs`): the grammar + guards
  (`is_group_admin`, `group_admin_can_manage`, `validate_user_admin_access`,
  `validate_api_key_or_admin_or_group_admin`, `validate_group_admin_user_change`,
  `validate_group_admin_user_create`) with 16 unit tests.
- Endpoints: `GET/PUT/PATCH /users`, `POST /users`, `GET /users/{id}`,
  `GET /users/email/{email}`, `DELETE /users/{id}/webauthn/...` (MFA reset),
  `GET /sessions`, `DELETE /sessions/{user_id}`, `GET /blacklist`, `POST /events` +
  `/events/stream`, and `/auth_check_admin`. Full user deletion, blacklist writes, single
  session/all-session deletion and event-test stay full-admin only.
- `temp_migrations.rs`: the non-breaking startup warning.

## Frontend

Reuses the existing admin UI, gated by one capability flag derived from the session info
the UI already has (`src/utils/adminScope.ts`):

- a group admin reaches the admin UI; the nav hides everything but Users, Sessions,
  Events, Blacklist and Docs;
- the user detail view drops the Attributes, Password, Devices and Delete tabs;
- in the edit and create forms, roles are read-only and only managed groups can be toggled
  (a small `disabledNames` option added to `SelectList`).

`svelte-check` and `prettier` are clean.

## Tests

- 16 `Principal` unit tests for the grammar and every escalation guard.
- A new `handler_group_admin` integration test that logs in as a real group admin and
  asserts the full matrix end-to-end (allowed actions 200, every escalation 403).
- Existing suites confirm the non-breaking full-admin behavior.

## Docs

- New Book page `work/group_admins.md` + `SUMMARY.md`.
- `CHANGELOG.md` entry.

## Deferred (per the thread, happy to add if you want any in this PR)

- the broader-prefix-manages-narrower admin hierarchy (v1 uses the strict-simple rule:
  group admins can't manage any `rauthy_admin*` user),
- a strict tenant-isolation (filtered-list) mode as an opt-in config,
- an "MFA reset by admin" notification email,
- group CRUD stays full-admin only (your call: clients can gate login by group prefix, so
  group editing could grant unintended access; renames are also expensive).


book/src/work/group_admins.md
```diff
@@ -0,0 +1,72 @@
...
+help with debugging user issues.
+
+## What a group admin cannot do
+
+The backend enforces all of the following, regardless of what the UI shows:
+
+- it cannot manage any user that holds a `rauthy_admin` or `rauthy_admin:<...>` role (no
+  privilege escalation, and group admins cannot manage each other),
+- it cannot modify a user's roles, or change group memberships outside of its managed groups,
+- it cannot set a user's password,
```
sebadob: I am currently thinking about it if this would helpful to allow. E.g. I know of a situation with someone using Rauthy, where people are actually not able to do the proper password reset (don't ask why ^^).
Since group admins cannot priv esc, they cannot update passwords of other admins. I think it would be good if they could not only trigger sending a new Password Reset E-Mail, but also maybe set a fixed password.

frontend/src/utils/adminScope.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export function parseRoles(roles: string | undefined): string[] {
+    return (roles ?? '')
+        .split(',')
+        .map(r => r.trim())
+        .filter(r => r.length > 0);
+}
+
+export function isFullAdmin(roles: string | undefined): boolean {
```
sebadob: It makes a lot of sense to add these helpers ti the `useSession()` hook into `frontend/src/global_state/session.svelte.ts`. This gets rid of additional imports everywhere and scattered logic. I would also rename `isFullAdmin` into just `isAdmin`. The "full" admin is the default and the GroupAdmin is specialized.

src/api/src/generic.rs
```diff
@@ -108,7 +108,26 @@ pub async fn get_auth_check(principal: ReqPrincipal) -> Result<HttpResponse, Err
 )]
 #[get("/auth_check_admin")]
 pub async fn get_auth_check_admin(principal: ReqPrincipal) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    if principal.is_admin() {
+        // unchanged full-admin behavior, incl. admin-MFA enforcement (-> 406)
```
sebadob: We can remove "unchanged" everywhere. In such comments as well as in the docs / book, and so on. "Unchanged" only makes sense in the context of this PR, but not when someone else reads it in the future.

src/api/src/users.rs
```diff
@@ -1443,34 +1449,47 @@ pub async fn delete_webauthn(
...
+    // Access tiers:
+    // - full Rauthy admin: may reset MFA for any user,
+    // - group admin: may reset MFA for a user it manages (see #1538),
+    // - the user itself: only with a valid `mfa_mod_token`.
+    if principal.validate_admin_session().is_ok() {
+        if principal.is_user(&id).is_err() {
+            warn!("Passkey delete from admin for user {} for key {}", id, name);
+        }
+    } else {
+        principal.validate_session_auth()?;
```
sebadob, line 1496: Group Admin checks are necessary in a lot of places, and there is some duplicated code here and there, especially when checking for force_mfa for admin access, and so on. It makes a lot of sense to add some dedicated functions to the `Principal` that check for properly authenticated Group admins. Basically, a clone of `.is_admin()` and `.validate_admin_session()`, just for group admins.

src/api/src/users.rs
```diff
@@ -1443,34 +1449,47 @@ pub async fn delete_webauthn(
...
+            let ip = real_ip_from_req(&req)?;
+            token.validate(principal.user_id()?, ip)?;
 
-        warn!("Passkey delete for user {} for key {}", id, name);
-    } else if principal.is_user(&id).is_err() {
-        warn!("Passkey delete from admin for user {} for key {}", id, name);
+            warn!("Passkey delete for user {} for key {}", id, name);
+        } else {
+            // a group admin resetting MFA for a user it manages
+            let target = User::find(id.clone()).await?;
```
sebadob, line 1515: When we have the additional checks in the session for making sure it's a group admin, and not just "any authenticated user", we can do this way more cheap check before fetching any additional data from the DB. Right now, any user could trigger additional queries with this, as long as the session is authenticated.

src/api/src/users.rs
```diff
@@ -1810,8 +1828,32 @@ pub async fn put_user_by_id(
     }
     .validate()?;
 
+    let id = id.into_inner();
+    let target = User::find(id.clone()).await?;
```
sebadob, line 1864: Here we want to have the same pre-checks as mentioned above. We should not even start fetching from the DB if the user is not allowed to do this in the first place.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +50,271 @@ impl Principal {
...
+    /// Order of precedence:
+    /// 1. a valid `ApiKey` with the required `access_group` / `access_rights`
+    ///    (unchanged: full power, never group-scoped),
+    /// 2. a full `rauthy_admin` session (unchanged: full power, admin-MFA enforced),
+    /// 3. a group-admin session, allowed only when [`Self::group_admin_can_manage`]
+    ///    returns `true` for the target user.
+    ///
+    /// `target_roles` / `target_groups` are the comma-split roles and groups of the
+    /// user being administered (e.g. `user.roles_iter()` / `user.groups_iter()`).
+    pub fn validate_user_admin_access<'a, R, G>(
```
sebadob: The name does not fit well. While this may be the situation right now, that it's only about users, this may change in the future. It should really be `validate_group_admin*` instead.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +50,271 @@ impl Principal {
...
+                ErrorResponseType::MfaRequired,
+                "Admin access only allowed with MFA active",
+            ));
+        }
+        if !self.is_group_admin() {
+            return Err(ErrorResponse::new(
+                ErrorResponseType::Forbidden,
+                "Rauthy admin or matching group admin access only",
+            ));
+        }
```
sebadob, line 170: I mentioned this above already, but the logic for checking if it's an authenticated group admin (with checking for MFA force) should be a separate function. This then should be called in endpoints before fetching any data to better protect against resource exhaustion attacks,

src/data/src/entity/principal.rs
```diff
@@ -49,6 +50,271 @@ impl Principal {
...
+        Ok(())
+    }
+
+    /// Like [`Self::validate_api_key_or_admin_session`], but also lets a delegated
+    /// group admin through. Used for endpoints without a single target user, where
+    /// per-user scoping is enforced separately on the individual write actions: the
+    /// user list (group admins see all users, see #1538) and the read-only
+    /// Sessions / Events / Blacklist views.
+    pub fn validate_api_key_or_admin_or_group_admin(
```
sebadob: I would rename this slightly to not have such a long name. Just call it `validate_api_key_or_group_admin`. If it's a real admin with a session, it can do anything all the time anyway.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +50,271 @@ impl Principal {
...
+        access_rights: AccessRights,
+    ) -> Result<(), ErrorResponse> {
+        match self.validate_api_key(access_group, access_rights) {
+            Ok(()) => return Ok(()),
+            Err(err) if err.error == ErrorResponseType::Forbidden => return Err(err),
+            Err(_) => {}
+        }
+        if self.is_admin() {
+            return self.validate_admin_session();
+        }
```
sebadob, line 197: I would move the `.is_admin()` to the top. This will be true in most cases. You should always check the most-likely cases first. The API Key will probably be the least likely with this additional feature.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +50,271 @@ impl Principal {
...
+            ));
+        }
+        for g in groups.iter().filter(|g| !g.is_empty()) {
+            if !self.group_admin_matches(g) {
+                return Err(ErrorResponse::new(
+                    ErrorResponseType::Forbidden,
+                    "Group admins can only assign groups they manage",
+                ));
+            }
+        }
```
sebadob, line 458: This groups validation is a bit weird. I don't see the reason for the `managed` check upfront and iterating twice. If you would just do the second iter and make sure only admin-managed groups are added, then you only need to do an `!groups.is_empty()` check afterwards to achieve the same as the first iter and check. Or even less expensive: return an error in the very beginning if the groups are empty / None, with the validation afterwards that only managed groups must be specified.

sebadob (changes requested):
Have you tested the UI? I's not really working.

I checked out your branch and the very first thing you see is that we need to allow reading all groups and roles for group admins. Without this information, the UI won't work and you will see the `Rauthy admin access only` error all the time everywhere.

Apart from that, the `get_user_by_id()` should not return the full information if the user is not managed by a group admin, but it does that right now. I created a group admin for testing and it can see all information no matter the groups membership.
A group admin could of course add a user to a manged group and see everything, but then it's obvious because of the membership. It makes sense for them to see the minimal version from the GET all users, but not the details.

---

The way an API error is displayed in the Admin UI right now is not sufficient anymore for group admins. This error should never be seen (if you don't screw up your reverse proxy or something like that), so it's kind of ugly. But when a group admin does not have access to a users details, this should rather be an overlay or similar.

This is the current behvaior:

Image

Also, instead of throwing an error in the UI, all buttons a group admin is not able to use anyway should be hidden. For instance the save button for all other admin users that are forbidden anyway (or maybe even hide them from the list in the UI dynamically).

The `<SelectList>` component also needs a few more updates. The button for editing should not even be shown if disabled, and the `x` for deleting a value should be hidden as well.

The next thing is that the `put("/users/{id}/self/preferred_username")` also needs an update in the backend and special handling in the UI. It should work in a way that a Group Admin can set a name if it's empty, but not overwrite an existing one. This should be reserved for a full admin, since it has many side effects. This needs special validation in the backend, and the UI component needs to behave differently and check this as well.

And the last one I saw so far is that the sessions page needs an update as well. The `Invalidate All Sessions` buttons should be hidden for Group Admins, and the 2 delete buttons for a specific session, I am not sure about them. We could allow session deletion for managed users, but we would not easily know in the UI which ones are managed. So I would maybe also hide these buttons for Group Admins. Deleting a session is more like an incident response action which would probably be done by someone higher in the chain. A group admin can always log out a user by the logout action from the user details tab. 

Image

When these buttons are hidden we also don't have to deal with the ugly error on that page.

This applies to the IP Blacklist page as well of course. Simply hide the add new button thtat's forbidden anyway.

bartekus:
Thanks, great review, all of it makes sense. I'll push an update to this branch covering everything:

- group admins get read access to roles + groups (that was the cause of the `Rauthy admin access only` errors), and `get_user_by_id` will only return full details for a managed user;
- hide forbidden affordances instead of erroring: the save button / unmanaged users, the `SelectList` edit pencil + the `x` when disabled, the session invalidate/delete buttons, and the blacklist add button;
- a dedicated authenticated-group-admin check on `Principal` (mirroring `validate_admin_session`, force-MFA included), called before any DB lookup, plus the renames you suggested and the `is_admin()`-first ordering;
- move the UI helpers into `useSession()` (`isFullAdmin` becomes `isAdmin`), simplify the create-group validation, and drop the "unchanged" wording everywhere;
- `preferred_username`: a group admin can set it when empty but not overwrite an existing one.

On passwords: agreed. I'll let a group admin trigger a reset and set a password for a managed user (they can't touch other admins anyway), and update the docs to match.

I'll ping you once it's pushed.

bartekus:
Pushed (`6895d8d3`). This addresses everything from the review:

- group admins can now read the roles + groups lists (that was the cause of the `Rauthy admin access only` errors), and `get_user_by_id` only returns the full details for a managed user;
- added `Principal::validate_group_admin_session()` and run it before any DB lookup; renamed the guards to `validate_api_key_or_group_admin` / `validate_group_admin_can_manage`, with `is_admin()` checked first, and simplified the create-time group validation;
- group admins can now reset or set a password for a managed user, and set (but not overwrite) an empty `preferred_username`;
- UI: hidden the affordances that are forbidden anyway (the `SelectList` edit button when read-only, the session invalidate / per-session delete buttons, the blacklist add / delete buttons), a clean notice instead of the raw error for a user a group admin cannot manage, the helpers moved into `useSession()` (`isFullAdmin` is now `isAdmin`), and the "unchanged" comments removed;
- book + CHANGELOG updated for the password change.

Verified locally: the group-admin integration test (including the new roles/groups read, managed-only detail, and password cases), the `Principal` unit tests, and the existing user / session / role / group / auth suites all pass.

bartekus:
Also pushed `3de67436`: the `preferred_username` UI now matches the backend rule (a group admin can set it while empty, but the editor and the force-overwrite option stay full-admin only).

I went back and tested the whole thing end to end in a real browser, logged in as an actual `rauthy_admin:test` user (real login, the login PoW solved by the wasm), and walked the admin UI. The browser console was clean on the admin pages: the only HTTP error was the expected `403 GET /users/{id}` when the group admin opens a user it does not manage, which is now handled by the notice below (plus the usual `events/stream` SSE abort on navigation). No JS errors.

**Reduced nav, full (minified) user list, `Add New` available:**

nav + user list

**Managed user, tabs reduced to Info / MFA / Logout (no Attributes, Password, Devices or Delete):**

managed user detail

**Opening a user outside the managed groups now shows a clean notice instead of the raw error:**

unmanaged user notice

**Blacklist is read-only (no `Add New`):**

blacklist read-only

**Sessions page (the `Invalidate All` and per-session delete buttons are hidden for group admins):**

sessions

bartekus:
Rebased onto current `main` (post-#1604 `custom claims on client_credentials`).

Small rebase: the only conflict was the `CHANGELOG.md` feature entry, resolved by keeping both (the #1604 claims entry and this one). `src/common/src/constants.rs` auto-merged cleanly. No migrations involved, no other code conflicts.

Re-verified green on both backends after the rebase: `handler_group_admin` passes on Hiqlite and on Postgres. Branch is now mergeable.

sebadob:
I took a 2nd look at the updates and the UI looks way better now. The only case we now need a solution for is when a Group Admin wants to add a user to one of the managed groups.

To solve this, I think we need to update the backend response for getting user details a bit. The current behavior is that when we receive a `403`, the 

```
You are not allowed to manage this user
```

is shown. This is much cleaner now than just showing the error previously. However, we need to distinguish between a user that has any `rauthy_admin` role, and one that is no admin, but just not in any managed group.

When the request for user details is coming from a group admin, and the user is not any `rauthy_admin`, and it's not in any managed group, we can solve this easily by returning another status code. In this case I think a `428` is most appropriate.

The UI can then show an option like "add user to my managed groups" or something like that. On click, show the groups input, and a save button. The actual group assignments cannot be seen there, so it would be slightly incorrect. User details can only be seen once the user is actually managed. However, this is not an issue at all. The UI is doing a `PATCH`, not a `PUT`, and all groups the Group Admin cannot managed cannot produce any conflicts.

The only thing that needs testing and probably a special solution is the "add to my groups" function. I think we cannot easily reuse the current `<UserInfo>` component, because it compares to the old values. But a non-managed user will not have any old values, because they are not returned at all. So this must be a very specific `PATCH` for the user, only for groups.

While thinking about it, the `PATCH` endpoint needs a tiny bit of special validation as well. It may reject the request when the user is not managed yet, but it would be after the patch.

---

The "Password" tab needs to be shown in the UI for Group Admins.

---

I think we should have 2 different types or error messages, and slightly update the current one. 

Instead of just saying "You are not allowed to manage this user", saying "You cannot manage another Admin user" is probably a bit more clear, and avoids additional questions to full admins.

The 2nd message should be matching the above-mentioned case, when you cannot manage this user yet, but you could, if you would add it to your groups.

My concern here is that I want to avoid unnecessary support questions from group admins to a full admin, because they will come for sure, especially for bigger deployments.

---

There is some mismatch in validation for MFA management for a Group Admin.

For instance, when I open the details for a managed user, and then switch over to the MFA tab, I see:

grafik

This is the `GET /auth/v1/users/{id}/webauthn` endpoint (probably also the `DELETE` then). The other user info view looks fine, and the logout works as expected as well.

---

Another thing I noticed was that a group admin cannot manage itself via the Admin UI now. This might be a bit inconvenient, because of the "is any admin" check. We could of course have special cases where we check "is admin admin or myself", but this will become messy pretty quickly I guess. We could solve this in a way that we add a new icon next to the logout in the nav. That one could redirect to the account dashboard. The dashboard then could have a link to the admin UI in the bottom corner, if there is any role starting with `rauthy_admin`.

As a link to the dashboard, we can use:

```html
<svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="size-6">
  <path stroke-linecap="round" stroke-linejoin="round" d="M17.982 18.725A7.488 7.488 0 0 0 12 15.75a7.488 7.488 0 0 0-5.982 2.975m11.963 0a9 9 0 1 0-11.963 0m11.963 0A8.966 8.966 0 0 1 12 21a8.966 8.966 0 0 1-5.982-2.275M15 9.75a3 3 0 1 1-6 0 3 3 0 0 1 6 0Z" />
</svg>
``` 

To navigate to the Admin UI from the dashboard, maybe reuse `<IconUserGroup>`?

This makes it easy for a Group Admin to self-manage without special handling in lots of other situations.

bartekus:
Thanks for the very thorough round 2! I've addressed every point and rebased onto current `main` (post-#1605); the branch is mergeable again and green locally.

### Rebase note
The only conflict was the CHANGELOG feature entry (kept both). While running the suite I noticed `main` itself currently does not compile its tests: the `put_client_claims` helper added in #1604 still sets the `id` field that #1605 removed from `UpdateClientRequest`. Since there is no integration-test CI workflow, this slipped through. I included the one-line fix in a separate commit (`fix(tests): drop removed id field ...`) so the suite compiles and I could run it; happy to drop that commit if you'd rather fix it on `main` directly.

### 1. Add an unmanaged user to my groups (the `428` case)
`get_user_by_id` now returns `428 Precondition Required` (new `ErrorResponseType::PreconditionRequired`) when a group admin opens an ordinary user that is not in any managed group yet. The UI catches the `428` and shows a dedicated "add to my groups" form (a small `<UserAddToGroups>` component, not `<UserInfo>`, exactly because there are no old values to compare against) that does a groups-only `PATCH`.

The `PATCH` handler got the special validation you mentioned:
* it allows the change if the user is manageable **before or after** it, so it can bring a previously unmanaged user into scope, and
* it scopes the submitted groups: memberships outside my scope (which I cannot see) are preserved instead of being clobbered, so the partial groups list "cannot produce any conflicts" as you put it.

add to my groups

### 2. Password tab for group admins
The `Password` tab is shown for group admins now (managed users). The backend already allows setting a password for a managed user.

password tab

### 3. Two distinct error messages
* An admin target is a plain `403` with the clearer message **"You cannot manage another Admin user"**.
* The not-managed-yet case is the `428` above, which surfaces the add-to-groups affordance instead of an error.

cannot manage another admin

### 4. MFA tab mismatch
`GET /users/{id}/webauthn` now allows a group admin that manages the target (mirroring the existing `DELETE`), so the MFA tab loads cleanly.

mfa tab

### 5. Group admin self-management
Added a user-circle icon next to logout in the nav that links to the account dashboard, and an **"Admin UI"** link on the dashboard for anyone holding a `rauthy_admin*` role (reusing `<IconUserGroup>`). A group admin may now always view its own account (the "is any admin" gate would otherwise reject it), so no messy per-endpoint special-casing is needed.

account dashboard with admin link

### Two extra bugs I hit while testing the UI (both fixed here)
Testing the full flow surfaced two issues that broke the group-admin UI before the above could even show:
* `GET /users/values_config` was full-admin only. Without it the Admin UI could not render the user detail view for a group admin at all (this is likely what produced part of the "admin only" errors you saw).
* `get_user_by_id` for a group admin's own id returned `403`, which broke the account dashboard with an infinite login redirect.

### Verification
* `handler_group_admin` passes on Hiqlite **and** Postgres, with new sub-cases for `428` vs `403`, self-view `200`, the webauthn `GET`, and the add-to-groups `PATCH` with the scope-preserving merge. Added unit tests for the new `Principal` methods too.
* `cargo fmt --check`, `cargo clippy -- -D warnings`, `npm run check`, `npm run format-check` all clean.
* All seven screenshots were captured end-to-end as a real group admin (`rauthy_admin:gatest`) via headless Chromium, console clean.

(Reduced view of the managed-user detail with the new tab set:)

managed user tabs

sebadob:
The way of adding users to a managed groups looks really good now. The password actions are working fine as well. So does the account dashboard switch.

I came across a few things while having another look:

---

It's a nice addition that the groups for the "add to my groups" action are filtered and only contain the ones that the admin can actually manage. It would be nice if the "Add New User" would behave in the same way, that it does not even show groups you cannot manage.

When modifying groups for an existing, already managed user, it makes sense to show all of them. So I would keep that as it is.

---

When I am a group admin, and try to modify my own values via the UI, which is allowed in the latest version, the UI throws an error about the anti-lockuout below the "Save" button:

```
Anti-Lockout Rule: rauthy_admin rule cannot be removed
```

If we want to keep allowing that from the Admin UI view, that check needs an update. 

---

I just noticed that we still have the buttons for sending E-Mails to all users and the E-Mail jobs in the view for group admins.

It would actually be nice to allow this for group admins as well. However, if we want this, we should probably only allow a "In Group" filter, and then of course only allow groups that are actually managed by this admin. The validation in the backend would be pretty simple, and it's only a single endpoint. The `GET` for email jobs would be nice to allow as well. That would be a very nice addition without much work. It only needs the group admin validation in the backend and should work right away.

---

The switching between Admin UI and Account Dashboard works fine, but UI would re-order the buttons. The discrepancy now becomes very obvious since you can switch very easily now. For the Account Dashboard, I would probably push the buttons for switching to Admin View and also the Logout buton both in the bottom left corner. Keeping the labels makes sense for the Account View. The location should match the one from the Admin UI though to improve the UX for Admins.

---

I would hide the "Test" event button for Group Admins from the live-events panel on the right. This is mostly helpful for making sure the whole notification pipeline works, which is a full admins task. Fixing any issues requires config adoptions, and the group admins would maybe not even have access to these notification channels.

---

Not sure when this happened, but the margin for the popup from the language selector is too big. It should start right above the selector:

grafik

---

Since we have multiple buttons now in the Admin UI Nav, I would make the Logout button the same red as the one from the account dashboard to make it a bit more clear.

---

I think we are almost there. When merged, this is definitely a super nice addition. I had this on my TODO since quite a while. This comment is only about the UI. I have not taken another look at the backend code, which I will do next.

frontend/src/i18n/admin/de.ts
```diff
@@ -519,6 +519,11 @@ export let I18nAdminDe: I18nAdmin = {
             werden als String / JSON Wert gehandhabt.`,
         forceLogout: `Sollen sämtliche, für diesen Benutzer existierenden Sessions invalidiert und 
             Refresh Tokens gelöscht werden?`,
+        groupAdmin: {
+            notManagedTitle: 'Nutzer außerhalb Ihrer Gruppen',
+            notManagedDesc: `Dieser Nutzer ist in keiner von Ihnen verwalteten Gruppe, daher bleiben die Details verborgen. Fügen Sie den Nutzer zu einer oder mehreren Ihrer Gruppen hinzu, um ihn zu verwalten. Mitgliedschaften außerhalb Ihrer Gruppen bleiben erhalten.`,
```
sebadob: Adding the manual line-breaks here as well for all the translations would be nice.

frontend/src/lib/admin/users/UserInfo.svelte
```diff
@@ -44,13 +44,17 @@
         roles,
         groups,
         providers,
+        fullAdmin = true,
```
sebadob: I like to always have `is` or `has` in front of booleans. Makes it more obvious what the type is.

frontend/src/lib/PreferredUsername.svelte
```diff
@@ -17,11 +17,15 @@
         preferred_username = $bindable(),
         config,
         isAdmin,
+        groupAdmin,
```
sebadob: Here the same as mentioned above as well. I always have `is` or `has` in front of booleans. Also, we should probably invert the logic for the other mentioned value to match this one. Instead of passing "fullAdmin", I would invert it and call it "isGroupAdmin" in the other component. Right now, we have sometimes the value for a full admin, and in other places for a group admin. Making it consistent and have "isGroupAdmin" everywhere is cleaner.

frontend/src/utils/adminScope.ts
```diff
@@ -0,0 +1,64 @@
...
+        return false;
+    }
+    return groupMatchers(roles).some(m => (m.glob ? name.startsWith(m.prefix) : name === m.prefix));
+}
+
+/// Subset of `allGroups` that this principal manages (drives the prefix-bounded
+/// group editor: group admins may only add / remove these).
+export function managedGroups(roles: string | undefined, allGroups: string[]): string[] {
+    return allGroups.filter(g => managesGroup(roles, g));
+}
```
sebadob: Did I see it correctly that these helpers are now only used inside the `frontend/src/global_state/session.svelte.ts`? If so, I would move them into the same file, since they are very closely related. The `useSession()` should be the interface we use to manage the component views. We can then remove the `export` from all of these to never get confused and try to use these directly.

src/api/src/generic.rs
```diff
@@ -108,7 +108,26 @@ pub async fn get_auth_check(principal: ReqPrincipal) -> Result<HttpResponse, Err
...
+                ErrorResponseType::MfaRequired,
+                "Admin access only allowed with MFA active",
+            ));
+        }
+        if !principal.is_group_admin() {
+            return Err(ErrorResponse::new(
+                ErrorResponseType::Forbidden,
+                "Rauthy admin or group admin access only",
+            ));
+        }
```
sebadob: Since you added the `is_group_admin()` and `validate_group_admin_session()` fns to the `Principal`, we can use these functions here directly. They do the same validations.

src/api/src/users.rs
```diff
@@ -2127,14 +2210,36 @@ pub async fn put_user_self_preferred_username(
...
+                    "Only an admin can 'force_overwrite' the 'preferred_username'",
+                ));
+            }
+        } else {
+            // a group admin may set a managed user's username, but only while it is still
+            // empty, and may never force-overwrite (reserved for full admins) (see #1538)
+            principal.validate_group_admin_session()?;
+            let target = User::find(id.clone()).await?;
+            principal.validate_group_admin_can_manage(target.roles_iter(), target.groups_iter())?;
+            if force_overwrite {
```
sebadob: We have a duplication for the `force_overwrite` check here. We can move it out in front of the `if`. Then, we can invert the logic to `if principal.validate_user_session(&id).is_err() {` to simplify and clean up this code section a bit.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +66,396 @@ impl Principal {
...
+    where
+        CG: IntoIterator<Item = &'a str>,
+    {
+        if self.api_key.is_some() || self.is_admin() {
+            return submitted.to_vec();
+        }
+        // keep every current membership the admin does not manage ...
+        let mut out: Vec<String> = current_groups
+            .into_iter()
+            .filter(|g| !g.is_empty() && !self.group_admin_matches(g))
```
sebadob, line 343: The `self.group_admin_matches(g)` already includes an `is_empty()` check.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +66,396 @@ impl Principal {
...
+        }
+
+        let current_roles: BTreeSet<&str> = current_roles
+            .into_iter()
+            .filter(|r| !r.is_empty())
+            .collect();
+        let new_roles: BTreeSet<&str> = new_roles
+            .iter()
+            .map(|r| r.as_str())
+            .filter(|r| !r.is_empty())
```
sebadob, line 384: Usually you `filter()` before `map()`, or you combine both into `filter_map()`. Filtering after mapping is usually more expensive. The same below for groups.

src/data/src/entity/principal.rs
```diff
@@ -49,6 +66,396 @@ impl Principal {
...
+                "Group admins cannot assign roles",
+            ));
+        }
+
+        // the new user must be placed into at least one group, so it stays in scope
+        let groups: Vec<&str> = new_groups
+            .map(|g| {
+                g.iter()
+                    .map(|s| s.as_str())
+                    .filter(|s| !s.is_empty())
```
sebadob, line 440: Tiny thing again, but `filter` before `map`, or combine them.

sebadob (changes requested):

sebadob:
I just noticed that the user picture API endpoints need to be working for group admins as well.

bartekus:
Thanks for round 3! Rebased onto current `main` (now past #1608 and #1609) and addressed every point. The temporary `fix(tests)` commit dropped out of the rebase, since that fix is already on `main`.

### Backend
- `get_auth_check_admin` now calls `Principal::validate_group_admin_session()` for the non-full-admin branch instead of repeating the session / MFA / role checks inline.
- `put_user_self_preferred_username`: hoisted the `force_overwrite` guard out of both branches and inverted the inner check to `if principal.validate_user_session(&id).is_err()`, so the group-admin path is the only remaining special case and the `force_overwrite` rejection is a single check.
- `validate_group_admin_user_change` / `validate_group_admin_user_create`: `filter` before `map` when building the role / group sets.
- One small pushback on the `is_empty()` in `group_admin_scoped_groups`: `group_admin_matches()` is indeed empty-safe, but this is the *inverse* of the match (`!group_admin_matches(g)`), and `User::groups_iter()` yields a single `""` for a user with no groups (exactly the add-to-groups case). Without the guard that phantom `""` would be preserved as a membership. I kept it, documented why, and added a regression test asserting `group_admin_scoped_groups([""], ["dev"]) == ["dev"]`. Happy to change it if you'd still prefer it gone.

### Frontend
- Moved the role-grammar helpers into `session.svelte.ts` as private functions and deleted the standalone `adminScope.ts`. `useSession()` is now the only interface; the one external user (`AccMain`, the account dashboard) reads `session.isAnyAdmin()`.
- Collapsed `groupAdmin` / `fullAdmin` into a single `isGroupAdmin` everywhere (`PreferredUsername`, `UserInfo`, `UserDetails`, `UserAddNew`, `NavSide`), with the `is` prefix on the booleans.
- Wrapped the long `notManagedDesc` string with manual line breaks in all 9 locales.

### Verification
`cargo fmt --check`, `cargo clippy -- -D warnings`, `npm run format-check` and `npm run check` are all clean, and the full Hiqlite integration suite passes (including `handler_group_admin` and the new `Principal` unit test).

(Unrelated housekeeping: the round-2 screenshots above were on a branch that never made it onto my fork, so they were 404ing. I re-hosted them on the working `pr1606-screenshots` branch and the links render again now.)

bartekus:
Since round 3 was mostly a frontend refactor (the `groupAdmin`/`fullAdmin` consolidation into a single `isGroupAdmin`, and moving the role-grammar helpers into `useSession()`), I re-verified the UI end to end as a real pure group admin (a `rauthy_admin:gatest` user with no full-admin role) in a headless browser. All gating still behaves correctly:

**Reduced nav** (the `{#if !isGroupAdmin}` gates): only the user-centric, read-only sections are shown, everything full-admin-only is hidden.

reduced nav

**Managed user**: the tab set is still reduced to Info / Password / MFA / Logout (no Attributes, Devices or Delete).

managed user tabs

**A user outside the managed groups** still surfaces the `428` "not managed" notice instead of a raw error, and the re-wrapped `notManagedDesc` renders cleanly.

not managed notice

**The account dashboard** still shows the **Admin UI** link, now resolved via `useSession().isAnyAdmin()`.

account admin link

sebadob:
> Since round 3 was mostly a frontend refactor

Yeah, but basically the whole comment https://github.com/sebadob/rauthy/pull/1606#issuecomment-4728349020 and https://github.com/sebadob/rauthy/pull/1606#issuecomment-4729810452 is missing.

These were things that I could not directly comment inside the code. Can you take a look at them? Apart from these, looks good so far now.

bartekus:
Thanks, and apologies: round 3 only picked up the inline code-review comments and I completely missed these two top-level ones. Both are addressed now (pushed `12985e68`).

## From [the round-3 UI comment](https://github.com/sebadob/rauthy/pull/1606#issuecomment-4728349020)

**1. "Add New User" group filter.** A group admin now only sees the groups it can manage on the create form, and the role editor is hidden entirely (it can assign none). Editing an existing user still shows all groups, as you asked.

add user, no roles, managed groups only

**2. Anti-lockout on self-edit.** Kept it allowed and fixed the check. The frontend anti-lockout now only fires for a *full* admin removing its own `rauthy_admin` role; a group admin never holds that exact role. The backend needed it too: `PUT` / `PATCH /users/{id}` skip the admin-target manage guard when the target is the caller itself. The role and group guards still run, so a group admin still cannot change its own roles or touch out-of-scope groups.

**3. Bulk E-Mail / E-Mail jobs.** Built the scoped version you described. A group admin may send, but the filter is locked to a single `in_group` for a group it manages (full admins keep the full set). `GET /email` is allowed but scoped: a group admin only sees jobs targeting a group it manages, and may cancel those. Everything else (no filter, `not_in_group`, role filters, unmanaged groups) is a `403`.

group admin email locked to In Group of a managed group

**4. Button ordering.** On the account dashboard, theme, language, the "Admin UI" switch and Logout now sit in one row in the bottom-left corner, in the same order as the Admin UI nav.

account dashboard bottom-left row

**5. "Test" event button.** Hidden for group admins in the live-events panel.

group admin events, no Test button

**6. Language selector popup margin.** Fixed. The open-upwards offset used a per-item factor smaller than the actual row height, so the popup crept over the selector as the language count grew. It now uses the real row height (~2.1rem), so the popup bottom sits just above the selector regardless of how many languages are configured.

**7. Logout red.** The Admin UI nav Logout icon is now the same red as the account dashboard one (visible in the reduced-nav shot below).

group admin reduced nav, red logout

## From [the picture comment](https://github.com/sebadob/rauthy/pull/1606#issuecomment-4729810452)

**8. User picture endpoints.** `GET` / `PUT` / `DELETE /users/{id}/picture*` now work for a group admin that manages the target user (mirroring the existing managed-user checks). A group admin still cannot touch the picture of an unmanaged user or another admin, so avatars of unmanaged users in the list fall back to initials, consistent with the rest of the delegation model. Happy to widen this to all pictures if you'd prefer.

## One side note
While adding the email integration test I hit a pre-existing quirk: an `EmailJob` uses its creation second as its primary key, so two jobs created within the same second collide (`500`). Not touched here (out of scope), just flagging it; happy to open a separate issue.

## Verification
- `cargo fmt --check`, `cargo clippy -- -D warnings`, `npm run format-check` and `npm run check` are all clean.
- `handler_group_admin` passes on Hiqlite with new sub-cases for self-edit (`PATCH` + `PUT`) and the scoped email gating / listing / cancel.
- The screenshots above were captured as a real group admin (a `rauthy_admin:gatest` user with no full-admin role) via a headless browser.

Book and CHANGELOG updated.

<details><summary>GA account dashboard, for completeness</summary>

group admin account dashboard

</details>

sebadob:
The `<LangSelector>` margin became worse, but I think the issue is because of the manual positioning under the hood when we are opening it to the top, and then the list internally is overflowing. The amount of margin depends on the elements, but that only works properly without a scrollbar.

grafik

The fix here should be done in another way and also in another issue. As mentioned, the component tries to calculate the offset dynamically. This is of course very brittle. The good news is, by now all major browsers support CSS anchor positioning. This means we should be able to very easily fix this for all cases without any manual calculations. We can switch this, but in another PR.

Apart from that, the UI parts are looking very good now. I think we caught all issues now. I will take another look at the backend separately.

frontend/src/lib/select_list/SelectList.svelte
```diff
@@ -11,10 +11,19 @@
...
     }: {
         items: SelectItem[];
         maxWidth?: string;
+        // names that cannot be toggled (rendered read-only). Used by the delegated
+        // group-admin UI (#1538) to lock roles and out-of-scope groups.
+        disabledNames?: string[];
+        // names that are not rendered at all (and never counted as selected). Used by the
+        // "Add New User" form (#1538) so a group admin does not even see groups it cannot
+        // manage, while editing an existing user still shows all of them.
+        hiddenNames?: string[];
```
sebadob: `hiddenNames` should not even exist. If you want to hide something here, don't pass it with the `items` in the first place. If there is a really good reason why you cannot filter the original input directly, create a specialized wrapper component.

I would also not name it `disabledNames`, but match the original name, and call it `disabledItems`.

Also, there is no advantage from even mentioning the issue or PR in the comment. We don't need these at all. This is a generic component. If we actually need a specialized version for the users, I would create a a new wrapper component that makes use the of the `<SelectList>`.

frontend/src/lib/select_list/SelectList.svelte
```diff
@@ -27,13 +36,20 @@
     let closeModal: undefined | (() => void) = $state();
 
     let compact = $derived(innerWidth && innerWidth < 800);
+    // hidden entries are not rendered and never count as selected (#1538)
```
sebadob: This comment does not add any value. The variable name says everything we need to know already.

frontend/src/lib/LangSelector.svelte
```diff
@@ -30,7 +30,13 @@
...
     );
 
-    let offsetTop = $derived(openTop ? `-${(languages?.length || 0) * 2 + 2}rem` : undefined);
+    // when opening upwards, shift the popup up by (its own height + the selector height) so its
+    // bottom sits just above the selector with a small constant gap. Each option row is ~2.1rem,
+    // so the per-item factor must match it for the gap to stay constant regardless of how many
+    // languages are configured; the constant covers the selector height plus the small gap. The
+    // previous `* 2 + 2` used a too-small per-item factor, so the popup crept over the selector
+    // as the language count grew (#1538).
+    let offsetTop = $derived(openTop ? `-${(languages?.length || 0) * 2.1 + 2.3}rem` : undefined);
```
sebadob: As mentioned in the other comment already, this actually made it worse on my machine. The issue came up when the list was overflowing, because it takes the amount of items into account. But, the container stops growing at some point. We should keep the original one here and fix it with probably CSS anchor positioning in another PR.

src/api/src/email.rs
```diff
@@ -42,6 +63,7 @@ pub async fn get_email_jobs(principal: ReqPrincipal) -> Result<HttpResponse, Err
 ///
 /// **Permissions**
 /// - rauthy_admin
+/// - group admin (only an `in_group` filter for a group it manages, #1538)
```
sebadob: We usually never want any references to issues as comments in code. They don't add any value, unless an issue is a very special edge case and you may need (a lot) more context to understand.

The PR shows up in so many comments by now. A global search and replace to remove them makes sense.

src/api/src/email.rs
```diff
@@ -58,7 +80,25 @@ pub async fn post_send_email(
     principal: ReqPrincipal,
     Json(mut payload): Json<EmailJobRequest>,
 ) -> Result<HttpResponse, ErrorResponse> {
-    principal.validate_admin_session()?;
+    // a delegated group admin may only send to a single group it manages; full admins keep
+    // the unrestricted filter set (#1538)
```
sebadob: This comment also only makes sense in the context of this PR, but not when you read the code. Keep it simple and only say

```
a delegated group admin may only send to a single group it manages
```

Of course a real admin can do everything. And when we say something like they "keep" access rights or something, it does not make sense as a code comment, only in the context of some PR. But we don't want or need PR references in code comments.

I would actually not even add any comment here at all. When you take a look at what the code does, it's very obvious.

sebadob (changes requested):
Despite my earlier comment about the UI, I found another issue with it.

When you are on the account dashboard and you render in smartphone view, the "Admin UI" link button shows up at the top in the wrong place. It also breaks the layout and makes the content collide with the buttoms on the bottom. 

Image

To fix this, either make it show up in the bottom, or move it into a Tab Bar in the same way as the Logout. The nicer solution would be to have both Admin UI and Logout buttons at the bottom, but that may also be a bit trickier with CSS, not sure.

bartekus:
Round 5 addressed, thanks for the thorough review.

### LangSelector popup positioning

I went with the proper fix rather than keeping `* 2 + 2`. The brittle offset math is gone; the popup now uses CSS anchor positioning. I kept the blast radius minimal: it is an opt-in `anchored` prop on the shared `Popover`, and only `LangSelector` opts in, so every other consumer keeps the existing manual-offset path untouched. The popup's bottom is pinned to the toggle button regardless of list height, so a scrolling language list can no longer creep over the selector.

You suggested a separate PR for this since `Popover` is shared. Given the opt-in design keeps it isolated I folded it in here, but I am happy to split it into a follow-up PR if you would rather keep this one scoped; just say the word.

Verified live across all three placements (login, admin nav, account dashboard): constant `0.4rem` gap, left-aligned, and height-independent (I forced the popover height down with the list scrolling and the gap held).

LangSelector popup anchored above the selector

### SelectList

Dropped the `hiddenNames` prop; the "Add New User" form now filters the group input directly in `UserAddNew`. Renamed `disabledNames` to `disabledItems`. Verified live as a delegated group admin: the create form shows no role editor and only the managed group.

Add New User as a group admin: groups filtered to the managed group, no role editor

### Comments

Stripped the bare `#1538` tokens from code comments, keeping the explanatory prose (edge cases and rationale). Collapsed the `post_send_email` permission comment to one line.

All green: `cargo fmt`, `cargo clippy -- -D warnings`, `prettier`, `svelte-check` (0/0).

sebadob (changes requested):
https://github.com/sebadob/rauthy/pull/1606#pullrequestreview-4531548360 was not addressed. The rest looks good now.

bartekus:
Good catch, thanks! Fixed in `618f53f8`.

I went with the "nicer" option you described: on the smartphone layout the **Admin UI** link and the **Logout** button now both live in a single control bar at the bottom, right next to the theme and language switches (same order as the Admin UI nav, so nothing jumps around when switching back and forth). The Admin UI link no longer renders at the top, and the bar is a normal in-flow row at the end of the flex column, so it can never overlap the scrollable content above it.

What changed in `AccMain.svelte`:
- Extracted the shared theme / language / Admin UI / Logout row into a `bottomControls` snippet that is reused by both the phone and the wide layouts.
- On phone, the controls are the last in-flow row of the column (the content area flex-grows and scrolls between the header and the bar), instead of being absolutely positioned overlays.
- Dropped Logout from the phone tab bar (it lives in the bottom bar now). As a side effect this also removes the duplicate Logout that previously appeared in the wide-compact layout (tab + button).

I also rebased the branch onto the latest `main` (`047b2819`, includes #1603 and #1610). `cargo`/`clippy`, `prettier` and `svelte-check` (0 errors / 0 warnings) are all green.

Smartphone view after the fix (group admin on the account dashboard):

smartphone account dashboard bottom bar

sebadob (approved):
Everything looks good now. I think we got all edge cases covered. 

Thank you!

bartekus:
I believe we did, and you are most welcome!
Thank you for all your ongoing work on this great project!

PR #1607, opened 2026-06-14
feat(helm): add official Helm chart

bartekus:
## Official Helm chart

Follows up on the offer in #131. This adds a Helm chart under `charts/rauthy`
that deploys Rauthy as a Hiqlite-backed Raft `StatefulSet`.

It is built around the constraints from this issue and from the review on #1131:

- **Config is a file, not re-templated.** You provide the complete `config.toml`
  (the Reference Config or the Minimal Production Config) as a Secret and the
  chart mounts it verbatim at `/app/config.toml`. The chart does not duplicate
  any config value into `values.yaml`, so there is nothing to keep in sync when
  config keys move or change. This is the main thing you said you wanted to
  avoid, and it is the whole design of the chart.
- **The chart owns only the Raft topology.** The single thing it injects over
  your config is the cluster membership: `HQL_NODE_ID_FROM=k8s` and `HQL_NODES`,
  derived from `replicas`. Those cannot live in a static file because they
  depend on the StatefulSet and the install-time replica count.
- **No autoscaling.** There is no HorizontalPodAutoscaler anywhere, not even a
  gated-off template. Scaling is documented as up-only in both `values.yaml` and
  the chart README, because a removed Raft member stays in every other node's
  config.
- **No generated secrets / no cargo-cult defaults.** The chart never invents
  `secret_raft`, `secret_api` or encryption keys, and `values.yaml` carries no
  defaults for values that are hardcoded or live in your config. Rendering fails
  closed if no config is provided.

The chart is essentially a Helmification of the manifests already documented in
`book/src/getting_started/k8s.md`: same hardened securityContext (non-root, drop
ALL caps), the readiness probe on the Hiqlite `/ready` port, the liveness probe
on `/auth/v1/health`, a headless service with `publishNotReadyAddresses`, a
per-pod PVC, and (for `replicas > 1`) a PodDisruptionBudget plus a preferred pod
anti-affinity. That book page now points at the chart.

### Prior art

@wt's MVP in #1131 is the right starting shape (StatefulSet, headless service,
ingress) and I leaned on it. The two points you raised there carry into this
design: the HPA is gone entirely, and there are no default variables for values
that are hardcoded anyway. The difference is that this chart consumes the config
as a file instead of generating it from values.

### Verification

- `helm lint` is clean; `helm template` renders correctly at 1 and 3 replicas
  (including the newline-separated `HQL_NODES` that Hiqlite parses with
  `.lines()`, with pod ordinal `i` mapped to node id `i+1`).
- Verified on a `kind` cluster with `ghcr.io/sebadob/rauthy:0.35.2`: a fresh
  single-node install boots, forms the Raft cluster, passes the readiness probe
  on `/ready`, and passes the bundled `helm test` health check.
- The second commit adds a self-contained `charts/**` CI workflow that
  additionally exercises the `1 -> 3` scale-up and a restart re-join on kind. It
  is a separate commit so you can drop it if you do not want chart CI.

### Notes / questions

- Home: I put it in-tree under `charts/`, matching #1131; happy to move it to a
  separate `rauthy-charts` repo with a chart-releaser index instead if you
  prefer that for releases.
- `appVersion` is pinned to `0.35.2` and is also the default image tag, so the
  two never drift; bump both together on a release.
- One small thing worth noting for the docs: a minimal config also needs
  `webauthn.rp_id` / `rp_origin` to boot, so I included them in the chart's
  `files/config.toml.example`.


sebadob:
Thanks already. I think we should delay this slightly until we have all the secrets loading from files, which would be the preferred setup then.

bartekus:
Completely agree, that's the right call. Once secrets load from files, the chart gets simpler and safer in exactly the place it needs to: the config Secret stops being the one spot where sensitive values still sit inline, and the file-based path becomes the natural default to document and lead with.

I'll be honest about part of the motivation for opening this now rather than waiting: it's partly a conversation enhancer. The chart is shaped around the way I actually run Rauthy, so putting it out in the open surfaces that setup for others to see, critique, and challenge, which feels more useful than keeping it private until everything lines up perfectly.

Happy to let it sit until the file-based secrets land, then come back and reshape the chart so it leads with that approach.

bartekus:
Reshaped to lead with the file-based secrets path now that #1623 landed in 0.36.0, as discussed when we benched this.

The chart is now a two-object split:

- `config.toml` is non-secret and mounted from a ConfigMap; every sensitive value is a `"$SECRETS"` placeholder.
- `secrets.toml` holds only the secrets and is mounted from a Secret. It is optional: leave it out and supply the secrets via env vars (`extraEnv`) or your own external Secret, and the chart creates nothing.

So no sensitive value sits in the config object anymore. Bumped `appVersion` to 0.36.0 (first release with `$SECRETS`), rebased the branch onto current main, and updated the chart CI to install on kind via the new split (fresh install + 1 -> 3 scale-up + restart all green, plus lint/template). The book's k8s page is updated to match.

Ready for another look whenever you have a moment.

PR #1608, opened 2026-06-15
feat(oidc): RFC 8707 resource indicators

bartekus:
## feat(oidc): RFC 8707 Resource Indicators

Implements [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707) resource indicators, designed and
agreed in #1562 (your ack: [#issuecomment-4689747298](https://github.com/sebadob/rauthy/issues/1562#issuecomment-4689747298),
scope confirmation: [#issuecomment-4703047022](https://github.com/sebadob/rauthy/issues/1562#issuecomment-4703047022)).

Clients can send a `resource` parameter to request an audience-restricted access token, so a token
minted for one resource server can no longer be replayed against another. This is the missing piece
for Rauthy as an authorization server for a fleet of resource servers (for example MCP servers).

### What it does

- `resource` is accepted on the authorization request (`GET /oidc/authorize`, carried through the
  auth code) and the token request (`POST /oidc/token`) for the `authorization_code`,
  `client_credentials`, and `refresh_token` grants.
- The granted resource is added to the access token `aud`. On `refresh_token` the resource is
  carried inside the refresh token and reused, so a refreshed token keeps its binding and can never
  widen it.
- A malformed / unknown / disallowed resource returns the RFC error code `invalid_target`.

### Design decisions (per the #1562 thread)

- **Two per-client CSV columns** (both nullable, mirroring `allowed_origins`):
  - `allowed_resources`: the allow-list a requested `resource` is validated against. Empty means
    deny (`invalid_target`), so a client must explicitly opt in to a resource.
  - `default_aud`: audiences always added to this client's tokens, independent of any request. This
    came out of #1590 and covers less-capable / IoT clients that cannot send a `resource`.
- **Single `resource`** per request for v1 (the actix `serde_urlencoded` extractor does not fold
  repeated keys into a `Vec`); multi-value can follow.
- **Additive `aud`**: final `aud` = `client.id` + `default_aud` + granted `resource`, de-duplicated.
- **`aud` is now a string-or-array**: a single audience stays a string (no change for existing
  consumers), two or more become a JSON array (valid per RFC 7519). This also fixes the `solid_aud`
  case, which previously emitted a string that merely looked like an array.
- **Ephemeral clients deny by default**: an ephemeral client document may declare its own
  `allowed_resources`; absent that, a requested `resource` is rejected unless the operator sets the
  new `ephemeral_clients.allow_unvalidated_resource`. Deny-by-default avoids a trivial priv-esc.

### Scope boundaries (v1)

The `password` and device-code grants, and logins brokered through an upstream identity provider,
do not process the `resource` parameter yet. Introspection reflects the restricted `aud`
automatically.

### Implementation notes

- New per-client columns with migrations for both backends (`migrations/hiqlite/30_*`,
  `migrations/postgres/V25__*`). The auth code carries the chosen resource in its existing
  cache entry, so no migration was needed there.
- New `ErrorResponseType::InvalidTarget`, serialized as the RFC code `invalid_target`.
- Admin UI fields for both columns (`ClientConfig.svelte`) with i18n across all admin locales, and
  the login page forwards a `resource` query parameter into the login request.
- `config.toml` and the Rauthy Book gained a "Resource Indicators (RFC 8707)" section; CHANGELOG
  updated.

### Tests

`src/bin/tests/zzf_handler_resource_indicators.rs` covers the `client_credentials` grant
end-to-end: the multi-valued `aud` array, `default_aud` always-added, an allowed `resource` landing
in `aud`, a disallowed `resource` returning `invalid_target`, and deny-by-default with no
allow-list.

### Test evidence

- `cargo check --workspace --tests`, `cargo clippy --workspace --tests`, and `cargo fmt --check`
  are clean.
- Frontend `prettier -c .` passes and `svelte-check` reports no new problems.
- `just test-hiqlite`: green, including the new `test_resource_indicators`.
- `just test-postgres`: green, including `test_resource_indicators`. The new `V25` migration
  applies cleanly. (Note: running the postgres suite against current `main` first requires the
  in-flight `v24`→`V24` migration-prefix fix; otherwise `main`'s lowercase `v24` is skipped by the
  postgres migrator and the dev-data bootstrap fails before any test runs. That is unrelated to this
  PR.)

Refs #1562


book/src/work/resource_indicators.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Ephemeral clients
+
+An ephemeral client document may declare its own `allowed_resources`, which is validated the same
+way as for a normal client. If it declares none, a requested `resource` is rejected unless the
+operator opts in with:
+
+```toml
+[ephemeral_clients]
+allow_unvalidated_resource = true
```
sebadob: I am thinking if we should maybe even put a `danger_` in front, because it can lead to very easy priv esc if this is set without really knowing what you're doing. And maybe also mention this in the comment for the value, that it should only be set with a very good reason.

src/data/src/entity/clients.rs
```diff
@@ -42,8 +42,9 @@ SET name = $1, enabled = $2, confidential = $3, secret = $4, secret_kid = $5, re
     post_logout_redirect_uris = $7, allowed_origins = $8, flows_enabled = $9, access_token_alg = $10,
     id_token_alg = $11, auth_code_lifetime = $12, access_token_lifetime = $13, scopes = $14,
     default_scopes = $15, challenge = $16, force_mfa= $17, client_uri = $18, contacts = $19,
-    backchannel_logout_uri = $20, restrict_group_prefix = $21
-WHERE id = $22"#;
+    backchannel_logout_uri = $20, restrict_group_prefix = $21, allowed_resources = $22,
+    default_aud = $23
```
sebadob: Depending on what's merged first, this will clash with #1604 then.

It also just came to my mind that the new values for all the new migrations also need to be handled in `src/data/src/migration/db_migrate.rs` and  `src/data/src/migration/inserts.rs`.

src/data/src/entity/clients.rs
```diff
@@ -880,6 +907,77 @@ impl Client {
...
+        Some(self.default_aud_iter().map(String::from).collect())
+    }
+
+    /// Validates an RFC 8707 `resource` request value against this client's policy: the
+    /// value must be an absolute URI without a fragment and must be covered by the
+    /// client's `allowed_resources`. Ephemeral clients without their own allow-list are
+    /// only permitted when `ephemeral_clients.allow_unvalidated_resource` is enabled.
+    /// On any failure an `invalid_target` error (RFC 8707 §2) is returned.
+    pub fn validate_resource_request(&self, resource: &str) -> Result<(), ErrorResponse> {
+        if resource.is_empty() || !resource.contains("://") || resource.contains('#') {
```
sebadob: Even if the RFC says it must be a URI, we can drop this and allow anything? If have not looked into it yet, but why must these be URIs instead of "anything that's configured to be allowed"?

src/service/src/oidc/grant_types/device_code.rs
```diff
@@ -163,6 +163,8 @@ pub async fn grant_type_device_code(peer_ip: IpAddr, payload: TokenRequest) -> H
             code.nonce.map(TokenNonce),
             code.scopes.map(TokenScopes),
             None,
+            // device flow does not support resource indicators in v1
```
sebadob: The `in v1` in comments make only sense in the context of this PR, but not as a general code comment. Better say `not yet` or even remove the comment completely.

src/service/src/oidc/authorize.rs
```diff
@@ -183,6 +184,8 @@ pub async fn post_authorize_refresh(
             nonce: req_data.nonce,
             code_challenge: req_data.code_challenge,
             code_challenge_method: req_data.code_challenge_method,
+            // resource indicators are not re-sent on a session refresh
```
sebadob: Not really helpful comment. You can see directly that they are not included. Better say why this is the case. A comment explaining what the code does is usually a bad thing. Instead, explain why it does something.

src/service/src/oidc/grant_types/client_credentials.rs
```diff
@@ -60,7 +60,14 @@ pub async fn grant_type_credentials(
         ClientDyn::update_used(&client.id).await?;
     }
 
-    let ts = TokenSet::for_client_credentials(&client, dpop_fingerprint).await?;
+    // RFC 8707: validate the requested resource against the client's allow-list
+    if let Some(resource) = req_data.resource.as_deref() {
```
sebadob, line 63: Simply removing this comment is better. You can derive it from the function name directly what it does. We never want to comment what code does, only why. Can be removed in `authorize.rs` as well.

src/service/src/token_set.rs
```diff
@@ -262,10 +280,16 @@ impl TokenSet {
         } else {
             JwtAmrValue::Pwd.as_str()
         };
+        // Solid-OIDC ephemeral clients additionally carry the `solid` audience. With the
+        // typed `aud` this is now emitted as a real JSON array instead of a string that
```
sebadob: This is "now" emiited as ...

again only makes sense in the context of this PR, not as a general code comment.

config.toml
```diff
@@ -1420,6 +1420,14 @@ allowed_scopes = ['openid', 'profile', 'email', 'webid']
 # overwritten by: EPHEMERAL_CLIENTS_CACHE_LIFETIME
 cache_lifetime = 3600
 
+# RFC 8707: when an ephemeral client document declares no `allowed_resources`,
+# a requested `resource` is rejected by default. Set this to `true` to instead
+# allow such clients to request any (well-formed) resource indicator.
```
sebadob: Mentioned this above already. I think we should add a short sentence here saying that you should only enable this is you know what you are doing and have a good reason to do this. This can lead to priv esc easily if not handled with care.

sebadob (changes requested):

bartekus:
Thanks for the review. Addressed all points:

- **`db_migrate.rs` / `inserts.rs`:** good catch, `inserts::clients` now writes `allowed_resources`
  and `default_aud` too. `db_migrate.rs` reads via `SELECT *` and writes through `inserts::clients`,
  so it is covered by the same change.
- **`danger_` prefix:** renamed to `ephemeral_clients.danger_allow_unvalidated_resource`, and added
  a clear "only enable this if you know what you're doing, it can lead to easy priv-esc" note to the
  config comment, the Book, and the field doc.
- **URI validation:** dropped. The requested `resource` is now matched verbatim against the client's
  `allowed_resources`, so the operator decides what a valid value looks like. The RFC recommends an
  absolute URI, but there is no real reason for Rauthy to enforce it rather than just matching the
  configured allow-list.
- **Comments:** removed the "what" comments (client_credentials / authorize) and reworded the
  PR-context ones ("in v1", "now emitted ...") to describe the behaviour or the reason, or dropped
  them. Thanks, agree on why-not-what.
- **Migration clash with #1604:** noted. Whichever of the two lands first, I will rebase and
  renumber the `V25` / `30_` migrations on the other.

Still green on both backends (`just test-hiqlite` and `just test-postgres`, the new
`test_resource_indicators` included).

bartekus:
Rebased onto current `main` (post-#1604 `custom claims on client_credentials`).

Since #1604 also adds per-client columns, the client stack conflicted. Resolution keeps both feature sets side by side (`claims`/`claims_at_root` from #1604 and `allowed_resources`/`default_aud` here) across the `Client` entity, all `SQL_SAVE` bindings, the request/response types, the admin UI and the test fixtures. The `create` INSERT keeps only the resource columns, matching #1604's design where `claims` are set via PUT.

#1604 also took the `V25`/`30_` migration slots for `client_claims`, so the resource-indicator migrations were renumbered to `V26__client_resource_indicators.sql` (Postgres) and `31_client_resource_indicators.sql` (Hiqlite).

Re-verified green on both backends after the rebase: the targeted client, custom-claims and resource-indicator integration suites pass on Hiqlite and on Postgres. Branch is now mergeable.

bartekus:
Rebased onto the latest `main` to resolve the merge conflict.

The only conflict was in `CHANGELOG.md`: #1605 landed its `Stricter Client ID validation` entry in the same spot, so I kept both sections (order #1604, #1605, then RFC 8707). All client code auto-merged cleanly.

One follow-on fixup: #1605 removed the `id` field from `UpdateClientRequest`, so I dropped the now-stale `id` assignments from the two affected test setups (`zzf_handler_resource_indicators` and the `zzd_handler_clients` update helper). On update the id is taken from the URL path anyway, so behaviour is unchanged.

Verified on the rebased tip (`56f3834a`): `cargo fmt --check`, `cargo clippy -- -D warnings`, and the full hiqlite integration suite all pass. The RFC 8707 runtime logic is byte-identical to before the rebase.

The PR is mergeable again and ready for another look whenever you have time. Thanks!

sebadob (approved):
Thanks!

PR #1609, opened 2026-06-16
feat(api-keys): allow API keys to manage API keys

caniko:
## feat(api-keys): allow API keys to manage API keys

Adds a new `AccessGroup::ApiKeys` variant and gates the API key CRUD endpoints
with `validate_api_key_or_admin_session()` instead of requiring a full admin session.

This is part of the work towards https://github.com/sebadob/rauthy/issues/1584
(RFC: advanced bootstrap config). The generated-secret container that this builds
on was introduced in https://github.com/sebadob/rauthy/pull/1599.

### Changes

- **api_types**: Add `AccessGroup::ApiKeys` variant (appended to end of enum
  to preserve bincode variant indices)
- **entity**: Same variant + all `From` conversions
- **handlers**: `validate_admin_session()` → `validate_api_key_or_admin_session()`
  on all 5 API key endpoints (GET/POST/PUT/DELETE + secret rotation)
- **tests**: Full coverage — verifies isolation (groups-only key cannot manage
  API keys) and delegation (a key with `ApiKeys` access can CRUD other keys)

### DIFF
--- a/frontend/src/api/types/api_keys.ts
+++ b/frontend/src/api/types/api_keys.ts
@@ -11,7 +11,8 @@ export type AccessGroup =
     | 'Scopes'
     | 'UserAttributes'
     | 'Users'
-    | 'AuthProviders';
+    | 'AuthProviders'
+    | 'ApiKeys';
 export type AccessRight = 'read' | 'create' | 'update' | 'delete';
 
 export interface ApiKeyAccess {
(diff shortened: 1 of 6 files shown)


sebadob (changes requested):
Changes look good so far, but the UI parts are missing. Check `frontend/src/lib/admin/api_keys/ApiKeys.svelte` and the linked components. Not sure right now, but maybe it's only `frontend/src/lib/admin/api_keys/ApiKeyMatrix.svelte` that needs an update.

sebadob (approved):
Thanks!

PR #1610, opened 2026-06-16
feat(bootstrap): generated API key tokens + idempotent reconciliation

caniko:
## feat(bootstrap): generated API key tokens

Builds on PR A (RBAC for API keys). Finishes the bootstrap API key story from
https://github.com/sebadob/rauthy/issues/1584 by refining how `"secret": "generate"` works.

The generated-secret container and `rauthy bootstrap get/purge` CLI landed in
https://github.com/sebadob/rauthy/pull/1599. Generated secrets are written to the
encrypted container first, then the DB row is inserted — no token loss window.

### Changes

- **bootstrap/api_key**: `"secret": "generate"` writes the plaintext token to the
  encrypted container first, then hashes and inserts the DB row (matching the same
  pattern used by users and clients)
- **entity**: `insert()` takes `self` instead of a named parameter. Removed
  `PreparedApiKey` / `prepare()` / `insert_prepared()`. Made `cache_idx()` pub.
- **api_key_exists()**: replaced `find_all()` + filter with direct `find().is_ok()`
- **database.rs**: removed per-startup `reconcile_declared_api_keys()` (bootstrap
  already gated by the first-boot check in `migrate_init_prod`)
- **set_api_key_secret()**: uses `ApiKeyEntity::cache_idx()` instead of manually
  building the cache index string
- **bootstrap/types**: `Generate` doc updated
- **docs**: extraction via `rauthy bootstrap get --field token`,
  env var `RAUTHY_BOOTSTRAP_API_KEY_<NAME>_TOKEN`
- **tests**: roundtrip + failure-mode for container writes


sebadob:
FYI CI is not passing, so a review does not make too much sense before the pipeline is green.

caniko:
Thanks! I didn't have time to fix it; tomorrow

migrations/postgres/V24__cust_attrs_token_root.sql
```diff
@@ -1,2 +1,2 @@
 ALTER TABLE scopes
-    ADD claims_at_root bool DEFAULT false NOT NULL;
+    ADD COLUMN IF NOT EXISTS claims_at_root bool DEFAULT false NOT NULL;
```
sebadob: We should definitely remove the `IF NOT EXISTS` here. If the database is in a different state than we expect, we cannot guarantee consistency. The DB was modified manually by something else, which should never happen. When this migration was not applied already, there must not be any `claims_at_root` column.

Even worse here, this is a "fix" for only half of the DB choices. It's simple: the column must not exist when this migration is being applied. If anything else is the case, something has messed with the DB, and we cannot even know if the other migrations are still "true" or if earlier ones have been messed with as well. This could even lead to serious security issues, depending on what and how the DB was modified.

src/data/src/migration/bootstrap/api_key.rs
```diff
@@ -60,25 +59,78 @@ async fn bootstrap_api_keys_json() -> Result<(), ErrorResponse> {
...
 
     Ok(())
 }
 
+async fn create_or_update_api_key_access(
+    name: String,
+    expires: Option<i64>,
+    access: Vec<crate::entity::api_keys::ApiKeyAccess>,
+) -> Result<(), ErrorResponse> {
+    if api_key_exists(&name).await? {
```
sebadob: I don't get that function. Why outsourcing it into a separate function, and then using a `find_all()` and manually filtering for it, when you could simply use 

```rust
if ApiKeyEntity::find(&name).await.ok()
```

src/data/src/database.rs
```diff
@@ -244,6 +244,7 @@ impl DB {
         let config = RauthyConfig::get();
         if !config.vars.dev.dev_mode && config.is_primary_node {
             bootstrap::migrate_init_prod().await?;
+            bootstrap::reconcile_declared_api_keys().await?;
```
sebadob: I don't see why you would do that at this point. This means the whole API key bootstrapping logic, which will only ever run once for a deployment, will run at each single startup.

book/src/config/bootstrap.md
```diff
@@ -313,10 +336,10 @@ Users can request a generated password with `"password": "generate"`:
 Rauthy writes the plaintext password to the encrypted container before hashing it and inserting the
 user row. The database stores only the password hash.
 
-API keys in `api_keys.json` can request a generated secret with `"secret": "generate"`. Rauthy
-writes the plaintext API-key secret to the encrypted container before hashing and storing it in the
-database. Extract it with `--kind api-key --id <key-name> --field secret`; combine it with the key
-name as `API-Key <key-name>$<secret>` for the `Authorization` header.
+API keys in `api_keys.json` can request a generated token with `"secret": "generate"`. The full
+usable `name$secret` token is written to the encrypted container before the DB row is inserted.
```
sebadob: > The full usable `name$secret` token is written to the encrypted container before the DB row is inserted.

Actually, no one cares about the internals. They don't care about an order or internal logic for such processes. I would only mention that it's written to the container, that's it. That also makes sure, that whenever you update the internal logic, your docs are still valid.

src/data/src/entity/api_keys.rs
```diff
@@ -22,6 +22,11 @@ pub struct ApiKeyEntity {
     pub access: Vec<u8>,
 }
 
+pub struct PreparedApiKey {
```
sebadob: That's a bad naming. When you have multiple versions of something like an "ApiKey", you want to have the "ApiKey" first and then follow with a more descriptive wording, like "ApiKeyEntity" or "ApiKeyPrepared". That makes maintaining, understanding and managing code a lot easier. Also, in lots of places where you have e.g. automatic sorting like in the API documentation, all your "ApiKey"s show up in a one place next to each other instead of being scattered all over the place. 

This is always the case, not just here. Always have similar objects start with the same general name and be more precise at the end of the name, never at the front. Makes your life so much easier down the road, and improves the UX in lots of situations as well.

sebadob (changes requested):
Actually, I don't really see the point in doing all of this and over-complicating a simple bootstrap. If anything would go wrong during the bootstrapping process, there is no real (clean) way to recover from this anyway. Of course you can try to catch all possible situations, then try to figure out where something went wrong, and have more complicated code, but all of that for a process that runs exactly once only for a fresh deployment. 

And, if you start doing that for API keys, you also would need to catch all other generated secrets, and your code would blow up quite a bit for such a tiny edge case.
For instance, generated user passwords are hashed as well and they could get lost.

The absolute best solution in such a case is to simply clean the volume and restart a fresh instance. You would not even lose anything because your instance was never live anyway.

These changes are actually inconsistent. They only care about API Keys. If you would want to tackle this in a better way, save the API key secret (or better all of them) inside the container in plain text. Than you can write the file first, and the DB afterwards. This makes you never lose anything in such an edge case. The container is encrypted anyway, so writing plain text secrets inside it is no issue at all. It just makes everything a lot simpler. The same would be true for auto-generated user passwords.

caniko:
**Simplified per review feedback.**

Removed `PreparedApiKey` / `prepare()` / `insert_prepared()` — the generate-write-insert is now inlined directly in the bootstrap code, matching the same simple pattern already used by users and clients.

Also addressed the other review points:
- Removed per-startup `reconcile_declared_api_keys()` from `database.rs` (bootstrap already gated by the first-boot check)
- Replaced `api_key_exists()` full-table scan with direct `find().is_ok()`
- Removed `IF NOT EXISTS` from `V24__cust_attrs_token_root.sql`
- Simplified docs/comments — dropped internal ordering details

src/data/src/entity/api_keys.rs
```diff
@@ -33,23 +33,7 @@ impl Debug for ApiKeyEntity {
...
-            .into_bytes()
-            .to_vec();
-
-        let access_bytes = serialize(&access)?;
-        let access_enc = EncValue::encrypt(&access_bytes)?.into_bytes().to_vec();
-
-        let enc_key_active = &EncKeys::get_static().enc_key_active;
-        let secret_fmt = format!("{name}${secret_plain}");
-
+    pub async fn insert(entity: ApiKeyEntity) -> Result<(), ErrorResponse> {
```
sebadob: Changing the params from `entity: ApiKeyEntity` to just `self` would make the calling code a bit more clear.

src/data/src/migration/bootstrap/api_key.rs
```diff
@@ -98,43 +166,89 @@ async fn set_api_key_secret(name: &str, secret_plain: &str) -> Result<(), ErrorR
         DB::pg_execute(sql, &[&secret_enc, &name]).await?;
     }
 
+    DB::hql()
+        .delete(Cache::App, format!("api_key_{name}"))
```
sebadob: Missed that in earlier reviews, but you should not manually build the cache index here. This makes it very easy to screw up in the future when anything about it is changed. When you need the cache index of any value outside of its own CRUD functions, please make use of the `cache_idx()` I added to lots of structs already. In this case, you only need to make it `pub`. This ensures that future updates to the index will not introduce issues here, since it's very easy to overlook the update here.

sebadob (changes requested):

caniko:
**Addressed latest review points.**

- Changed `ApiKeyEntity::insert(entity)` to `insert(self)` — calling code is now `entity.insert().await` instead of `ApiKeyEntity::insert(entity).await`
- Made `cache_idx()` `pub` and used it in `set_api_key_secret()` instead of manually building the cache index string
- Rebased on latest `main`

sebadob (approved):
Thanks

Issue #1618, opened 2026-07-01
Configurable WebAuthn attestation trust + per-client authenticator allow-list


bartekus:
# RFC: configurable WebAuthn attestation trust + per-client authenticator allow-list

## Summary

Rauthy requests `AttestationConveyancePreference::None` on every passkey
registration today, so it never receives or validates authenticator attestation
and never learns the AAGUID (authenticator model id) of a registered credential.
That means an operator cannot express policies such as "only company-approved
security keys may be enrolled" or "only a login from an approved hardware
authenticator gets elevated assurance."

This RFC proposes an opt-in path with four small, separable pieces:

1. let an operator configure a set of trusted attestation CA certificates
   directly (no external metadata service required);
2. capture and persist the authenticator AAGUID on attested registrations;
3. let each OIDC client carry an allow-list of acceptable AAGUIDs;
4. reflect "authenticated by an allow-listed hardware authenticator" in the
   issued token's `acr` / `amr` (and optionally its audience).

Everything is opt-in. With no CA list configured, behavior is byte-for-byte what
it is today.

Verified against HEAD `f6223305`, 2026-07-01.

## 1. Problem statement

`start_passkey_registration` hard-codes `AttestationConveyancePreference::None`
(`webauthn-rs-0.5.5/src/lib.rs:564`) and discards attestation entirely. Because
the AAGUID is only conveyed under an attested ceremony, every credential Rauthy
stores today has no model identity attached (`PasskeyEntity`,
`src/data/src/entity/webauthn.rs:5`, stores `passkey`, `credential_id`,
`user_verified`, and no AAGUID / attestation column).

Two operator needs follow from that gap:

- **Enrollment policy.** An enterprise that issues a specific line of hardware
  authenticators wants "only those models may be registered as an MFA factor."
  There is no way to express or enforce that today.
- **Assurance signaling.** Even where any authenticator may enroll, an operator
  wants downstream apps to distinguish "logged in with an approved hardware key"
  from "logged in with an arbitrary passkey," via a stronger `acr` / `amr` on the
  token. Today `amr` is only ever `mfa` vs `pwd`
  (`src/service/src/token_set.rs:305-309`).

Depending on the FIDO Alliance Metadata Service (MDS) to obtain trust anchors is
a non-starter for many self-hosted and airgapped Rauthy deployments. Those
operators would rather supply their own trust anchors directly (their enterprise
attestation CA, or a vendor root they already hold) in config, and not reach out
to an external service at all.

[...]

PR #1619, opened 2026-07-01
feat(clients): per-client allowed_aaguids allow-list

bartekus:
## What

Adds an optional per-client allow-list of authenticator AAGUIDs
(`Client.allowed_aaguids`, stored as a CSV of canonical dashed UUIDs), surfaced through
the client API and the Admin UI. An empty or missing list applies no restriction, so
existing clients are unaffected.

This is the storage and management layer only. It mirrors, field for field, the
per-client CSV pattern introduced in #1608 (`allowed_resources` / `default_aud`): one
nullable column, an api_types field on the request and response, a service-layer
`Vec<String>` to CSV mapping, paired hiqlite + postgres migrations, and an Admin UI list
editor with i18n.

## Why

Operators who run their own attestation policy want to say "this client only accepts
specific authenticator models." This PR lands the generic, self-contained piece of that:
a place to store and manage the allow-list per client. It is useful on its own (it is the
data model an operator configures), and it is the first of the three separable pieces in
the attestation design proposed in #1618. It intentionally carries no WebAuthn-ceremony
behavior yet: capturing an AAGUID at registration and enforcing this list at attestation
time are the planned follow-ups (P2/P3 in #1618), so nothing here changes token issuance
or the login flow.

## Changes

Backend:
- `Client.allowed_aaguids: Option<String>` (CSV) with `allowed_aaguids_iter()` /
  `get_allowed_aaguids()` helpers, wired through every SQL site (save/create/update),
  the `Debug` impl, `Default`, the DB-to-DB migration insert, the anti-lockout literal,
  and the `ClientResponse` mapping.
- `service/client.rs`: request `Vec<String>` to CSV mapping (same shape as
  `allowed_resources`).
- api_types: `allowed_aaguids` on `UpdateClientRequest` (validated) and `ClientResponse`.
- Validation: `validate_vec_aaguid` + `RE_AAGUID` (canonical dashed UUID), so a malformed
  value is rejected with 400 at the API boundary.
- Migrations: `hiqlite/32_client_allowed_aaguids.sql`,
  `postgres/V27__client_allowed_aaguids.sql` (each `ADD COLUMN allowed_aaguids`, nullable).

Frontend:
- `ClientConfig.svelte`: a "Hardware Attestation" AAGUID list editor
  (state init, `$effect` reset, submit payload), plus `PATTERN_AAGUID`.
- api types + i18n strings (`allowedAaguids`, `descAllowedAaguids`, `validation.aaguid`)
  across all admin locales.

Tests:
- `src/bin/tests/zzg_handler_allowed_aaguids.rs`: create a client, set a two-entry
  allow-list, assert it round-trips on the PUT response and a fresh GET, assert clearing
  it persists as absent, and assert a malformed AAGUID is rejected with 400. Runs on both
  hiqlite and postgres.
- Updated the existing client-request test fixtures for the new field.

Docs: `CHANGELOG.md` entry.

## Notes

- Opt-in and backward compatible: the new column is nullable; existing rows read as
  `NULL` (no restriction), so behavior is unchanged after upgrade.
- No new dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## Unreleased
+
+### Changes
+
+#### Per-Client Allowed AAGUIDs
+
+Clients can now define an optional allow-list of authenticator AAGUIDs
+(`allowed_aaguids`, canonical UUID form) via the Admin UI and the client API. This is the
+storage and management layer for restricting a client to specific hardware
+authenticators; an empty / missing list applies no restriction. Enforcement during the
+WebAuthn attestation ceremony is a planned follow-up.
+
+[#1619](https://github.com/sebadob/rauthy/pull/1619)
+
 ## v0.36.0
 
 ### Breaking
--- a/frontend/src/api/types/clients.ts
+++ b/frontend/src/api/types/clients.ts
@@ -74,6 +74,10 @@ export interface UpdateClientRequest {
     /// Audiences always added to this client's tokens, independent of any request.
     /// Validation: PATTERN_URI
     default_aud?: string[];
+    /// Allow-list of authenticator AAGUIDs (canonical dashed UUID form) accepted for
+    /// hardware-attested sessions with this client.
+    /// Validation: PATTERN_AAGUID
+    allowed_aaguids?: string[];
     scim?: ScimClientRequestResponse;
 }
 
@@ -107,6 +111,7 @@ export interface ClientResponse {
     claims_at_root: boolean;
     allowed_resources?: string[];
     default_aud?: string[];
+    allowed_aaguids?: string[];
     scim?: ScimClientRequestResponse;
 }
 
--- a/frontend/src/i18n/admin/de.ts
+++ b/frontend/src/i18n/admin/de.ts
@@ -40,6 +40,8 @@ export let I18nAdminDe: I18nAdmin = {
         size: 'Größe',
     },
     clients: {
+        allowedAaguids: 'Erlaubte AAGUIDs',
+        descAllowedAaguids: `Optionale Positivliste von Authenticator-AAGUIDs (kanonische UUID-Form), die für hardwareattestierte Sitzungen mit diesem Client akzeptiert werden. Eine leere Liste wendet keine AAGUID-Einschränkung an.`,
         allowedResources: 'Erlaubte Ressourcen',
         defaultAud: 'Standard-Audiences',
         descAllowedResources: `Optionale RFC 8707 Resource Indicators, die dieser Client anfordern darf. Eine leere Liste lehnt jeden 'resource'-Parameter mit 'invalid_target' ab.`,
@@ -557,6 +559,7 @@ export let I18nAdminDe: I18nAdmin = {
         sendResetEmail: 'Reset E-Mail Senden',
     },
     validation: {
+        aaguid: 'Ungültige AAGUID',
         css: 'Ungültiger CSS Wert',
         origin: 'Ungültige Origin',
         uri: 'Ungültige URI',
--- a/frontend/src/i18n/admin/en.ts
+++ b/frontend/src/i18n/admin/en.ts
@@ -37,6 +37,8 @@ export let I18nAdminEn: I18nAdmin = {
         size: 'Size',
     },
     clients: {
+        allowedAaguids: 'Allowed AAGUIDs',
+        descAllowedAaguids: `Optional allow-list of authenticator AAGUIDs (canonical UUID form) accepted for hardware-attested sessions with this client. An empty list applies no AAGUID restriction.`,
         allowedResources: 'Allowed Resources',
         defaultAud: 'Default Audiences',
         descAllowedResources: `Optional RFC 8707 resource indicators this client may request. An empty list rejects any 'resource' request parameter with 'invalid_target'.`,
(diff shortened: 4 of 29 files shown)


sebadob:
I am not sure if it makes sense to specify this on a per-client basis. 

If an admin wants to limit here, it's because only some are considered being "safe enough" to meet the requirements. But how can they be safe enough only for some clients? It would also require an admin to always copy&paste the data for each  new client, which is very cumbersome. 

I think having a global, static config option makes a lot more sense. You specify a list which Passkeys you consider being safe only once, and you don't need to care about it on a daily basis. It would also make this PR a lot simpler and smaller. 

---

Apart from that, at the same time it makes sense to add a (super tiny) TODO I still have on my list, which is a config option to only allow hardware-based Passkeys. But that will only make sense when you actually touch the Webauthn Ceremony.

bartekus:
Makes sense, a global list is the right call. Whether an authenticator model is "safe enough" is a property of the model, not the client, and a copy&paste on every new client is exactly the daily friction that shouldn't exist. I'll drop the per-client column, admin UI and i18n and move it to a single global config option; that shrinks the PR a lot.

One coupling worth naming before I rework, so we sequence it well: an AAGUID allow-list only has teeth once the registration ceremony captures the AAGUID, and Rauthy requests `None` conveyance today (`start_passkey_registration`), so it never receives attestation or learns the AAGUID. That is the same ceremony your "only allow hardware-based Passkeys" TODO needs, so the list and that toggle depend on the same one change.

So I'd rather do this in small increments than reshape #1619 into a config surface that still enforces nothing. How would you like to sequence it? My suggestion, in whatever order you prefer:

1. attested ceremony + AAGUID capture, opt-in (empty config = today's exact behavior);
2. the global config on top (trusted-roots list + the hardware-only toggle), which now actually enforces.

One thing to flag for that config surface: verifying attestation needs the attestation-CA certificate(s), not just a bare AAGUID (a raw AAGUID is self-asserted), so I'd shape it as CA PEMs with an optional AAGUID narrowing on top. The finer calls (enrollment-gate vs a stronger `acr` / `amr`, exact config keys) I'm happy to settle in the PR once you point me at the first increment.

sebadob:
This is not an easy topic.

Requiring the user to provide CAs is a pretty bad UX. There are databases online we can make use of, e.g. `https://mds.fidoalliance.org`. When we do this, it should be backwards compatible and have a good UX.

Providing strict AAGUIDs is very fine-grained, but it also requires knowledge and research from the admin. When we use such an existing DB, we could also do things like instead of accepting AAGUIDs, also accept a list like `usb`, `nfc`, or simply `hardware` or `hardware-strict`, which then filters this DB automatically. 

There are also crates like `webauthn_rs_device_catalog` which does some nice things. [Data::strict()](https://docs.rs/webauthn-rs-device-catalog/0.5.0-20230418/webauthn_rs_device_catalog/struct.Data.html#method.strict) is something that probably most people would want. The issue here is that this is pretty outdated, and did not even have a stable release, not sure why. I just remember that there were some issues with the attestation, maybe that's why they have not done this and dropped development. Or maybe this list did not need an update since then.

We then need an additional DB column `passkeys.aaguid` which is optional for backwards compat. The UI then should show an additional hint next to the UV one, or it should change the symbol from the current fingerprint to something that shows it's an even more-stricter key, because attested devices must have UV anyway. This makes it possible to update keys afterwards, and make sure that they re-register ones that do not have an AAGUID registered. The most important thing is that it must not break anything.

---

I think the best way we can do this is:

- Create a `passkeys_cas` table that holds the necessary information for validation
- Create a `passkeys_cas_meta` that contains metadata about the `passkeys_cas` table. The reason I would add metadata here is because the full DB is > 10MB, and in most situations you only want a small subset. It's that big in JSON format, so probably a lot smaller when deserialized and filtered for the actual values we need. Maybe we can simply add the whole thing to our own DB. 
  I would then have a cron job that periodically checks if this DB was updated, or if our config was changed and we therefore need to update it. Then download, read it, filter it, and update our own DB with the filtered data. This makes future lookups fast and we don't need to keep the whole thing in memory only for registrations. The metadata is necessary to keep track about the filter that was applied during that procedure, and things like `last_update` timestamp, so we know when we actually need to update, or when we can skip it.
- Add a nullable `aaguid` column to `passkeys` (backwards compat) with an FK to `passkeys_cas`.

Regarding the behavior: 

If nothing is configured to narrow down the allowed authenticators (AAGUIDs, `usb`, `nfc`, or maybe `hardware-bound`, `strict`, ...), keep the current behavior. You only want attestation if you actually control the users on your instance, like for a company. For public instances with open registration, you never want this. Most people use synced software passkeys and you will drown in support calls if you limit these.

We would want config var that narrows the authenticator to something like `hardware` or `strict` (behavior from `webauthn_rs_device_catalog::Data`). In addition, a 2nd one where you can specify explicit AAGUIDs, which then filters the downloaded DB even further.

We probably want to save this config in the above-mentioned `passkeys_cas_meta` as well so we know when we may need to update our DB.

bartekus:
Agreed, MDS is the better call, and I'd rather not ship the manual-CA version at all either given that. Closing this one: it's per-client `allowed_aaguids` plus its UI and migrations, and none of that survives the redirect to a global, MDS-backed filter. The follow-up belongs on #1618 as a fresh PR against the design you sketched, so I'll consolidate it there.

PR #1620, opened 2026-07-02
feat(OTP): implementing OTP

11037luna:
As discussed in #1547 and #1344, these are the initial building blocks for implementing OTP in Rauthy.

If the user has a registered passkey, this takes precedence over the OTP in every situation.
This behaviour seems more logical since the Passkey is much more secure than OTP. However, we could discuss other options. 
OTP is disabled by default.

I intend to enhance other aspects of OTP, but I thought it would make more sense to propose an initial PR first.


frontend/src/lib/account/AccMFA.svelte
```diff
@@ -7,24 +7,29 @@
...
+    import type { WebauthnAdditionalData, WebauthnServiceReq } from '$mfa/webauthn/types.ts';
     import UserPasskey from '$lib5/UserPasskey.svelte';
     import type { MfaModTokenResponse, UserMfaTokenRequest } from '$api/types/mfa_mod_token';
     import Modal from '$lib/Modal.svelte';
     import InputPassword from '$lib/form/InputPassword.svelte';
     import Form from '$lib/form/Form.svelte';
     import IconArrowPathSquare from '$icons/IconArrowPathSquare.svelte';
+    import Template from '$lib5/Template.svelte';
+    import { TPL_OTP_LENGTH, TPL_IS_OTP_ENABLED } from '$utils/constants';
+    import type { _ } from '$env/static/private';
```
sebadob: That import is not used and should be removed.

frontend/src/lib/account/AccMFA.svelte
```diff
@@ -166,6 +187,103 @@
...
+        if (isInputError || !userId || !otpKind) {
+            return;
+        }
+
+        let tokenId = mfaModToken?.id;
+        if (!tokenId) {
+            showModal = true;
+            return;
+        }
+        let res = await otpRequest(userId, otpKind, otpName, tokenId);
```
sebadob: I would rename that function into `postOtp` or `postOtpRequest` to make it more clear what it does. The current name could also be a GET or anything else by just looking at the name. When outsourcing API requests, please always have the method in front to make it more clear. The API handlers in the backend use the same naming structure.

frontend/src/lib/account/AccMFA.svelte
```diff
@@ -357,7 +458,81 @@
...
+                    level={3}
+                    onclick={() => {
+                        showOtpInput = false;
+                    }}>{t.common.cancel}</Button
+                >
+            </Form>
+        {:else}
+            <div class="button">
+                <Button level={hasOtp === false ? 1 : 2} onclick={handleCreateOtp}
+                    >{t.mfa.registerNew}</Button
```
sebadob: We need a new `registerNew` i18n entry here. For instance, in german is makes no sense to have the same wording, it does not work. The current one only makes sense for Passkeys. OTPs definitely need their own one.

Apart from that, since right now only email is supported, I would hide the register new button if email OTP is already registered for this user. You cannot register multiple ones of that type and it just throws errors, which is a bad UX.

frontend/src/lib/OtpRequest.svelte
```diff
@@ -0,0 +1,201 @@
...
+                                        bind:ref={refInput}
+                                        name="otp"
+                                        autocomplete="one-time-code"
+                                        label={t.mfa.otp.code}
+                                        placeholder={'0'.repeat(otpSize)}
+                                        maxLength={otpSize}
+                                        minLength={otpSize}
+                                        pattern={PATTERN_OTP_CODE}
+                                        bind:isError={isInputError}
+                                    />
```
sebadob: It may make sense to create a custom `<input>` here.

I would make the input values way bigger for an OTP, probably in the range of like `2rem` or maybe even more, and `bold` maybe. Then, I would add CSS to the input to show spacing every 3 characters to make it better readable. 

Apart from that, having a custom input elem here makes it possible to make this a lot more pretty in the future and get rid of the default input look to better match existing OTP inputs out there.

frontend/src/lib/OtpRequest.svelte
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+    });
+
+    $effect(() => {
+        refInput?.focus();
+    });
+
+    async function onLoginOtpSubmit(_form: HTMLFormElement, params: URLSearchParams) {
+        let otpCode = params.get('otp');
```
sebadob: Make sure to remove all whitespace before going on from here. That makes it possible to relax the pattern (in the frontend only) and provide a better UX. Some people will use spaces during input for sure at some point, especially with 8 digit codes, e.g. when reading the code from a smartphone display.

frontend/src/lib/OtpRequest.svelte
```diff
@@ -0,0 +1,201 @@
...
+        color: hsl(var(--action));
+    }
+
+    /*.muted {*/
+    /*    color: hsla(var(--text) / .8)*/
+    /*}*/
+
+    /*progress {*/
+    /*    accent-color: hsl(var(--accent));*/
+    /*}*/
```
sebadob: These out-commented styles can be cleaned up.

frontend/src/utils/patterns.ts
```diff
@@ -26,6 +26,7 @@ export const PATTERN_LINUX_HOSTNAME = '^[a-zA-Z0-9][a-zA-Z0-9\\-.]*[a-zA-Z0-9]$'
 export const PATTERN_LINUX_USERNAME = '^[a-z][a-z0-9_\\-]{1,61}$';
 export const PATTERN_LOWERCASE = '^[a-z0-9\\-_\\/]{2,128}$';
 export const PATTERN_ORIGIN = '^(http|https)://[a-z0-9.:\\-]+$';
+export const PATTERN_OTP_CODE = '^[0-9]*$';
```
sebadob: Mentioned this above already. I would relax that pattern in the UI to allow spaces, and then clean them up before the submit. This improves the UX a bit.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+    if !&RauthyConfig::get().vars.otp.enable {
+        warn!("Request to get OTPs but `otp.enable` is set to `false`");
+        return Err(ErrorResponse::new(
+            ErrorResponseType::Forbidden,
+            "feature disabled",
+        ));
+    }
+
+    let user_id = id.into_inner();
+    principal.validate_user_or_admin(&user_id)?;
```
sebadob: An API Key with `users:read` should also be allowed to retrieve information here.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+    let user_id: String = id.into_inner();
+    let otp_kind = OtpKind::from(payload.otp_kind);
+    principal.validate_user_session(&user_id)?;
+
+    let token = MfaModToken::find(&payload.mfa_mod_token_id).await?;
+    let ip = real_ip_from_req(&req)?;
+    token.validate(principal.user_id()?, ip)?;
+
+    // this prevent to have multiple otp of the same kind
+    // should we allow multiple time-based otp?
```
sebadob: Right now, only email-based OTPs are allowed and it makes sense to allow only a single one. 

Later on, it could be useful to allow multiple TOTP, and give them a name during registration, just like for Passkeys. The user then can decide which one to use, if muiltiple are registered. Having 2 registered makes sure a user cannot be locked out, if one authenticator is lost.

So, I would keep this check in place right now. The only thing I would change is the return code. A `406 Not Acceptable` would be better. The UI could catch that easily and show a translated error message instead of the default one from the API.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+        warn!("Request to create an OTP but `otp.enable` is set to `false`");
+        return Err(ErrorResponse::new(
+            ErrorResponseType::Forbidden,
+            "feature disabled",
+        ));
+    }
+    payload.validate()?;
+
+    let user_id: String = id.into_inner();
+    let otp_kind = OtpKind::from(payload.otp_kind);
```
sebadob: You can request an `enum` on the payload instead of a `String`. This is preferrable, as it is reflected on the Swagger UI.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+
+    if let Ok(mut otp) = OneTimePassword::find_by_id_for_user(&payload.otp_id, &user_id).await {
+        otp.validate(&user_id, &payload.otp_code).await?;
+        otp.activate().await?;
+        return Ok(HttpResponse::Ok().finish());
+    }
+    Err(ErrorResponse::new(
+        ErrorResponseType::NotFound,
+        "otp does not exist",
+    ))
```
sebadob: You don't need to catch that error manually. A `?` behind `find_by_id_for_user()` will return a `404` automatically.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+    if !&RauthyConfig::get().vars.otp.enable {
+        warn!("Request to delete an OTP but `otp.enable` is set to `false`");
+        return Err(ErrorResponse::new(
+            ErrorResponseType::Forbidden,
+            "Feature disabled",
+        ));
+    }
+    payload.validate()?;
+
+    let is_admin = match principal.validate_admin_session() {
```
sebadob: An API Key with `users:delete` should be allowed as well.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+
+    if let Ok(otp) = OneTimePassword::find_by_id_for_user(&payload.otp_id, &user_id).await {
+        OneTimePassword::delete(&otp.id).await?;
+        // make sure to delete any existing MFA cookie when a key is deleted
+        let cookie = ApiCookie::build(COOKIE_MFA, "", 0);
+        let mut resp = HttpResponse::Ok().finish();
+        if let Err(err) = resp.add_cookie(&cookie) {
+            error!("Error deleting MFA cookie in delete_user_otp: {}", err);
+        }
+        return Ok(resp);
```
sebadob: You can build that response in a single step:

```rust
HttpResponse::Ok()
    .cookie(cookie)
    .finish()
```

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+            user_id, payload.otp_id
+        );
+    } else if principal.is_user(&user_id).is_err() {
+        warn!(
+            "Otp delete request from admin for user {} for otp {}",
+            user_id, payload.otp_id
+        );
+    }
+
+    if let Ok(otp) = OneTimePassword::find_by_id_for_user(&payload.otp_id, &user_id).await {
```
sebadob: The same thing as above. Make use of the Resuilt directly. There is no need to catch the error manually for the `404`.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+    request_body = OtpAuthFinishRequest,
+    responses(
+        (status = 200, description = "Ok", body = OtpAdditionalData),
+        (status = 401, description = "Unauthorized", body = ErrorResponse),
+        (status = 403, description = "Forbidden", body = ErrorResponse),
+    ),
+)]
+#[post("/users/{id}/otp/auth/finish")]
+pub async fn post_otp_auth_finish(
+    _id: web::Path<String>,
```
sebadob: It makes no sense to require the `id` in the path when it's not used at all. For webauthn, there are also paths like `/users/webauthn_start` and `/users/webauthn_finish` for an unauthenticated context that were added recently. The old paths only still exist for backwards compat. We want to get rid of the user ID in the path for these to never leak it into an unauthenticated context.

The only thing you need to watch out for is the ordering when the server routes are built so make sure to not overwrite the routes with the ones that catch "any id".

Edit:

I saw you have these additional routes below already. Then it's best to get rid of these with the `id` inside the path.

src/api/src/users.rs
```diff
@@ -1297,6 +1312,441 @@ async fn user_revoke_handle(
...
+    request_body = OtpAuthStartRequest,
+    responses(
+        (status = 200, description = "Ok", body = OtpAuthStartResponse),
+        (status = 401, description = "Unauthorized", body = ErrorResponse),
+        (status = 403, description = "Forbidden", body = ErrorResponse),
+    )
+)]
+#[post("/users/{id}/otp/auth/start")]
+pub async fn post_otp_auth_start(
+    id: web::Path<String>,
```
sebadob: Instead of grabbing the user ID from the path, grab it from the session. Then we can get rid of the id in the path all together to prevent possibly leaking it to an unauthenticated user during login. The same rework was done for webauthn recently as well.

src/api_types/src/users.rs
```diff
@@ -392,6 +392,80 @@ pub struct MfaModTokenResponse {
...
+pub struct OtpDeleteRequest {
+    pub otp_id: String,
+    #[validate(length(min = 32, max = 32))]
+    pub mfa_mod_token_id: Option<String>,
+}
+
+#[derive(Serialize, ToSchema)]
+pub struct ActiveOtp {
+    pub otp_id: String,
+    pub otp_kind: String,
```
sebadob: Use an `enum` for all possible `kind`s in all situations.

src/data/src/email/otp.rs
```diff
@@ -0,0 +1,97 @@
...
+    };
+
+    let req = EMail {
+        typ: EmailType::OtpRequest,
+        recipient_name: user.email_recipient_name(),
+        address: user.email.to_string(),
+        subject: format!("{email_sub_prefix} - {subject}"),
+        text: Some(text.render().expect("Template rendering: EmailOtpTxt")),
+        html: Some(html.render().expect("Template rendering: EmailOtpHtml")),
+    };
```
sebadob, line 69: These E-Mails do not provide good information about the expiry:

```
Code expires: 5
```

It's unclear what the `5` even means. I would definitely send an absolute datetime here, or even easier, remove it all together (also the "This code is only valid for a short period of time for security reasons." line). The UI could show a counter and auto-close the modal when the code expires, like it is the case fo MFA codes or Webauthn requests already. That would be a much improved UX, remove the translations issues here and make emails even simpler.

I would also change `You will find your OTP code below.` into something like `Your OTP is the following:`. "below" makes more sense to me if there is a bit more text in between.

The Text-only version needs additional line breaks in between to separate the OTP a bit more.

For the HTML version, If would also make the OTP way bigger, probably like 2 or 3 rem (match the UI then), ane apply the same spacing-rules via CSS as I already mentioned in the frontend part. People will type this in manually from smartphone displays at some point.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+impl From<&str> for OtpKind {
+    fn from(value: &str) -> Self {
+        match value {
+            "email" => Self::Email,
+            "phone" => Self::Phone,
+            "time" => Self::Time,
+            _ => Self::default(),
+        }
+    }
+}
```
sebadob: Instead of `From<String>` and `From<&str>`, a single `FromStr` works fine as well. Then use `#[column(parse)]` below.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+
+#[derive(Clone, Serialize, Deserialize, FromPgRow)]
```
sebadob: `Serialize` is not needed.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+        user_id: String,
+        name: Option<String>,
+        kind: OtpKind,
+    ) -> Result<Self, ErrorResponse> {
+        // if the len is longer than the algorithm it will be compressed by the digest
+        // if the len is shorter than the algorithm it will be padded with 0x30
+        let secret: [u8; digest::SHA512_OUTPUT_LEN] =
+            rand::generate(&rand::SystemRandom::new())?.expose();
+        let otp = Self {
+            id: get_rand(64),
```
sebadob: Why do you make the `id` that long here? Any special reason because it's sensitive later on? If not, please use `new_store_id()` as 24 chars are plenty here.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+impl OneTimePassword {
+    pub async fn create(
+        user_id: String,
+        name: Option<String>,
+        kind: OtpKind,
+    ) -> Result<Self, ErrorResponse> {
+        // if the len is longer than the algorithm it will be compressed by the digest
+        // if the len is shorter than the algorithm it will be padded with 0x30
+        let secret: [u8; digest::SHA512_OUTPUT_LEN] =
+            rand::generate(&rand::SystemRandom::new())?.expose();
```
sebadob: No need to generate a secret for email OTPs. Do that conditionally depending on the `kind`, and just use `Vec::default()` when not needed. Everything else wastes resources.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+        let otp = Self {
+            id: get_rand(64),
+            user_id,
+            name,
+            secret: secret.to_vec(),
+            last_used: 0,
+            kind,
+            is_active: false,
```
sebadob, line 102: Why do we need an `is_active` field and also default to `false`? Is this because of some TOTP flow later on?

migrations/hiqlite/32_one_time_password.sql
```diff
@@ -0,0 +1,17 @@
...
+            ON UPDATE CASCADE ON DELETE CASCADE,
+    name        TEXT    NULL,
+    secret      BLOB    NOT NULL,
+    last_used   INTEGER NOT NULL,
+    kind        TEXT    NOT NULL,
+    is_active   INTEGER DEFAULT false NOT NULL
+) STRICT;
+
+CREATE INDEX one_time_password_kind_index
+    ON one_time_password (kind);
```
sebadob: That's a useless index right now. After taking a look at the queries you're making, this index only makes sense to be a compound index of `(user_id, kind)`.

The current index is actually never used. The same is true for the Postgres migration.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+    pub async fn activate(&mut self) -> Result<(), ErrorResponse> {
+        self.is_active = true;
+        self.save().await
+    }
+
+    pub async fn validate(&self, user_id: &str, code: &str) -> Result<(), ErrorResponse> {
+        if self.user_id != user_id {
+            return Err(ErrorResponse::new(
+                ErrorResponseType::NotFound,
+                "otp does not exist",
```
sebadob: That error does not match the check that is being done.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+            ));
+        }
+        match self.kind {
+            OtpKind::Email => {
+                let current_time = OffsetDateTime::now_utc().unix_timestamp();
```
sebadob: You cannot know this, but I started migrating away from  `time`. Please make use of `chrono` instead.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+                let timeout = (current_time - self.last_used) / 60;
```
sebadob: When using `chrono` you can make a type-safe calculation here with `chrono::Duration`.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+            }
+        };
+
+        Ok(())
+    }
+
+    pub async fn request_otp(&mut self) -> Result<(), ErrorResponse> {
+        let user = User::find(self.user_id.clone()).await?;
+
+        let current_time = OffsetDateTime::now_utc().unix_timestamp();
```
sebadob: The same as above, please use `chrono` instead.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,841 @@
...
+    }
+}
+
+impl From<OneTimePassword> for OtpGetResponse {
+    fn from(value: OneTimePassword) -> Self {
+        Self {
+            id: value.id,
+            name: value.name,
+            last_used: value.last_used,
+            kind: value.kind.to_string(),
```
sebadob: This is another reason for using an enum. We can get rrid of the heap allocation here by doing that.

src/data/src/entity/users.rs
```diff
@@ -1793,6 +1794,30 @@ impl User {
...
+    }
+
+    pub async fn has_otp_of_kind_enabled(&self, kind: &OtpKind) -> Result<bool, ErrorResponse> {
+        Ok(OneTimePassword::find_kind_for_user(kind, &self.id)
+            .await
+            .is_ok_and(|o| o.is_active))
+    }
+
+    #[inline(always)]
+    pub async fn has_otp_enabled(&self) -> Result<bool, ErrorResponse> {
```
sebadob: That function can be optimized and made cheaper. 

Instead of retrieving all OTPs for a user and then manually filtering them, create a new query on `OneTimePassword` directly like

```sql
SELECT 1 FROM one_time_password WHERE user_id = $1 AND active = true
```

And then return `true` if the result is not empty.

src/data/src/rauthy_config.rs
```diff
@@ -573,6 +574,16 @@ impl Default for Vars {
             mfa: VarsMfa {
                 admin_force_mfa: true,
             },
+            otp: VarsOtp {
+                enable: false,
+                length: 6,
+                exp: 5,
```
sebadob: I would either `exp` that into `exp_mins` or even better save a type-safe `Duration` during config parsing directly to make it impossible to misuse. That's also still a TODO for some other config values.

src/data/src/rauthy_config.rs
```diff
@@ -1021,6 +1032,125 @@ Your account has not been compromised and no data was leaked."#.into()),
...
+                        subject: "One Time Password".into(),
+                        header: "One Time Password for".into(),
+                        text: Some("You will find your OTP code below.".into()),
+                        click_link: None,
+                        validity:
+                        Some("This code is only valid for a short period of time for security reasons."
+                                .into()),
+                        expires: Some("Code expires: ".into()),
+                        button: None,
+                        footer: Some("If this code has expired, you can request a new one.".into()),
```
sebadob: That `footer` translation makes no sense if it's actually not even possible to do this in the same way as for password reset links. This is only possible when triggering a new login flow. Just get rid of the not even used vars here.

We really only need:

- subject
- header
- text

The validity part, as mentioned above already, can be dropped as well when the UI shows an expiry counter and auto-closes the modal.

src/middlewares/src/csrf_protection.rs
```diff
@@ -152,6 +152,8 @@ fn is_path_csrf_exception(path: &str) -> bool {
                 || path.starts_with("/.well-known/")
                 // Webauthn has additional validation via Origin internally
                 || path.contains("/webauthn/auth/")
+                // Otp has additional validation via Origin internally
```
sebadob: > Otp has additional validation via Origin internally

That is definitely not true. OTP does not validate the `Origin` automatically. If this exception is in place, it possibly opens a CSRF vulnerability.

If this exception is necessary, at least do manual CSRF checks on the endpoints.

src/service/src/oidc/authorize.rs
```diff
@@ -55,24 +58,25 @@ pub async fn post_authorize(
...
-        } else {
-            None
-        };
+    // If a possibly existing mfa cookie does not match the given email, or the user
+    // has webauthn or otp disabled in the meantime, ignore it
+    let has_mfa_cookie = if user.has_webauthn_enabled()
+        && let Ok(c) = WebauthnCookie::parse_validate(&ApiCookie::from_req(req, COOKIE_MFA))
+    {
+        c.email == user.email
+    } else if user.has_otp_enabled().await.unwrap_or_default()
```
sebadob: The `.unwrap_or_default()` is used in multiple locations already. It would be way nicer to not even return a `Result<_>` from `has_otp_enabled()`. Instead, catch errors inside if and always return a `bool`, which makes it easier to use.

src/service/src/oidc/authorize.rs
```diff
@@ -55,24 +58,25 @@ pub async fn post_authorize(
...
+        && let Ok(c) = OtpCookie::parse_validate(&ApiCookie::from_req(req, COOKIE_MFA))
```
sebadob: I have an issue with the conflicting cookie names and types.

They use the exact same fields. This means you can deserialize a `WebauthnCookie` as an `OtpVookie` and vice versa without knowing what it actually is. The whole `OtpCookie` is just acopy&paste of `WebauthnCookie`, which does not make much sense.

So we have 2 options at this point:

1. Keep the same cookie for everything and have 2 different `new()` functions, one for Webauthn and another one for OTP. This is easier, but less flexible.
2. Actually have 2 separate cookies to have more flexibility, but even then do not copy&paste the functions that do the exact same things on the same expected output data, but somewhat unify it into `MfaCookie` and have separate functions.

Even though they have different levels of security, if an admin allows OTP, it's considered "good enough", so a single cookie makes sense from that perspective, and I would also not add more data to it to know the original MFA type. So I would go with Option 1, which is the cleaner approach. Rename it to `MfaCookie` and reuse builder and parser. Especially because the `c.email == user.email` is the same check for both versions as well. So I would even push the `user.email` into `parse_validate()` and have a clean, single function call.

src/service/src/oidc/authorize.rs
```diff
@@ -170,8 +177,12 @@ pub async fn post_authorize_refresh(
     user.check_enabled()?;
     user.check_expired()?;
 
+    // Webauthn overrides otp
     let require_webauthn =
         user.has_webauthn_enabled() && RauthyConfig::get().vars.lifetimes.session_renew_mfa;
+    let require_otp = !require_webauthn
+        && user.has_otp_enabled().await?
+        && RauthyConfig::get().vars.lifetimes.session_renew_mfa;
```
sebadob: Do the `RauthyConfig::get().vars.lifetimes.session_renew_mfa` check before `has_otp_enabled()`, as it's a lot cheaper. Always do cheaper checks first.

src/service/src/oidc/authorize.rs
```diff
@@ -299,6 +316,34 @@ pub(crate) async fn finish_authorize(
...
 
         Ok(AuthStep::AwaitWebauthn(step))
+    } else if data.require_otp {
+        // OneTimePassword enabled account
+        let step = AuthStepAwaitOtp {
+            code: get_rand(48),
+            header_csrf: Session::get_csrf_header(&session.csrf_token),
+            header_origin: data.header_origin,
+            email: user.email.clone(),
+            active_otps: user.get_otp_kind().await?,
```
sebadob, line 324: If in the future multiple authenticators for TOTP are allowed, we probably need to return the names / ids here as well, so the user can select which one to use, but that's fine for now with email only.

templates/email/otp.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+
+{% block title %}E-Mail OTP{% endblock %}
+
+{% block content %}
+<h1>{{ header }} {{ email_sub_prefix }}</h1>
+<p>{{ text }}</p>
+<p><b>{{ code }}</b></p>
```
sebadob: Mentioned this above already, but I would make the code way bigger and better readable here. Also use a `monospace` font for it.

config.toml
```diff
@@ -2160,6 +2160,58 @@ msc3861_enable = false
...
+# default: 6
+# overwritten by: OTP_LENGTH
+length = 6
+
+# The lifetime in minutes for OTP requests. Within
+# this time, an OTP request must have been validated.
+#
+# default = 5
+# overwritten by: OTP_EXP
+exp = 5
```
sebadob: We should add `_mins` to the value name to make it self-describing. That's an open TODO for some others where I forgot it as well.

config.toml
```diff
@@ -2160,6 +2160,58 @@ msc3861_enable = false
...
+# default: 'false'
+# overwritten by: OTP_ENABLE
+enable = false
+
+# The length of the one-time password generated by the HMAC-based algorithm.
+#
+# NOTE: Cannot be above 8 digits, will default to 8 otherwise.
+# default: 6
+# overwritten by: OTP_LENGTH
+length = 6
```
sebadob, line 2230: It makes sense to have a min length here. I would consider anything below `6` too insecure. Not sure if there are authenticators out there that can output less, but adding a check during `parse_otp()` and mentioning it in the comment makes sense to me.

sebadob (changes requested):
The UI parts needs some UX improvements:

Image

It should not be possible to even try trigger the error shown. I mentioned this already in the code somewhere. The button should not even be shown if email OTP exists already, and any errors that might happen here should return a proper code from the backend that the UI can catch and translate nicely.

The errors for OTP and Webauthn should be separated as well, as it's shown in both locations now.

11037luna:
Thanks for the feedback, I'm going to work on this this week.

src/data/src/email/otp.rs
11037luna: Thanks for the feedback, I wasn't really creative on this part. I will improve this.
Instead of the modal auto-closing, I was thinking the UI could show a button to request a new code. I was planning to create a route for this anyway, in case the user didn't receive the code.

src/service/src/oidc/authorize.rs
11037luna: I wasn't sure if you were okay to merge the cookie for WebAuthn with OTP.
The current implementation was very naive, as it was just to make OTP work like WebAuthn.
I'll unify the `MfaCookie` like option 1. This wil make things easier if anothe MFA method is implemented in the future.
Can't move the `user.email` check in the `parse_validate()` though, in `get_authorize` Rauthy does not know the user email.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,820 @@
...
+                ],
+            )
+            .await?;
+        }
+
+        Ok(otp)
+    }
+
+    pub async fn delete_all_otp_for_user(user_id: &str) -> Result<(), ErrorResponse> {
+        let sql = "DELETE FROM one_time_password WHERE user_id = $1";
```
sebadob, line 144: The migrations do not have an index on `user_id`, which would make this query insanely expensive, since it needs a full table scan.

src/data/src/entity/one_time_password.rs
sebadob: This was not addressed.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,820 @@
...
+        self.save().await?;
+        match self.kind {
+            OtpKind::Email => {
+                let code_len = code_len as usize;
+                let code = format!("{} {}", &code[0..code_len / 2], &code[code_len / 2..]);
+                send_email_otp(&code, &user).await;
+            }
+            // Unreachable should never panic since these kind aren't implemented
+            OtpKind::Time | OtpKind::Phone => {
+                unreachable!()
```
sebadob: In this case, an `unimplemented!()` makese more sense. You can then even remove the comment.

src/data/src/entity/one_time_password.rs
```diff
@@ -0,0 +1,820 @@
...
+        );
+        match self.kind {
+            OtpKind::Email => {
+                let code_len = code_len as usize;
+                let code = format!("{} {}", &code[0..code_len / 2], &code[code_len / 2..]);
+                send_email_otp(&code, &user).await;
+            }
+            // Unreachable should never panic since these kind aren't implemented
+            OtpKind::Time | OtpKind::Phone => {
+                unreachable!()
```
sebadob, line 391: The same as above - `unimplemented!()` makes more sense in that case.

src/data/src/entity/users.rs
```diff
@@ -1803,6 +1804,30 @@ impl User {
         self.roles.as_str().split(',')
     }
 
+    pub async fn get_otp_kind(&self) -> Result<Vec<ActiveOtp>, ErrorResponse> {
+        Ok(OneTimePassword::find_for_user(&self.id)
+            .await?
+            .iter()
```
sebadob: When you `.into_iter()`, you don't have to clone values below.

sebadob (changes requested):
Code is almost fine now. 

However, some CSS issues:

1. The OTP section on the account dashboard clashes with the above Webauthn button. We need more space here:

Image

2. When I add an OTP and then delete it, we button to register a new one does not re-appear. Instead, I see the "Test" button with no existing OTP:

Image

3. The OTP input for authentication during login or test looks way better now, but still needs some smaller fixes. 
    - The buttons are in the wrong order for a proper UX. "Ok" should be on the left with level `1`. This has "send" as text, which is very confusing. If you want to have a "resend code" at this stage, it should be on the right, and have a level of `2` or `3`. There should preferrably only ever be a single button on a page with level `1` to guide the users eye.         
    - The Input -> Buttons -> Text below is also very cramped. Please add some spacing here.
    - The input width expands until the full available width. It should depend on the required length, or if you don't know the expected length, make it fit for 8 digits and then stop growing in width.
    - I am not sure right now, but if you know the configured length from the backend (could be added to the API response if not), you could add an auto-submit if the input values is detected to match the expected length. This would get rid of the need to still manually press the button, and improve UX quite a bit.
    - This input should explicitly disable auto-fill / -complete. In my case, the password manager pops up with auto-fill suggestions, which can never work for an OTP, and it only gets in the way of the buttons below.

Image

Edit:

Sorry the review took a while. It's a big PR and I was short on time.

11037luna:
I've applied your changes requests, I haven't add an auto-submit for OTP code, I couldn't find a decent way to implement this right now.
I've also removed the auto-fill/complete feature.
_If a user uses OTP with their password manager, it could auto-fill the code. Maybe Rauthy should allow password managers to auto-fill when TOTP is implemented._

Thank you for taking the time to review the code, I've learned a lot from your feedback c:

sebadob (approved):
> I haven't add an auto-submit for OTP code, I couldn't find a decent way to implement this right now.

This is pretty easily done when you check the bound value in an `$effect()`, but no worries, I can add that. I need to update german translations anyway.

Thank you.

PR #1621, opened 2026-07-03
Feat: RFC 8252 + relaxed client name regex

alari:
Four independent bugs that break standard OAuth login for MCP clients (Claude Code, opencode) when Rauthy runs on postgres. Ran into them wiring Rauthy up as the auth server for an MCP resource server; all reproduce on current main.

- client name regex rejects `()`, so a client named `Claude Code (nks)` 400s on registration
- loopback `redirect_uri` is not port-agnostic (RFC 8252 §7.3) — a registered `http://127.0.0.1/cb` never matches the requested `http://127.0.0.1:PORT/cb`, so CLI clients 400 with "Invalid redirect uri"
- `ClientDyn::update_used` binds one param to a two-param UPDATE, so the dynamic-client auth_code token exchange always 500s on postgres (hiqlite is fine)
- `danger_disable_introspect_auth` is checked after `find_enabled_client(azp)`, so introspecting a token issued to an ephemeral client 401s even with the flag on

One commit each. Happy to split into separate PRs if you prefer.

### DIFF
--- a/src/common/src/regex.rs
+++ b/src/common/src/regex.rs
@@ -27,7 +27,7 @@ pub static RE_CLIENT_ID: LazyLock<Regex> =
 pub static RE_CLIENT_ID_STRICT: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9._\-]{2,256}$").unwrap());
 pub static RE_CLIENT_NAME: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new(r"^[a-zA-Z0-9À-ɏ-\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{2,128}$").unwrap()
+    Regex::new(r"^[a-zA-Z0-9À-ɏ()-\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{2,128}$").unwrap()
 });
 pub static RE_CODE_CHALLENGE: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9-._~]{43,128}$").unwrap());
--- a/src/common/src/regex.rs
+++ b/src/common/src/regex.rs
@@ -27,7 +27,7 @@ pub static RE_CLIENT_ID: LazyLock<Regex> =
 pub static RE_CLIENT_ID_STRICT: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9._\-]{2,256}$").unwrap());
 pub static RE_CLIENT_NAME: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new(r"^[a-zA-Z0-9À-ɏ()-\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{2,128}$").unwrap()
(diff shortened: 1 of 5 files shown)


sebadob:
> client name regex rejects (), so a client named Claude Code (nks) 400s on registration

That's not a bug at all. It's intentional. That's a Claude Code issue. But yeah we can allow `()`.

> loopback redirect_uri is not port-agnostic (RFC 8252 §7.3) — a registered http://127.0.0.1/cb never matches the requested http://127.0.0.1:PORT/cb, so CLI clients 400 with "Invalid redirect uri"

It's not, on purpose. If you configure it properly with a `*` like `http://127.0.0.1:*`, it will work. This is absolutely on purpose and will not be changed. The current way is a lot more intentional and a lot safer. It may not 100% match the RFC, but for good reason, as long as you configure it properly.

src/common/src/regex.rs
```diff
@@ -27,7 +27,7 @@ pub static RE_CLIENT_ID: LazyLock<Regex> =
 pub static RE_CLIENT_ID_STRICT: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9._\-]{2,256}$").unwrap());
 pub static RE_CLIENT_NAME: LazyLock<Regex> = LazyLock::new(|| {
-    Regex::new(r"^[a-zA-Z0-9À-ɏ-\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{2,128}$").unwrap()
+    Regex::new(r"^[a-zA-Z0-9À-ɏ()._\-\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{2,128}$").unwrap()
```
sebadob: Allowing `._\` in a human-readable name does not make too much sense to me. There are usually placeholders when spaces are not allowed. Especially the `\` can become unnecessarily dangerous in some situations since it's used as escape char. I don't see any need to allow this here. `()` are find for me though.

src/data/src/entity/clients.rs
```diff
@@ -1293,6 +1293,7 @@ impl Client {
         let has_any = self.get_redirect_uris().iter().any(|uri| {
             (uri.ends_with('*') && redirect_uri.starts_with(uri.split_once('*').unwrap().0))
                 || uri.as_str().eq(redirect_uri)
+                || loopback_redirect_match(uri, redirect_uri)
```
sebadob: Yeah, as mentioned already, please just get rid of it and user the properly syntax to allow any port specifically with the wildcard instead of doing it by default for everyting. The RFC approach is actually pretty unsafe.

sebadob (changes requested):

alari:
Updated: narrowed the client-name change to just `()`, and dropped the loopback commit. Fair enough — I will use `:*` where I control the client.

The one case that keeps me on a fork: with CIMD the `redirect_uris` come from the client's metadata document (claude.ai's, in my case), which advertises `http://localhost/callback` with no port and which I can't edit to add `:*`. So there's no config on my side that makes the ephemeral client's loopback port match. Not asking you to change the default — just flagging it in case there's an angle I'm missing (e.g. an opt-in per-client flag). Either way, thanks a lot for the quick review.

sebadob:
> with CIMD the redirect_uris come from the client's metadata document (claude.ai's, in my case), which advertises http://localhost/callback with no port and which I can't edit to add :*. So there's no config on my side that makes the ephemeral client's loopback port match.

Got it. That's a bit tricky. We could solve this in 2 ways:

- Always allow redirects by RFC for ephemeral and dynamic clients when they redirect to `localhost`
- Or add a config option to opt-in to matching the RFC. In this case, we should accept both syntaxes for `localhost`. I would not allow this for static clients though, even with this option set, because an admin has full control over these anyway.

I think the config option is the most flexible. Maybe adding an option like `access.rfc_8252_enable`. This should be `false` by default to be as secure as possible out of the box. Then add a good description for this value and mention what it actually does, because most people will not know what the RFC actually says.

Apart from that, if you merge from `main`, I pushed the PG query params fix with `v0.36.0` as well to include it in the release.

alari:
Done — went with option 1: loopback any-port matching only for dynamic and ephemeral clients (static clients keep exact/`*` matching). Scoped it in `validate_redirect_uri` via `is_dynamic() || is_ephemeral()`.

Also rebased onto current main and dropped the `update_used` commit — I see you already fixed that param count there. So this is now 3 commits: client-name `()`, the introspect-flag ordering, and the scoped loopback. Thanks!

alari:
Sorry — I misread your preference and pushed the always-on variant first. Redone as the config option you suggested: `access.rfc_8252_enable`, `false` by default, applied to dynamic and ephemeral clients only (static always require an exact / `*` match, even when set). Added a description covering what the RFC actually does. Force-pushed. Thanks!

sebadob (approved):

PR #1082, opened 2025-07-02
Update changelog with vault example configs

cocoon:
a quick change, but I might need to test later, as the next 2 days I will be away.
But I will fix and test then.

currently I am even looking into an All-in-One Docker-Compose Example, but I need a delay for rauthy and of course a new nightly docker container that contains the vault Feature.

Update:
I think this should work as delay to wait for setup done and then start rauthy:
```
vault-setup-svc:
        condition: service_completed_successfully
```
Now I only need a new nightly.
And then test and hope, that I can inject the config with newlines like this ....

In this example the docker host has the IP address 10.0.0.100.
This needs to be changed to own IP.

todo: add shared File for one time token

```
version: "3"

networks:
  rauthy-test:
    driver: bridge

services:
  dev-vault-svc:
    image: hashicorp/vault:1.20.0
    container_name: dev-vault
    cap_add:
      - IPC_LOCK  # Lock memory to prevent sensitive data from swapping to disk
    environment:
      - VAULT_DEV_ROOT_TOKEN_ID=unsafe
      - VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:8200
      - VAULT_ADDR=http://10.0.0.100:8202  # Set the Vault address
    ports:
      - "8202:8200"  # Expose Vault port 8200 on host port 8202
    command: server -dev  # Start Vault in development mode
    healthcheck:
      test: ["CMD", "vault", "status"]
      interval: 30s
      timeout: 10s
      retries: 5

  vault-setup-svc:
    image: hashicorp/vault:1.20.0
    environment:
      - VAULT_ADDR=http://10.0.0.100:8202
    depends_on:
      - dev-vault-svc
    restart: "no"
[...]

PR #1634, opened 2026-07-08
[Book] Add example to load config from a vault

cocoon:
As it was missing, I now started to create a file (unlinked) with examples for loading the configuration from a vault as talked about here: https://github.com/sebadob/rauthy/pull/1082#issuecomment-3043817880

Might need to verify the rauthy version in the docker files if it still exists or if using a curent one works without changes.

And I had no chance to upgrade to the latest  version currently so I don't know how the latest changes regarding the "config parser" will affect it:

`Technically, this is not a breaking change, but it might be for you. The config parser now has an additional validation for left-over, unknown keys after parsing has finished. If any data is left, Rauthy will panic. This makes sure you cannot have typos or data in your config that does nothing at all. This means this might be a somewhat "breaking" change if you have invalid or additional data in your config. Just be prepared to adjust your config when you update.`

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -7,6 +7,7 @@
   - [Kubernetes](./getting_started/k8s.md)
   - [First Start](./getting_started/first_start.md)
   - [Shutdown](./getting_started/shutdown.md)
+  - [Vault](./getting_started/vault.md)
 
 - [Production Config](config/production_config.md)
   - [Rauthy CLI](config/cli.md)
--- a/book/src/getting_started/vault.md
+++ b/book/src/getting_started/vault.md
@@ -1,4 +1,4 @@
-# Rauthy + Vault config (short guide)
+# Rauthy + Vault config
 
 Support for loading the config from a Vault source was added in v0.31.0. To enable it, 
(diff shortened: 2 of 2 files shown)


sebadob:
The subject says "Draft" but it's marked as ready for review. What is it?

cocoon:
> The subject says "Draft" but it's marked as ready for review. What is it?

yes, it is still a draft, because one needs to verify If it still works with current version. I have a project including next week, then I might be able to verify it.

cocoon:
OK today I could test it and it is working with the example and the current rauthy docker v0.36.1 and added one more hint.

Is it ok like this?

Still not linked anyhwere, just a new file.

sebadob (changes requested):
Sorry took quite some time.

Looks good so far. You only have too long lines. Would be nice to add line breaks at 100 chars matching the other files.

And then of course add it to `SUMMARY.md`.

cocoon:
> Sorry took quite some time.

No problem, I have the same experience. Just when I think I’ll have some free time, something else comes up...
Will try to fix the long lines and add it to summary.

cocoon:
so this should look nicer and still works, only the long password line I think I cannot shorten more.
Added it to Summary as "Vault"

sebadob (approved):
Thanks

PR #1640, opened 2026-07-16
feat(users): accept preferred_username on user creation

### LINKED ISSUE #1615: `/users` does not accept `preferred_username` for user creation

Building on https://github.com/sebadob/rauthy/issues/1534.

AFAICT, this is not possible: the requester can specify `preferred_username` when creating a user via `/user` on the API with `curl`.

https://github.com/sebadob/rauthy/blob/f6223305c1b2183bb8fa9e24e3ed44045a6c26a2/src/api_types/src/users.rs#L54-L81

It's possible to specify `preferred_username` when creating a new user.

Creating it manually after I fire the `curl` request. It seems to be possible via an additional API request.

https://github.com/sebadob/rauthy/blob/f6223305c1b2183bb8fa9e24e3ed44045a6c26a2/src/api_types/src/users.rs#L129-L137

**Additional context**
N/A.

bartekus:
### What

`POST /users` (`NewUserRequest`) did not accept `preferred_username`, so operators could only set it with a second request after creation (as raised in #1615). This adds it to the creation payload.

### Changes

- `NewUserRequest` gains an optional `preferred_username`, validated with `RE_PREFERRED_USERNAME`, the same rule the self-registration request (`NewUserRegistrationRequest`) already uses.
- `User::create_from_new` persists it through `UserValues`, mirroring `create_from_reg`: a `validate_preferred_username_free` pre-check for a clean early error, with the existing `users_values.preferred_username` unique index as the authoritative guard. The tz-only persistence branch is generalized so a `preferred_username` (or a non-UTC tz) triggers the `UserValues` insert; UTC stays unpersisted as before.
- The update path is unchanged.

### Notes

`post_users` returns `user.into_response(None)`, so the create response body does not echo `user_values`; the value is persisted and visible on the subsequent `GET /users/{id}`, which the extended `test_users` integration test asserts.

### Tests

`cargo check --workspace --tests`, `fmt`, and `clippy` clean on the changed crates. `test_users` now creates a user with `preferred_username` and asserts it round-trips on read.

Closes #1615

### DIFF
--- a/src/data/src/entity/users.rs
+++ b/src/data/src/entity/users.rs
@@ -46,6 +46,7 @@ use serde::{Deserialize, Serialize};
 use std::cmp::max;
 use std::default::Default;
 use std::fmt::{Debug, Formatter};
+use std::mem;
 use std::ops::Add;
 use time::OffsetDateTime;
 use tracing::{debug, error, trace};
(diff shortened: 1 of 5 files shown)


src/data/src/entity/users.rs
```diff
@@ -226,23 +226,31 @@ impl User {
...
 
     pub async fn create_from_new(new_user_req: NewUserRequest) -> Result<User, ErrorResponse> {
+        // pre-uniqueness check for better UX and error handling; the DB unique index on
+        // `users_values.preferred_username` is the authoritative guard (see UserValues::insert).
+        if let Some(preferred_username) = &new_user_req.preferred_username {
+            UserValues::validate_preferred_username_free(preferred_username.clone()).await?;
+        }
+
         let tz = new_user_req.tz.clone();
+        let preferred_username = new_user_req.preferred_username.clone();
```
sebadob: No need to `clone()` here. You can use `mem::take()` instead.

sebadob (changes requested):

src/data/src/entity/users.rs
bartekus: Done in 9ae99803.

`from_new_user_req()` reads neither `tz` nor `preferred_username`, so both are now moved out with `mem::take()` rather than cloned; I took the `tz` line one above it as well since it was the identical pattern.

The remaining `clone()` in the uniqueness pre-check is load-bearing: `validate_preferred_username_free()` takes an owned `String` because the hiqlite `params!()` path needs owned parameters.

sebadob (approved):
Thanks.

PR #1649, opened 2026-07-17
fix(users): upgrade session to MFA after passkey registration

### LINKED ISSUE #1636: UX: session not upgraded after passkey registration — admin UI forces a logout/login round-trip
Title: UX: session not upgraded after passkey registration — admin UI forces a logout/login round-trip

## What happens (v0.36.0)

1. A fresh `rauthy_admin` user logs in with a password (no passkey yet).
2. They open `/auth/v1/admin` → correct gate message: *"A Rauthy admin account must have MFA enabled…"*
3. They go to `/auth/v1/account`, enroll a passkey — success, account shows *MFA activated ✓* (after a manual page reload, see note below).
4. `/auth/v1/admin` still shows the same gate message until they **log out and log back in**.

For a first-time admin bootstrap this is a confusing dead end: the user just proved possession
of the new authenticator (registration ceremony with user gesture), their account state says
MFA is active, but the UI keeps refusing with instructions that read like something went wrong.

## Why

`POST /users/{id}/webauthn/register/finish` (`src/api/src/users.rs`,
`post_webauthn_reg_finish`) calls `webauthn::reg_finish(...)` and returns `201` — the **current
session's `is_mfa` stays `false`**. The upgrade primitive already exists
(`Session::set_mfa`, `src/data/src/entity/sessions.rs`) and is used by the login paths
(`src/service/src/oidc/authorize.rs` line ~133, provider `login_finish.rs`), just not by the
registration-finish path. The admin UI checks the session flag, hence the forced re-login.

## Proposal

Either (a) after a successful non-magic-link `reg_finish` on the principal's **own** session,
upgrade it in place — `session.set_mfa(true)` — since the registration ceremony required both
an authenticated session and a fresh user gesture on the new authenticator; or, if you consider
registration weaker than an assertion, (b) have the account SPA immediately trigger a WebAuthn
**assertion** with the just-registered credential and upgrade the session on success. Both
remove the logout/login gymnastics; (b) keeps the "MFA session = a completed assertion"
invariant strict.

A smaller companion nit: after enrolling the passkey, the account page still shows *MFA
activated ✗* until a manual reload — the SPA state doesn't refresh on registration success.

Happy to test a patch on our deployment.

bartekus:
Resolves #1636.

Registering a Passkey from the Account dashboard did not touch the current session, so `is_mfa`
stayed `false` until the next login. A fresh admin account that just enrolled its first Passkey was
therefore still rejected by `admin_force_mfa`, with a message that reads like something went wrong,
and had to log out and back in.

As discussed, the session is now upgraded in place on a successful `reg_finish`.

### Backend

`post_webauthn_reg_finish` checks `is_mfa` on the principal's session after a successful
`webauthn::reg_finish` and upgrades it via the existing `Session::set_mfa`, the same primitive the
login paths already use. It only writes when the session is not already MFA, to avoid a pointless DB
write plus cache put on every additional Passkey.

Two guards fall out of the existing structure rather than being added:

- the `magic_link_id` branch returns before this code, so the reset / new-account flow is untouched
  (it has no session anyway);
- `principal.is_user(&id)?` already runs above, so an admin registering a credential for someone
  else can never upgrade their own session.

The registration start already requires an `MfaModToken`, so the user has proven recent credentials
inside a short window, and the ceremony itself is a fresh proof of possession for this very session.

### Frontend

The companion nit from the issue is fixed too: the Account dashboard showed *MFA activated ✗* until
a manual reload. The indicator reads `user.webauthn_user_id`, but `AccMFA` received `user` as a
read-only prop and only refreshed its own `passkeys` array. `user` is `$bindable()` now and is
refetched after registering or deleting a Passkey, so the indicator flips immediately.

Deleting the last Passkey had the mirror bug (indicator stuck on the check mark) and is fixed by the
same change. As a side effect, the re-auth modal no longer offers password re-auth off a stale
`user` right after the first Passkey was registered.

### Verification

- `cargo check` / `cargo clippy` / `cargo fmt --check` clean.
- `just test-hiqlite` full suite green.
- `npm run check` (svelte-check) reports no new problems, `prettier --check` clean.

### DIFF
--- a/src/api/src/users.rs
+++ b/src/api/src/users.rs
@@ -1707,6 +1707,16 @@ pub async fn post_webauthn_reg_finish(
         principal.is_user(&id)?;
 
         webauthn::reg_finish(id, payload).await?;
+
+        // The registration ceremony is a fresh proof of possession for this very session, and
+        // starting it already required an `MfaModToken`. Upgrade the session in place so the user
+        // does not need a logout / login round-trip to satisfy `admin_force_mfa`.
+        let session = principal.get_session()?;
+        if !session.is_mfa {
+            let mut session = session.clone();
+            session.set_mfa(true).await?;
+        }
+
         Ok(HttpResponse::Created().finish())
     }
 }
--- a/src/api/src/users.rs
+++ b/src/api/src/users.rs
@@ -1715,10 +1715,9 @@ pub async fn post_webauthn_reg_finish(
         // The session lookup is deliberately graceful: the passkey has already been persisted at
         // this point, so a flow that gets here without a session (during an initial account setup,
         // for instance) must never have its successful registration turned into an error.
-        if let Some(session) = &principal.session
+        if let Some(mut session) = principal.into_inner().session
             && !session.is_mfa
         {
-            let mut session = session.clone();
             session.set_mfa(true).await?;
         }
 
--- a/src/api/src/users.rs
+++ b/src/api/src/users.rs
@@ -1711,8 +1711,13 @@ pub async fn post_webauthn_reg_finish(
         // The registration ceremony is a fresh proof of possession for this very session, and
         // starting it already required an `MfaModToken`. Upgrade the session in place so the user
         // does not need a logout / login round-trip to satisfy `admin_force_mfa`.
-        let session = principal.get_session()?;
-        if !session.is_mfa {
+        //
+        // The session lookup is deliberately graceful: the passkey has already been persisted at
+        // this point, so a flow that gets here without a session (during an initial account setup,
+        // for instance) must never have its successful registration turned into an error.
+        if let Some(session) = &principal.session
+            && !session.is_mfa
+        {
             let mut session = session.clone();
             session.set_mfa(true).await?;
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## Unreleased
+
+### Changes
+
+#### Session MFA Upgrade after Passkey Registration
+
+Registering a Passkey from the Account dashboard upgrades the current session to MFA now. Before,
+`is_mfa` stayed `false` until the next login, which meant a fresh admin account that just enrolled
+its first Passkey was still rejected by `admin_force_mfa` and had to log out and back in. The
+Account dashboard also updates the MFA indicator directly after registering or deleting a Passkey
+instead of only after a manual page reload.
+
+[#1649](https://github.com/sebadob/rauthy/pull/1649)
+
 ## v0.36.0
 
 ### Breaking
--- a/frontend/src/lib/account/AccMFA.svelte
+++ b/frontend/src/lib/account/AccMFA.svelte
@@ -22,7 +22,7 @@
     import Form from '$lib/form/Form.svelte';
     import IconArrowPathSquare from '$icons/IconArrowPathSquare.svelte';
 
-    let { user }: { user: UserResponse } = $props();
+    let { user = $bindable() }: { user: UserResponse } = $props();
 
     const isSupported = 'credentials' in navigator;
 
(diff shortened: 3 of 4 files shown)


src/api/src/users.rs
```diff
@@ -1707,6 +1707,16 @@ pub async fn post_webauthn_reg_finish(
         principal.is_user(&id)?;
 
         webauthn::reg_finish(id, payload).await?;
+
+        // The registration ceremony is a fresh proof of possession for this very session, and
+        // starting it already required an `MfaModToken`. Upgrade the session in place so the user
+        // does not need a logout / login round-trip to satisfy `admin_force_mfa`.
+        let session = principal.get_session()?;
```
sebadob: This can error unexpectedly. For instance during the very first account setup, there is no session, but the `.get_session()?` and break this flow. Instead, do a graceful check if the session exists and update.

sebadob (changes requested):

CognitiveSand:
@bartekus Tested your branch today — **both fixes work as described**. Since our
production gate only moves on stable releases, I built `fix/session-mfa-after-passkey-reg`
from source (via the `rauthy-builder` container) and ran it in a local stack
(Caddy `tls internal` in front, `PROXY_MODE`, Hiqlite), driving the WebAuthn ceremony
headlessly with Chromium's CDP virtual authenticator. Same script against stock
v0.36.0 as the baseline. All checks in one browser session, zero reloads or re-logins
after the initial password login:

| check (same session, no reload) | v0.36.0 | your branch |
|---|---|---|
| `GET /auth/v1/auth_check_admin` before registration | 406 | 406 |
| … immediately after passkey `reg_finish` | **406 — still gated** | **200 — upgraded in place** |
| account "MFA activated" indicator after register | stale ✗ | updates live ✓ |
| indicator after passkey delete | stale | updates live ✓ |
| `/auth/v1/admin/users` visit afterwards | gate message | full admin UI |

So the original #1636 dead end (fresh admin: register passkey → admin UI still
demands logout/login) is gone, and the companion indicator nit is fixed for both
register and delete.

One observation, not an objection: after deleting the session's only passkey the
session keeps `is_mfa = true` (`auth_check_admin` stays 200). That matches the
existing semantics ("an MFA proof happened in this session") and the login paths
behave the same way, so it reads as consistent — just noting it since it's now
reachable through the new in-place upgrade too.

Thanks for the quick turnaround — from our side this resolves #1636.

src/api/src/users.rs
bartekus: Done in c14581ce. The session lookup is now a graceful `if let Some(session) = &principal.session` instead of `get_session()?`, so a missing session can no longer turn a successful registration into an error.

Worth noting for the record: the passkey is already persisted by the time we reach this point, so failing here would have been wrong regardless of whether the path is reachable. That is now stated in the comment.

src/api/src/users.rs
```diff
@@ -1707,6 +1707,21 @@ pub async fn post_webauthn_reg_finish(
...
+        // starting it already required an `MfaModToken`. Upgrade the session in place so the user
+        // does not need a logout / login round-trip to satisfy `admin_force_mfa`.
+        //
+        // The session lookup is deliberately graceful: the passkey has already been persisted at
+        // this point, so a flow that gets here without a session (during an initial account setup,
+        // for instance) must never have its successful registration turned into an error.
+        if let Some(session) = &principal.session
+            && !session.is_mfa
+        {
+            let mut session = session.clone();
```
sebadob: This is an unnecessary `clone`. You can avoid it entirely with 

```rust
if let Some(mut session) = principal.into_inner().session
```

sebadob (changes requested):

src/api/src/users.rs
bartekus: Done in 54fdaa8e. `ReqPrincipal` is a `web::ReqData<Principal>` and this is its last use in the handler, so `into_inner()` hands the `Option<Session>` over directly and the clone is gone:

```rust
if let Some(mut session) = principal.into_inner().session
    && !session.is_mfa
{
    session.set_mfa(true).await?;
}
```

Behaviour is unchanged, a missing session is still a no-op rather than an error.

Also rebased onto current `main` (was still on `73b00c4c`), so the branch is clean against the RFC 8252 merge. `cargo clippy --all-features -- -D warnings` is green.

sebadob (approved):
Thanks

Issue #1641, opened 2026-07-16
RFC 9068: emit `at+jwt` typ header on access tokens


bartekus:
### Summary

Rauthy's access tokens are already RFC 9068-shaped JWTs (`iss`, `aud`, `sub`, `exp`, `iat`, `jti`, `scope`, `cnf`), but the JWT header `typ` is hard-coded to `"JWT"` (`src/jwt/src/token.rs:29`) and validation rejects anything else (`token.rs:97-100`). RFC 9068 §2.1 specifies the access-token header `typ` as `at+jwt` (media type `application/at+jwt`), so a resource server using a standard JWT library can assert it was handed an access token and reject an ID token or a plain JWT carrying the same claims.

### Proposal

- Emit `typ: at+jwt` on **access tokens** only; ID and refresh tokens keep `typ: JWT`.
- Relax the token validator to accept both `JWT` and `at+jwt`, so already-issued tokens and ID tokens still validate and rauthy's own consumers (introspection, forward_auth) keep working.
- No new discovery field; `typ` is a token-header change.

Small surface: the `typ` constant and the validation check in `src/jwt/src/token.rs`, plus the branch that decides access-token vs ID-token type.

### The one call for you

Emitting `at+jwt` changes what a downstream resource server sees in the header. Standard RFC-9068-aware libraries want exactly this; a resource server that today hard-checks `typ == "JWT"` on Rauthy access tokens would need to accept `at+jwt` too. So: emit it **unconditionally** (cleanest, fully spec-compliant), or gate it behind a config flag for a transition window? I lean unconditional (it is the standard, and Rauthy validates both), but it is your call given the downstream-visibility change.

Happy to open the PR once you pick.

sebadob:
It makes sense to change the token type unconditionally. I would merge the PR when it's clear that we will publish another minor version next, and not just a patch, since it's a breaking change. 

The `rauthy-client` then also needs an update for the JWT validation. When a new version for the client is released beforehand, this change can be done without any interruption in service.

PR #1650, opened 2026-07-17
Client: accept `at+jwt` JWT header typ

bartekus:
Refs #1641. Prep for the server-side change; safe to merge and release on its own.

RFC 9068 specifies `at+jwt` as the JWT header `typ` for OAuth 2.0 access tokens. The client's
validation only accepts `JWT` today, so it would reject Rauthy's access tokens the moment Rauthy
emits the RFC-compliant value. Accepting both here first is what makes that switch possible without
an interruption in service, as you described in #1641.

- `JwtToken::validate_claims_into` accepts `JWT` and `at+jwt`.
- Version bumped to `v0.14.3` + CHANGELOG entry, example `Cargo.lock`s refreshed.

Nothing else changes: tokens issued before the switch, and ID tokens, keep using `JWT`.

The server-side counterpart is #1651, which is breaking and can wait for a minor. Releasing this
client version first is what avoids any interruption in service.

### DIFF
--- a/rauthy-client/CHANGELOG.md
+++ b/rauthy-client/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.14.3
+
+The JWT header `typ` validation accepts `at+jwt` in addition to `JWT` now. RFC 9068 specifies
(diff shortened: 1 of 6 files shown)


sebadob:
Thanks.

PR #1651, opened 2026-07-17
feat(oidc): RFC 9068 `at+jwt` typ header on access tokens

### LINKED ISSUE #1641: RFC 9068: emit `at+jwt` typ header on access tokens

bartekus:
Closes #1641.

Rauthy's access tokens are already RFC 9068-shaped, but the JWT header `typ` was hardcoded to the
generic `JWT` for every token, so a resource server could not tell an access token from an ID token
by the header alone. Access tokens use `at+jwt` now; ID, refresh and logout tokens keep `JWT`.

As agreed, this is unconditional, with no config flag.

**Merge gating:** this is the breaking half, so per your note it wants a minor rather than a patch,
and ideally lands after `rauthy-client v0.14.3` (#1650) is released.

**Backwards compatibility:** validation accepts both `JWT` and `at+jwt` regardless of token type, so
access tokens issued before an update stay valid until they expire. The check is deliberately not
tightened to require `at+jwt` when `expected_type == Bearer`, since that would reject still-valid
pre-update tokens.

**Downstream:** only breaking for a resource server that checks the header `typ` for exactly `JWT`.
`rauthy-client` accepts `at+jwt` from `v0.14.3` (#1650), so releasing that first avoids any
interruption in service.

**One design note:** the header `typ` is modelled as a small `JwtHeaderTyp` enum rather than a bare
string parameter. `JwtToken::build` is generic over the claims type and cannot infer the token kind,
so the value has to be passed at all 4 call sites; an enum makes it impossible to typo, and keeps it
clearly apart from the `typ` *claim* in the payload, which is Rauthy's own token type with different
values (`Bearer` / `Id` / `Refresh` / ...) and is untouched here. Happy to switch it to a plain
`&'static str` if you prefer less surface.

### Verification

- `cargo check --workspace`, `clippy`, `fmt --check` clean.
- `just test-hiqlite` full suite green with `at+jwt` live, so introspection, userinfo, revocation
  and refresh all still validate.
- `test_authorization_code_flow` extended to assert the headers positively: access token `at+jwt`,
  ID token `JWT`. Header construction had no test coverage before.

### DIFF
--- a/src/jwt/src/token.rs
+++ b/src/jwt/src/token.rs
@@ -12,8 +12,27 @@ use tracing::warn;
 pub struct JwtHeader<'a> {
     pub alg: JwkKeyPairAlg,
     pub kid: &'a str,
-    // should always be JWT -> DPoP tokens have their own validation
+    // either `JWT` or `at+jwt` -> DPoP tokens have their own validation
     pub typ: &'a str,
+}
+
+/// The JWT header `typ`. Not to be confused with the `typ` claim inside the payload, which is
+/// Rauthy's own token type and has different values.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum JwtHeaderTyp {
+    /// The generic `JWT` for ID, Refresh and Logout Tokens.
+    Jwt,
+    /// `at+jwt` for Access Tokens as specified in RFC 9068.
+    AtJwt,
+}
+
+impl JwtHeaderTyp {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Jwt => "JWT",
+            Self::AtJwt => "at+jwt",
+        }
+    }
 }
 
 pub struct JwtToken;
@@ -22,13 +41,15 @@ impl JwtToken {
     pub fn build<C: Debug + Serialize>(
         jwk: &JwkKeyPair,
         claims: &C,
+        typ: JwtHeaderTyp,
     ) -> Result<String, ErrorResponse> {
         let mut token = String::with_capacity(1024);
 
         let header = format!(
-            "{{\"alg\":\"{}\",\"kid\":\"{}\",\"typ\":\"JWT\"}}",
+            "{{\"alg\":\"{}\",\"kid\":\"{}\",\"typ\":\"{}\"}}",
             jwk.typ.as_str(),
-            jwk.kid
+            jwk.kid,
+            typ.as_str()
         );
         base64_url_no_pad_encode_buf(header.as_bytes(), &mut token);
         token.push('.');
@@ -94,7 +115,10 @@ impl JwtToken {
 
         base64_url_no_pad_decode_buf(header, buf)?;
         let header = serde_json::from_slice::<JwtHeader>(buf)?;
-        if header.typ != "JWT" {
+        // Access Tokens use `at+jwt` (RFC 9068), everything else `JWT`. Both must be accepted
+        // independently of the token type, because Access Tokens issued before that change are
+        // still valid until they expire.
+        if header.typ != JwtHeaderTyp::Jwt.as_str() && header.typ != JwtHeaderTyp::AtJwt.as_str() {
             return Err(ErrorResponse::new(
                 ErrorResponseType::BadRequest,
                 "Invalid JWT Header `typ`",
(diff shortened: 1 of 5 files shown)


src/jwt/src/token.rs
```diff
@@ -12,23 +12,44 @@ use tracing::warn;
...
+/// Rauthy's own token type and has different values.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum JwtHeaderTyp {
+    /// The generic `JWT` for ID, Refresh and Logout Tokens.
+    Jwt,
+    /// `at+jwt` for Access Tokens as specified in RFC 9068.
+    AtJwt,
+}
+
+impl JwtHeaderTyp {
```
sebadob: I would name that `JwtHeaderType` intead of `JwtHeaderTyp`. It's only a `typ` inside the claims because `type` is usually always a reserved name.

src/service/src/oidc/bcl_logout_token.rs
```diff
@@ -98,7 +98,7 @@ impl LogoutToken<'_> {
             sid: self.sid,
             nonce: None,
         };
-        rauthy_jwt::token::JwtToken::build(kp, &claims)
+        rauthy_jwt::token::JwtToken::build(kp, &claims, rauthy_jwt::token::JwtHeaderTyp::Jwt)
```
sebadob: For all `std` or internal dependencies, we should import everything directly, so you can use `JwtHeaderTyp` directly.

sebadob (changes requested):

src/jwt/src/token.rs
bartekus: Done in fea7d3ff, both points: the enum is now `JwtHeaderType`, and `bcl_logout_token.rs` imports `JwtHeaderType` and `JwtToken` directly instead of path-qualifying them (the `JwtToken` qualification predated this PR, but it was on the line I was touching anyway).

sebadob (dismissed):
Looks good now. I will delay the merge a bit since it comes with a breaking change.

bartekus:
Heads up: this went conflicting after the merges of the last two days, so I rebased it onto current `main`. That dismissed your approval, which I did not expect it to, so it needs a re-approve. Sorry for the extra click. The conflict was the CHANGELOG only, and it needed doing either way since a conflicting PR cannot be merged at all.

Re-verified on the new base: `cargo clippy --all-features -- -D warnings` and `cargo fmt --check` clean, full integration suite green on hiqlite.

Also worth checking before you merge: `rauthy-client` on crates.io is still at `0.14.2`, so the `0.14.3` the changelog points readers at is not published yet.

sebadob:
> Also worth checking before you merge: rauthy-client on crates.io is still at 0.14.2, so the 0.14.3 the changelog points readers at is not published yet.

Yes I know. I probably want to include a preparation for #1663 as well.

bartekus:
Understood, and that makes sense as a bundle. Nothing needed from me on this PR then; it is rebased, green and mergeable whenever you want it, and I will leave it alone rather than churn the branch.

If #1663 is not already in progress on your side I am happy to take it. Reading your comment there, the shape is: add `Ed25519` as a selectable client `alg` that maps onto the same ed25519 key material and only changes what goes in the token header, keep `EdDSA` as the default, and make the JWKS representation switchable via a config default plus the request-param override you already use for the Cloudflare case, leaving `Ed448` alone. If that is what you have in mind, say so and I will open it against that design. If you would rather keep it yourself because it is tied to the client release, that is completely fine and I will stay out of it.

PR #1652, opened 2026-07-17
feat(oidc): RFC 8693 token exchange

### LINKED ISSUE #1639: RFC 8693 OAuth 2.0 Token Exchange
## Summary

Rauthy today issues tokens through five grants (`authorization_code`,
`client_credentials`, `password`, `refresh_token`, device code). None of them
lets one party obtain a token to act **on behalf of** another. That is the gap
RFC 8693 fills: a service that holds a token for a user can exchange it for a
new, narrower token to call a downstream service, carrying a verifiable record
of who is acting and for whom.

This proposes an opt-in `urn:ietf:params:oauth:grant-type:token-exchange` grant,
deny-by-default per client, that reuses machinery already in tree: the grant
dispatch in `service/src/oidc/mod.rs`, the JWT validation path, and the RFC 8707
`resource` / `default_aud` audience work from #1608. No new endpoint.

Grounded against `main` at v0.36.0.

## 1. Problem statement

The common shape is a chain of services: a client calls service A with a token,
and A must call service B on the user's behalf. Today A has only bad options:
forward its own inbound token verbatim (B now holds a token minted for A, with
A's full audience and scope, and no record that A is the caller), or ask the user
to log in again against B. Neither is safe or ergonomic.

RFC 8693 gives A a first-class move: present the inbound token to the AS and
receive a new token that is

- **audience-restricted** to B (nothing else can accept it), and
- **scope-narrowed** to what the call to B needs, and
- **attributed**: in the delegation case, the new token records both the user
  (`sub`) and the acting service (`act`), so B can make decisions on the full
  chain.

This is the standard building block behind API gateways, service meshes, and
downstream-API calls. It is also exactly what an MCP tool server needs when it
calls another protected resource on behalf of the logged-in user (the #1562
"Rauthy as the AS for MCP" line of work): the tool server exchanges the user's
token for one audience-bound to the next hop. Rauthy already grew RFC 8707
audience binding (#1608) for the same ecosystem; token exchange is the grant that
puts it to use.

## 2. This builds on machinery that already ships

- **The grant dispatch is a clean extension point.** `get_token_set`
  (`service/src/oidc/mod.rs:31-38`) matches `grant_type` and calls one
  `grant_type_*` per grant, each in its own `grant_types/*.rs`. Token exchange is
  one new arm plus a `grant_types/token_exchange.rs` sibling. No new endpoint,
  no new route.
- **Subject-token validation already exists.** Rauthy validates its own JWTs
  (signature, `iss`, `exp`, `nbf`, `typ`) on every request; the same path
  validates the presented `subject_token`.
- **Audience restriction is already built.** #1608 added per-client
  `allowed_resources` + `default_aud` and the `aud` composition logic in
  `token_set.rs` (single string vs array). Token exchange sets the target
  audience through the exact same path: the `resource` / `audience` request
  parameters, validated against the requesting client's `allowed_resources`, land
  in the new token's `aud`.
- **Deny-by-default per client is the house pattern.** Ephemeral-client resource
  validation (#1608) and the attestation discussion (#1618) both default to deny.
  Token exchange, being privileged, follows the same stance: a client cannot
  exchange unless an operator turns it on.

### 3.1 The grant

New arm in `get_token_set`:

```
"urn:ietf:params:oauth:grant-type:token-exchange" => grant_type_token_exchange(req, req_data).await,
```

Request parameters (RFC 8693 §2.1), added to the token request type:

- `subject_token` (required): the token representing the party to act for.
- `subject_token_type` (required): `urn:ietf:params:oauth:token-type:access_token`
  in v1 (id-token support can follow).
- `actor_token` / `actor_token_type` (optional): present => delegation, absent =>
  impersonation.
- `resource` / `audience` (optional): the target service, validated exactly as
  the #1608 `resource` parameter is.
- `scope` (optional): downscope only; must be a subset of the subject token's
  scope.
- `requested_token_type` (optional): `...:token-type:access_token` in v1
  (default).

Response is the RFC 8693 §2.2 body: `access_token`, `issued_token_type`,
`token_type`, `expires_in`, and `scope` when narrowed.

### 3.2 Impersonation vs delegation

- **Impersonation** (`subject_token` only): the new token keeps the subject's
  `sub`; no `act` claim. The caller becomes the subject toward the target.
- **Delegation** (`subject_token` + `actor_token`): the new token keeps the
  subject's `sub` and adds the RFC 8693 §4.1 `act` claim naming the actor
  (`"act": { "sub": "<actor>" }`), nestable for multi-hop chains. The target sees
  both who the call is for and who is making it.

### 3.3 Authorization (the one real security call)

Token exchange lets a client obtain a token for another subject, so it must be
gated. Proposal, deny-by-default:

- a per-client boolean (working name `token_exchange`, default `false`); only a
  client with it set may use the grant;
- the target audience is constrained by the client's existing `allowed_resources`
  allow-list from #1608 (no new column for the target set: an exchange to a
  `resource` not in the client's allow-list returns `invalid_target`);
- `scope` may only narrow, never widen, relative to the subject token.

### 3.4 Errors

Standard RFC 8693 / OAuth errors through `ErrorResponse`: `invalid_request`
(missing/!supported token type), `invalid_target` (audience not permitted, reuses
#1608), `invalid_grant` (subject token invalid/expired), `unauthorized_client`
(client not enabled for exchange).

## 4. Semantics that need a ruling

These are the calls I would not make unilaterally:

**(a) Subject token types in v1.** I propose accepting a Rauthy-issued
`access_token` only, validated through the existing JWT path. Adding
`id_token` (and the generic `...:token-type:jwt`) is easy to follow with, but I
would keep v1 to access tokens. Agree, or include id tokens now?

**(b) Impersonation + delegation, or impersonation first.** I lean on shipping
both (the `act` claim is small and is the more useful half for service chains),
but impersonation-only is a smaller v1 if you prefer.

**(c) Authorization model.** Per-client `token_exchange` boolean + reuse of
`allowed_resources` for the permitted targets (3.3), versus something coarser
(a global config) or finer (a dedicated target allow-list column). I lean on the
boolean + `allowed_resources` reuse: no new column, deny-by-default.

**(d) Cross-client trust.** Whether the requesting client may exchange a
`subject_token` that was issued to a *different* client. v1 proposal: yes, as
long as the token is a valid Rauthy token and the requesting client is enabled
and the target is in its allow-list; the audience narrowing is what contains it.
Open to restricting to same-client in v1 if you would rather.

**(e) `may_act`.** RFC 8693 §4.4 lets a subject token carry a `may_act` claim
naming who is allowed to act for it. I would treat honoring `may_act` as a v2
refinement (v1 authorizes purely on the requesting client), unless you want it in
v1.

**(f) Refresh tokens.** Whether an exchanged access token may itself carry a
refresh token. I propose no in v1 (exchanged tokens are short-lived and
re-derived on demand), which also sidesteps refresh-chain questions.

## 5. Surface impact (brief, factual)

- **Dispatch:** one arm in `service/src/oidc/mod.rs`, one new
  `grant_types/token_exchange.rs`.
- **Request type:** the token request struct gains the RFC 8693 optional fields
  (no `deny_unknown_fields` today, so back-compatible).
- **Discovery:** `grant_types_supported` in `well_known.rs` grows from 5 to 6
  entries (the exchange URN).
- **Schema:** one nullable per-client boolean (`token_exchange`), paired hiqlite
  + postgres migrations; reuses #1608 `allowed_resources` for targets, so no
  audience column is added.
- **Claims:** the `act` claim in the delegation path; `aud` composition is the
  existing #1608 path.
- **Tests:** `zz*` handler tests on both DBs (impersonation happy path,
  delegation with `act`, `invalid_target`, `unauthorized_client`, scope-narrowing,
  deny-by-default), a Book section, CHANGELOG.

## 6. Why this is the minimal mechanism

- Rides the existing grant dispatch: no new endpoint, no new route.
- Reuses #1608 audience binding verbatim for the target: no new audience column
  and no second audience code path.
- Reuses the shipped JWT validation for the subject token.
- Deny-by-default per client, matching the existing ephemeral-resource and
  attestation stances.
- Entirely opt-in: a client without `token_exchange` set behaves exactly as
  today, and the grant is simply unavailable to it.

## 7. Ask

Before I open a PR, could you ack (or redirect) on:

1. The opt-in `token-exchange` grant as a new dispatch arm, no new endpoint (3.1).
2. Shipping both impersonation and delegation with the `act` claim (3.2 / 4b).
3. The authorization model: per-client `token_exchange` boolean + reuse of
   `allowed_resources` for permitted targets, deny-by-default (3.3 / 4c).
4. Access-token-only subject tokens in v1 (4a), and no refresh token on the
   exchanged token (4f).
5. `may_act` deferred to v2 (4e).

Happy to split it (the grant + impersonation first, delegation `act` second) if
you prefer smaller PRs.

bartekus:
Closes #1639. Implemented exactly to your calls in #1639: no new column, both impersonation and
delegation with `act`, deny-by-default, `allowed_resources` reused for the targets, access tokens
only, no refresh token, `may_act` deferred. Single PR as you suggested.

### The grant

`urn:ietf:params:oauth:grant-type:token-exchange` is one new arm in `get_token_set`, plus a
`grant_types/token_exchange.rs` sibling modelled on `client_credentials.rs`. Client auth, DPoP and
the dynamic-client bookkeeping are the same as there, and only confidential clients may exchange.

**No schema change, no migration.** The flow lives in the existing `flows_enabled` CSV, and the
target allow-list is the existing `allowed_resources`, so `validate_resource_request` gives
deny-by-default and `invalid_target` for free.

### Four calls I made, each easy to change

**1. The `flows_enabled` value is the grant type URN, not `token_exchange`.** I started with the
underscore name but that forces `RE_GRANT_TYPES` to accept both spellings, since the same regex
validates request `grant_type`s and stored flow values. The device code flow already stores the URN
in `flows_enabled` (`bootstrap/clients.json`) and calls `validate_flow(GRANT_TYPE_DEVICE_CODE)`, so
using the URN follows that precedent, needs exactly one new regex alternative, and keeps the two
namespaces from mixing. Say the word if you'd rather have `token_exchange` in the CSV.

**2. `"act"` added to `RESERVED_ROOT_CLAIMS`.** This one is security relevant and not just
cosmetic: without it, a scope flagged `claims_at_root` with a custom attribute named `act` would
silently shadow or duplicate the delegation claim, which is exactly what
`validate_no_reserved_collision` exists to prevent. It is a one-line addition, but it belongs in
this PR rather than a follow-up.

**3. A new `InvalidGrant` error variant.** RFC 8693 §2.2.2 wants RFC 6749 §5.2 codes for a rejected
`subject_token`. `InvalidTarget` was the only variant with a `#[serde(rename)]`; this follows that
precedent so a bad subject token answers `invalid_grant` instead of `BadRequest`. Happy to drop it
and reuse `BadRequest` if you'd rather not grow the enum.

**4. Both `audience` and `resource` are accepted as the target**, and both route through
`validate_resource_request`. RFC 8693 names it `audience`, but `resource` is already on a token
request and means the same thing here. Giving both at once is an error rather than a silent
precedence rule.

**One extra:** `TokenRequest` got a `#[derive(Default)]` and the tests now use
`..Default::default()`. Adding 7 fields otherwise meant touching 16 literals across 6 test files
with explicit `None`s, and every future field would do the same. This deletes ~110 lines of
boilerplate and no test lost a real value (only `None` lines were removed).

### Behavior

- Impersonation (no `actor_token`): the token looks like the subject's, no `act`.
- Delegation (`actor_token`): adds `act.sub`; if the actor's own token had an `act`, that chain
  stays nested, per §4.1.
- Scopes may only narrow the subject token's, never widen.
- A revoked or non-access `subject_token` is rejected with `invalid_grant`.
- The exchanged token never carries a refresh or id token. Note `from_user` would have minted and
  persisted a `RefreshToken` row, so the new path calls `build_access_token` directly instead of
  nulling the field afterwards.

### Verification

- `cargo fmt --check`, `clippy --workspace --tests` clean (no new warnings; the workspace count is
  unchanged apart from one summary line for the new test binary).
- Full suite green on **both** `just test-hiqlite` and `just test-postgres`.
- New `zzg_handler_token_exchange.rs` with 10 blocks: deny-by-default, impersonation, delegation
  `act`, `invalid_target`, `actor_token` without its type, refresh-token-as-subject, garbage token,
  unsupported `subject_token_type`, scope widening rejected, scope narrowing accepted, plus the
  no-refresh/no-id-token assertions.
- It deliberately mints its own subject token rather than using `common::get_token_set_init_client`,
  because `zzd_handler_clients` rotates the `init_client` secret and anything ordered after it gets
  a stale secret. Worth knowing for future `zz*` tests.
- Docs: new `book/src/work/token_exchange.md`, and `resource_indicators.md` amended, since
  `allowed_resources` is now dual purpose and its text said otherwise.
- Admin UI: a `token_exchange` checkbox next to the other flows (no i18n needed, the flow labels are
  literals).

### DIFF
--- a/src/api_types/src/oidc.rs
+++ b/src/api_types/src/oidc.rs
@@ -4,17 +4,91 @@ use crate::sessions::SessionState;
 use actix_web::HttpRequest;
 use actix_web::http::header;
 use rauthy_common::regex::{
-    RE_ALNUM, RE_BASE64, RE_CLIENT_ID, RE_CODE_CHALLENGE_METHOD, RE_CODE_VERIFIER, RE_GRANT_TYPES,
-    RE_LOWERCASE, RE_RESOURCE, RE_SCOPE_SPACE, RE_URI,
+    RE_ALNUM, RE_BASE64, RE_CLIENT_ID, RE_CODE_CHALLENGE_METHOD, RE_CODE_VERIFIER, RE_LOWERCASE,
+    RE_RESOURCE, RE_SCOPE_SPACE, RE_URI,
 };
 use rauthy_common::utils::base64_decode;
 use rauthy_error::{ErrorResponse, ErrorResponseType};
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 use std::fmt::{Display, Formatter};
+use std::str::FromStr;
 use time::OffsetDateTime;
 use utoipa::{IntoParams, ToSchema};
 use validator::Validate;
+
+/// The OAuth 2.0 / OIDC grant types (flows) Rauthy supports.
+///
+/// The wire format is the grant type identifier itself, which for the extended grants is the
+/// full URN. Keeping this an enum rather than a validated `String` makes it type-safe in code
+/// and self-documenting in the OpenAPI spec.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+pub enum GrantType {
+    #[default]
+    #[serde(rename = "authorization_code")]
+    AuthorizationCode,
+    #[serde(rename = "client_credentials")]
+    ClientCredentials,
+    #[serde(rename = "password")]
+    Password,
+    #[serde(rename = "refresh_token")]
+    RefreshToken,
+    #[serde(rename = "urn:ietf:params:oauth:grant-type:device_code")]
+    DeviceCode,
+    #[serde(rename = "urn:ietf:params:oauth:grant-type:token-exchange")]
+    TokenExchange,
+}
+
+impl GrantType {
+    #[inline]
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::AuthorizationCode => "authorization_code",
+            Self::ClientCredentials => "client_credentials",
+            Self::Password => "password",
+            Self::RefreshToken => "refresh_token",
+            Self::DeviceCode => "urn:ietf:params:oauth:grant-type:device_code",
+            Self::TokenExchange => "urn:ietf:params:oauth:grant-type:token-exchange",
+        }
+    }
+
+    /// Joins the given flows into the CSV format used for `clients.flows_enabled`.
+    pub fn csv(flows: &[Self]) -> String {
+        flows
+            .iter()
+            .map(|f| f.as_str())
+            .collect::<Vec<_>>()
+            .join(",")
+    }
+}
+
+impl FromStr for GrantType {
+    type Err = ErrorResponse;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let slf = match s {
+            "authorization_code" => Self::AuthorizationCode,
+            "client_credentials" => Self::ClientCredentials,
+            "password" => Self::Password,
+            "refresh_token" => Self::RefreshToken,
+            "urn:ietf:params:oauth:grant-type:device_code" => Self::DeviceCode,
+            "urn:ietf:params:oauth:grant-type:token-exchange" => Self::TokenExchange,
+            _ => {
+                return Err(ErrorResponse::new(
+                    ErrorResponseType::BadRequest,
+                    format!("Invalid `grant_type`: {s}"),
+                ));
+            }
+        };
+        Ok(slf)
+    }
+}
+
+impl Display for GrantType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
 
 #[derive(Serialize, Deserialize, ToSchema)]
 pub struct AddressClaim {
@@ -235,12 +309,7 @@ pub struct DeviceVerifyRequest {
 #[derive(Default, Deserialize, Validate, ToSchema)]
 #[cfg_attr(debug_assertions, derive(Serialize))]
 pub struct TokenRequest {
-    /// Validation: `^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|urn:ietf:params:oauth:grant-type:token-exchange|password|refresh_token)$`
-    #[validate(regex(
-        path = "*RE_GRANT_TYPES",
-        code = "^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|urn:ietf:params:oauth:grant-type:token-exchange|password|refresh_token)$"
-    ))]
-    pub grant_type: String,
+    pub grant_type: GrantType,
     /// Validation: `[a-zA-Z0-9]`
     #[validate(regex(path = "*RE_ALNUM", code = "[a-zA-Z0-9]"))]
     pub code: Option<String>,
--- a/src/api_types/src/oidc.rs
+++ b/src/api_types/src/oidc.rs
@@ -232,13 +232,13 @@ pub struct DeviceVerifyRequest {
     pub device_accepted: DeviceAcceptedRequest,
 }
 
-#[derive(Deserialize, Validate, ToSchema)]
+#[derive(Default, Deserialize, Validate, ToSchema)]
 #[cfg_attr(debug_assertions, derive(Serialize))]
 pub struct TokenRequest {
-    /// Validation: `^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|password|refresh_token)$`
+    /// Validation: `^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|urn:ietf:params:oauth:grant-type:token-exchange|password|refresh_token)$`
     #[validate(regex(
         path = "*RE_GRANT_TYPES",
-        code = "^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|password|refresh_token)$"
+        code = "^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|urn:ietf:params:oauth:grant-type:token-exchange|password|refresh_token)$"
     ))]
     pub grant_type: String,
     /// Validation: `[a-zA-Z0-9]`
@@ -278,6 +278,50 @@ pub struct TokenRequest {
     /// Validation: `[a-zA-Z0-9,.:/_-&?=~!$'()*+%@]+$` (no `#`; RFC 8707 forbids a fragment)
     #[validate(regex(path = "*RE_RESOURCE", code = "[a-zA-Z0-9,.:/_-&?=~!$'()*+%@]+$"))]
     pub resource: Option<String>,
+
+    /// RFC 8693 token exchange: the token that represents the identity on whose behalf the
+    /// request is made. Rauthy only accepts an access token here.
+    ///
+    /// Validation: `[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$`
+    #[validate(regex(path = "*RE_URI", code = "[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$"))]
+    pub subject_token: Option<String>,
+    /// RFC 8693 token exchange: the type of the `subject_token`. Rauthy only accepts
+    /// `urn:ietf:params:oauth:token-type:access_token`.
+    ///
+    /// Validation: max length is 256
+    #[validate(length(max = 256))]
+    pub subject_token_type: Option<String>,
+    /// RFC 8693 token exchange: the token that represents the identity of the acting party.
+    /// If given, the exchanged token is a delegation and will contain an `act` claim.
+    /// Rauthy only accepts an access token here.
+    ///
+    /// Validation: `[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$`
+    #[validate(regex(path = "*RE_URI", code = "[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$"))]
+    pub actor_token: Option<String>,
+    /// RFC 8693 token exchange: the type of the `actor_token`. Rauthy only accepts
+    /// `urn:ietf:params:oauth:token-type:access_token`.
+    ///
+    /// Validation: max length is 256
+    #[validate(length(max = 256))]
+    pub actor_token_type: Option<String>,
+    /// RFC 8693 token exchange: the type of token the client wants back. Rauthy only issues
+    /// `urn:ietf:params:oauth:token-type:access_token`, which is also the default.
+    ///
+    /// Validation: max length is 256
+    #[validate(length(max = 256))]
+    pub requested_token_type: Option<String>,
+    /// RFC 8693 token exchange: the target of the exchanged token. Validated against the
+    /// client's `allowed_resources`, exactly like `resource`.
+    ///
+    /// Validation: `[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$`
+    #[validate(regex(path = "*RE_URI", code = "[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$"))]
+    pub audience: Option<String>,
+    /// RFC 8693 token exchange: the scopes for the exchanged token. May only narrow the
+    /// scopes of the `subject_token`, never widen them.
+    ///
+    /// Validation: `[a-zA-Z0-9-_/:\s*]{0,512}`
+    #[validate(regex(path = "*RE_SCOPE_SPACE", code = "[a-zA-Z0-9-_/:\\s*]{0,512}"))]
+    pub scope: Option<String>,
 }
 
 impl TokenRequest {
--- a/src/api_types/src/oidc.rs
+++ b/src/api_types/src/oidc.rs
@@ -90,6 +90,76 @@ impl Display for GrantType {
     }
 }
 
+/// The RFC 8693 token type identifiers (RFC 8693 §3). These are a different URN namespace from
+/// [`GrantType`]: `urn:ietf:params:oauth:token-type:*` rather than
+/// `urn:ietf:params:oauth:grant-type:*`, so the two are not interchangeable.
+///
+/// Rauthy only ever accepts and issues an access token, but every RFC-defined type is listed so
+/// that "a token type we do not support" can be told apart from "not a token type at all".
+///
+/// Unlike [`GrantType`], this is deliberately *not* the wire type of `subject_token_type`,
+/// `actor_token_type` and `requested_token_type`. Those stay `Option<String>` and are parsed
+/// here in the handler: RFC 8693 §2.2.2 requires the token endpoint to answer with an OAuth
+/// error object, and rejecting at `serde` level would replace ours with a generic
+/// deserialization error before the handler ever runs.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+pub enum TokenType {
+    #[serde(rename = "urn:ietf:params:oauth:token-type:access_token")]
+    AccessToken,
+    #[serde(rename = "urn:ietf:params:oauth:token-type:refresh_token")]
+    RefreshToken,
+    #[serde(rename = "urn:ietf:params:oauth:token-type:id_token")]
+    IdToken,
+    #[serde(rename = "urn:ietf:params:oauth:token-type:saml1")]
+    Saml1,
+    #[serde(rename = "urn:ietf:params:oauth:token-type:saml2")]
+    Saml2,
+    #[serde(rename = "urn:ietf:params:oauth:token-type:jwt")]
+    Jwt,
+}
+
+impl TokenType {
+    #[inline]
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::AccessToken => "urn:ietf:params:oauth:token-type:access_token",
+            Self::RefreshToken => "urn:ietf:params:oauth:token-type:refresh_token",
+            Self::IdToken => "urn:ietf:params:oauth:token-type:id_token",
+            Self::Saml1 => "urn:ietf:params:oauth:token-type:saml1",
+            Self::Saml2 => "urn:ietf:params:oauth:token-type:saml2",
+            Self::Jwt => "urn:ietf:params:oauth:token-type:jwt",
+        }
+    }
+}
+
+impl FromStr for TokenType {
+    type Err = ErrorResponse;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let slf = match s {
+            "urn:ietf:params:oauth:token-type:access_token" => Self::AccessToken,
+            "urn:ietf:params:oauth:token-type:refresh_token" => Self::RefreshToken,
+            "urn:ietf:params:oauth:token-type:id_token" => Self::IdToken,
+            "urn:ietf:params:oauth:token-type:saml1" => Self::Saml1,
+            "urn:ietf:params:oauth:token-type:saml2" => Self::Saml2,
+            "urn:ietf:params:oauth:token-type:jwt" => Self::Jwt,
+            _ => {
+                return Err(ErrorResponse::new(
+                    ErrorResponseType::BadRequest,
+                    format!("Invalid token type: {s}"),
+                ));
+            }
+        };
+        Ok(slf)
+    }
+}
+
+impl Display for TokenType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
 #[derive(Serialize, Deserialize, ToSchema)]
 pub struct AddressClaim {
     pub formatted: String,
--- a/src/common/src/constants.rs
+++ b/src/common/src/constants.rs
@@ -76,7 +76,6 @@ pub static DEVICE_KEY_LENGTH: u8 = 64;
 /// Max serialized length of a client's custom `claims` JSON object.
 pub const CLIENT_CLAIMS_MAX_LEN: usize = 1024;
 pub static EVENTS_LATEST_LIMIT: u16 = 100;
-/// RFC 8693 token type identifier for an OAuth 2.0 access token.
 pub const UPSTREAM_AUTH_CALLBACK_TIMEOUT_SECS: u16 = 300;
 pub const CACHE_TTL_APP: Option<i64> = Some(43200);
 pub const CACHE_TTL_AUTH_PROVIDER_CALLBACK: Option<i64> =
--- a/src/common/src/constants.rs
+++ b/src/common/src/constants.rs
@@ -76,8 +76,6 @@ pub static DEVICE_KEY_LENGTH: u8 = 64;
 /// Max serialized length of a client's custom `claims` JSON object.
 pub const CLIENT_CLAIMS_MAX_LEN: usize = 1024;
 pub static EVENTS_LATEST_LIMIT: u16 = 100;
-pub static GRANT_TYPE_DEVICE_CODE: &str = "urn:ietf:params:oauth:grant-type:device_code";
-pub static GRANT_TYPE_TOKEN_EXCHANGE: &str = "urn:ietf:params:oauth:grant-type:token-exchange";
 /// RFC 8693 token type identifier for an OAuth 2.0 access token.
 pub static TOKEN_TYPE_ACCESS_TOKEN: &str = "urn:ietf:params:oauth:token-type:access_token";
 pub const UPSTREAM_AUTH_CALLBACK_TIMEOUT_SECS: u16 = 300;
--- a/src/common/src/constants.rs
+++ b/src/common/src/constants.rs
@@ -77,6 +77,9 @@ pub static DEVICE_KEY_LENGTH: u8 = 64;
 pub const CLIENT_CLAIMS_MAX_LEN: usize = 1024;
 pub static EVENTS_LATEST_LIMIT: u16 = 100;
 pub static GRANT_TYPE_DEVICE_CODE: &str = "urn:ietf:params:oauth:grant-type:device_code";
+pub static GRANT_TYPE_TOKEN_EXCHANGE: &str = "urn:ietf:params:oauth:grant-type:token-exchange";
+/// RFC 8693 token type identifier for an OAuth 2.0 access token.
+pub static TOKEN_TYPE_ACCESS_TOKEN: &str = "urn:ietf:params:oauth:token-type:access_token";
 pub const UPSTREAM_AUTH_CALLBACK_TIMEOUT_SECS: u16 = 300;
 pub const CACHE_TTL_APP: Option<i64> = Some(43200);
 pub const CACHE_TTL_AUTH_PROVIDER_CALLBACK: Option<i64> =
--- a/src/common/src/constants.rs
+++ b/src/common/src/constants.rs
@@ -77,7 +77,6 @@ pub static DEVICE_KEY_LENGTH: u8 = 64;
 pub const CLIENT_CLAIMS_MAX_LEN: usize = 1024;
 pub static EVENTS_LATEST_LIMIT: u16 = 100;
 /// RFC 8693 token type identifier for an OAuth 2.0 access token.
-pub static TOKEN_TYPE_ACCESS_TOKEN: &str = "urn:ietf:params:oauth:token-type:access_token";
 pub const UPSTREAM_AUTH_CALLBACK_TIMEOUT_SECS: u16 = 300;
 pub const CACHE_TTL_APP: Option<i64> = Some(43200);
 pub const CACHE_TTL_AUTH_PROVIDER_CALLBACK: Option<i64> =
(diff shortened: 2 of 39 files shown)


src/api_types/src/oidc.rs
```diff
@@ -272,6 +272,50 @@ pub struct TokenRequest {
...
+    ///
+    /// Validation: `[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$`
+    #[validate(regex(path = "*RE_URI", code = "[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$"))]
+    pub subject_token: Option<String>,
+    /// RFC 8693 token exchange: the type of the `subject_token`. Rauthy only accepts
+    /// `urn:ietf:params:oauth:token-type:access_token`.
+    ///
+    /// Validation: max length is 256
+    #[validate(length(max = 256))]
+    pub subject_token_type: Option<String>,
```
sebadob, line 432: We are using these types / flows in quite a few places by now. I guess it makes sense to create an enum `GrantType` with all the flows. The extended ones simplified and use `serde` rename, and possibiliy impl a `as_str()` method is necessary. That makes everything a bit easier to handle in the code, and it also automatically is a type-safe documentation for the OpenAPI spec.

We can then also drop the regex validation. An enum would be a lot faster and more efficient.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,253 @@
...
+/// identity the new token should represent. Without an `actor_token` this is an impersonation:
+/// the new token looks like a token for the subject. With an `actor_token` it is a delegation:
+/// the new token additionally carries an `act` claim naming the acting party, so a resource
+/// server can tell "A acting for B" apart from "B".
+#[tracing::instrument(skip_all, fields(client_id = req_data.client_id))]
+pub async fn grant_type_token_exchange(
+    req: HttpRequest,
+    req_data: TokenRequest,
+) -> Result<(TokenSet, Vec<(HeaderName, HeaderValue)>), ErrorResponse> {
+    if req_data.client_secret.is_none() {
```
sebadob: I have not looked into the details of the RFC, but that check could lead to issues. By default (at least for OIDC), A client can authenticate either via body params or Authorization Basic header. 

A more robust way of doing it would be to theck the extracted `client_secret` below, since the `try_get_client_id_secret()` works for both versions. If however the RFC requires it in the body, this check makes sense of course.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,253 @@
...
+                    HeaderValue::from_str(nonce)?,
+                ));
+            }
+            Some(DpopFingerprint(proof.jwk_fingerprint()?))
+        } else {
+            None
+        };
+
+    if client.is_dynamic() {
+        ClientDyn::update_used(&client.id).await?;
```
sebadob: I guess if a client is dynamic, we should reject the request, and even do that check as soon as we have the client. This exchange is validated against the allowed resources, which a dyn client is not allowed to set to prevent priv esc.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,253 @@
...
+    // By RFC, `subject_token_type` is REQUIRED.
+    let Some(subject_token_type) = req_data.subject_token_type.as_deref() else {
+        return Err(ErrorResponse::new(
+            ErrorResponseType::BadRequest,
+            "'subject_token_type' is missing",
+        ));
+    };
+    // The claims borrow from these buffers, so they must live as long as the claims do.
+    let mut subject_buf = Vec::with_capacity(512);
+    let mut actor_buf = Vec::with_capacity(512);
```
sebadob: We don't really need 2 buffers here. We can pretty easily get away with only one, which is the reason the custom JWT validator takes it as input.

From the `subject_claims` we need 2 values later on:
1. The `common.scope`. This is only needed as borrowed data, but when the allocate before checking `act`, we only allocate a single `String` instead of a pretty big `Vec<_>` with 512 bytes upfront. If no scope narrowing is required, this is also allocated into a `String` anyway. 
2. `common.sub`, which is then needed as a `String` anyway, which means even when we borrow early, an allocation is made below anyway.

If we extract these 2 values upfront (which allocate later on anyway), we can re-use the same buffer for the `act` validation.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,253 @@
...
+                ErrorResponseType::InvalidTarget,
+                "'audience' and 'resource' must not be given at the same time",
+            ));
+        }
+        (Some(audience), None) => Some(audience),
+        (None, Some(resource)) => Some(resource),
+        (None, None) => None,
+    };
+    if let Some(target) = target {
+        client.validate_resource_request(target)?;
```
sebadob, line 166: I just noticed that when adding the `resource` claim, we should probably add a check to `client.validate_resource_request()` that makes sure it's not a dynamic client. It should not be possible for a dyn client to add custom resources anyway, but an additional (very quick) check does not hurt and makes it more robust if anything about it changes in the future, so it cannot be forgotten.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    // The exchanged token can only ever narrow the subject's scopes, never widen them.
+    let subject_scope = subject_claims.common.scope.as_deref().unwrap_or_default();
+    let scope = if let Some(requested) = req_data.scope.as_deref() {
+        let mut narrowed = Vec::new();
```
sebadob: We can do an educated guess here, and start with a capacity for `narrowed` of at least 2. If scope is narrowed, there is a least a single one, and when Rust increases the size from `0` it reserves more than that already.
Apart from that, `openid` should always be added. No matter if requested or not. This is a mandatory scope for OIDC.

This is a tiny improvement, but gets rid of at least 1 syscall under the hood.

src/service/src/token_set.rs
```diff
@@ -565,6 +567,84 @@ impl TokenSet {
...
+                }
+            }
+
+            attrs = if customs_access.is_empty() {
+                None
+            } else {
+                let attrs = UserAttrValueEntity::find_for_user_with_defaults(&user.id).await?;
+                let mut res = HashMap::with_capacity(attrs.len());
+                attrs.iter().for_each(|a| {
+                    res.insert(a.key.clone(), a.value.clone());
```
sebadob: Unnecessary `clone`ing here. When you `attrs.into_iter()` or `.drain(..)`, you can push values directly, saving 2 allocs per iteration.

sebadob (changes requested):

src/api_types/src/oidc.rs
bartekus: Done in fd36c56d. `GrantType` now lives in `api_types::oidc` with the extended grants as simplified variants (`DeviceCode`, `TokenExchange`) plus `serde` renames to the full URNs, an `as_str()`, a `FromStr` and a `csv()` helper for the `flows_enabled` CSV column.

It went a bit further than the one field, because that was the only way to actually drop the regex. It now covers `TokenRequest.grant_type`, `DynamicClientRequest.grant_types`, `UpdateClientRequest.flows_enabled`, and the client/dyn-client responses. Deleted along the way: `RE_GRANT_TYPES`, `RE_GRANT_TYPES_EPHEMERAL` (already dead, only referenced by `init_static_vars`), `validate_vec_grant_types`, `validate_vec_grant_type`, and the `GRANT_TYPE_DEVICE_CODE` / `GRANT_TYPE_TOKEN_EXCHANGE` constants. The enum is the single source of truth now, and `/.well-known` builds `grant_types_supported` from it.

The DB column stays a CSV `String`, so there is no migration: `get_flows()` parses into `Vec<GrantType>` and `GrantType::csv()` writes back.

Two things worth your call:

1. `validate_flow()` now takes a `GrantType` and matches whole CSV entries instead of `flows_enabled.contains(flow)`. The old substring check would have accepted a flow that merely appeared inside another entry. Nothing exploitable with the current six names, but the exact match is free.
2. An unknown `grant_type` is now rejected by `serde` at deserialization rather than by the validator, so the error body changed shape. Still a 400, just a different message. Say the word if you would rather keep a hand-written error there.

The non-empty check that `validate_vec_grant_types` did is preserved as `#[validate(length(min = 1))]`.

src/service/src/oidc/grant_types/token_exchange.rs
bartekus: You are right, and the RFC does not require it in the body. RFC 8693 §2.1 just says the client authenticates per the normal OAuth rules, which means `Authorization: Basic` is equally valid. The pre-check would have broken exactly those clients.

Removed in fd36c56d. `try_get_client_id_secret()` now handles both, and the extracted `client_secret` is the only thing checked, a few lines below.

src/service/src/oidc/grant_types/token_exchange.rs
bartekus: Agreed, and moved as early as you suggested: the check now sits directly after `Client::find()`, before `validate_enabled()`.

That also removed the `ClientDyn::update_used()` call that used to be here, since it is now unreachable for this grant.

src/service/src/oidc/grant_types/token_exchange.rs
bartekus: Done, one buffer now. I pull `sub` and `scope` out of the subject claims inside a small block, so the claims are dropped before the buffer is re-used for the `actor_token` (with a `clear()` first, since `validate_claims_into()` has a `debug_assert!(buf.is_empty())`).

Both values were being allocated into a `String` further down anyway, so this costs nothing extra and saves the second 512 byte `Vec`.

src/service/src/oidc/grant_types/token_exchange.rs
bartekus: Added in fd36c56d: `validate_resource_request()` now rejects dynamic clients up front, right after the empty check.

Since a dyn client can never have `allowed_resources` set, it could not have matched anyway; this just makes it explicit and cheap rather than relying on that invariant holding forever. The token exchange path also rejects dyn clients before it ever gets here, so this is the second of the two layers you asked for.

src/service/src/oidc/grant_types/token_exchange.rs
bartekus: Both done: `Vec::with_capacity(2)`, and `openid` is pushed unconditionally (with a `continue` so an explicitly requested `openid` is not duplicated).

One consequence I want to flag rather than decide silently: forcing `openid` in means the exchanged token can carry it even when the `subject_token` did not, which is the one case where this widens rather than narrows. It is reachable when the subject token came from `client_credentials`. `openid` carries no privilege by itself, so I read your "always, no matter if requested or not" as intended to cover this too, but if you would rather have it conditional on the subject actually holding `openid`, that is a one-line change.

src/service/src/token_set.rs
bartekus: Done, it is a `for a in attrs` consuming loop now, so `key` and `value` move straight into the map.

Note there is an identical block in `from_user()` at what is now line 709. It predates this PR so I left it alone, but it is the same two allocations per iteration if you want it swept in here.

bartekus:
Review round 1 is pushed as `fd36c56d`, with a per-comment reply on each thread.

One note so the description above does not mislead: **design call 1 in the PR body is now superseded.** It argued for storing the grant type URN in `flows_enabled` because it needed "exactly one new regex alternative" in `RE_GRANT_TYPES`. That regex no longer exists. The `GrantType` enum you asked for replaced it, along with `RE_GRANT_TYPES_EPHEMERAL`, both custom validators, and the `GRANT_TYPE_DEVICE_CODE` / `GRANT_TYPE_TOKEN_EXCHANGE` constants.

The conclusion still holds and is now structural rather than a judgement call: the URN is simply what `GrantType::TokenExchange` serialises to, so there is no second spelling to keep in sync. The stored value and the request value are the same enum.

Still no schema change. The column stays a CSV `String`; `get_flows()` parses into `Vec<GrantType>` and `GrantType::csv()` writes back.

Full suite green on both hiqlite and Postgres after the changes.

src/api_types/src/oidc.rs
```diff
@@ -4,18 +4,92 @@ use crate::sessions::SessionState;
...
 use utoipa::{IntoParams, ToSchema};
 use validator::Validate;
 
+/// The OAuth 2.0 / OIDC grant types (flows) Rauthy supports.
+///
+/// The wire format is the grant type identifier itself, which for the extended grants is the
+/// full URN. Keeping this an enum rather than a validated `String` makes it type-safe in code
+/// and self-documenting in the OpenAPI spec.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, ToSchema)]
+pub enum GrantType {
```
sebadob, line 26: The `GrantType` needs to be included in `src/api/src/openapi.rs` so the SwaggerUI can document it and does not throw errors.

src/api_types/src/oidc.rs
sebadob: > validate_flow() now takes a GrantType and matches whole CSV entries instead of flows_enabled.contains(flow). The old substring check would have accepted a flow that merely appeared inside another entry. Nothing exploitable with the current six names, but the exact match is free.

It was safe before because the auth flows were checked before updating them, but the exact match is even more safe, so that's a good one. It's not entirely free because of the `split()`, but even the earlier method needed to iterate over the data, so it's probably almost the same.

> An unknown grant_type is now rejected by serde at deserialization rather than by the validator, so the error body changed shape. Still a 400, just a different message. Say the word if you would rather keep a hand-written error there.

Having `serde` error when it's unknown or totally invalid is fine. A not-allowed flow still returns the custom error message, which is important for UX.

src/api_types/src/oidc.rs
```diff
@@ -272,6 +341,50 @@ pub struct TokenRequest {
...
+    ///
+    /// Validation: `[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$`
+    #[validate(regex(path = "*RE_URI", code = "[a-zA-Z0-9,.:/_-&?=~#!$'()*+%@]+$"))]
+    pub actor_token: Option<String>,
+    /// RFC 8693 token exchange: the type of the `actor_token`. Rauthy only accepts
+    /// `urn:ietf:params:oauth:token-type:access_token`.
+    ///
+    /// Validation: max length is 256
+    #[validate(length(max = 256))]
+    pub actor_token_type: Option<String>,
```
sebadob, line 445: The new `GrantTyper` can now also be used for `actor_token_type` and `requested_token_type`. The RFC requires a specific error on an invalid request, but that should not include not-existing auth flows, since these are specified by the RFC as well. So we can request the `GrantType` driectly and have a type-safe match later on imo, or does it include invalid types specifically?

src/data/src/migration/anti_lockout.rs
```diff
@@ -47,7 +48,7 @@ pub async fn anti_lockout() -> Result<(), ErrorResponse> {
         redirect_uris,
         post_logout_redirect_uris: None,
         allowed_origins,
-        flows_enabled: "authorization_code".to_string(),
+        flows_enabled: GrantType::AuthorizationCode.as_str().to_string(),
```
sebadob: An `impl Display for GrantType` would be nice. It would make such conversions simpler, and also make it more straight forward when logging it. We get `to_string()` for free when we `impl Display`.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,265 @@
...
+                return Err(ErrorResponse::new(
+                    ErrorResponseType::BadRequest,
+                    format!("the requested scope '{s}' is not granted to the 'subject_token'"),
+                ));
+            }
+            narrowed.push(s);
+        }
+        narrowed.join(" ")
+    } else {
+        subject_scope.to_string()
```
sebadob: The allocation was done above already, so you don't need that additional one anymore. Just get rid of the `.as_deref()` in line 166.

src/service/src/oidc/grant_types/token_exchange.rs
```diff
@@ -0,0 +1,265 @@
...
+    token_type: &str,
+    name: &str,
+    buf: &'a mut Vec<u8>,
+) -> Result<JwtAccessClaims<'a>, ErrorResponse> {
+    if token_type != TOKEN_TYPE_ACCESS_TOKEN {
+        return Err(ErrorResponse::new(
+            ErrorResponseType::BadRequest,
+            format!(
+                "the only supported '{name}_type' is \
+                 'urn:ietf:params:oauth:token-type:access_token'"
```
sebadob: No too big of a deal, but we should make use of `TOKEN_TYPE_ACCESS_TOKEN` in this error message.

src/service/src/oidc/grant_types/token_exchange.rs
sebadob: > It is reachable when the subject token came from client_credentials.

Right, totally forgot about that. So yes we should probably leave the decision to the client then. Since we only allow to exchange access tokens anyway for now, there is no identiy part anyway. So let's skip the `openid` force here.

src/service/src/token_set.rs
sebadob: > Note there is an identical block in from_user() at what is now line 709. It predates this PR so I left it alone, but it is the same two allocations per iteration if you want it swept in here.

Did not see that. Would be nice if you could improve that one as well. These are all tiny changes, but they add up at scale.

sebadob (changes requested):

sebadob:
After the merge of #1658, you need to merge `main` and resolve the `GrantType` conflicts, but it should not be too much overhead.

bartekus:
All addressed in `7b411645`, and the branch is rebased onto current `main` (it was still on `73b00c4c`).

**`GrantType` in the OpenAPI spec** — registered in `openapi.rs`.

**`impl Display for GrantType`** — this one was already in the previous commit, so I think it just got lost in the diff. What was missing was actually *using* it: the two remaining `GrantType::AuthorizationCode.as_str().to_string()` call sites are now `.to_string()`.

**`actor_token_type` / `requested_token_type`** — I want to flag a distinction here, because I think the suggestion does not quite hold as written. Those are token type URNs, `urn:ietf:params:oauth:token-type:access_token`, which is a different URN namespace from the grant types (`urn:ietf:params:oauth:grant-type:token-exchange`). `GrantType` cannot represent them, so reusing it would not compile.

What I did instead is add the parallel `TokenType` enum for the RFC 8693 §3 identifiers, which gives the type-safe match you were after:

```rust
if TokenType::from_str(token_type).ok() != Some(TokenType::AccessToken) { ... }
```

To your actual question, it does include invalid types specifically, and that is why `TokenType` is deliberately *not* the wire type. `subject_token_type` and friends stay `Option<String>` and get parsed in the handler. If they were the enum, `serde` would reject an unsupported value during deserialization and the client would get a generic deserialization error instead of our OAuth error object. That is fine for `grant_type`, where you said the `serde` error is acceptable and a not-allowed flow still hits the custom message, but for these three the RFC error *is* the response, so it has to stay ours. All six RFC-defined types are listed so that "a token type we do not issue" and "not a token type at all" can be told apart.

`TOKEN_TYPE_ACCESS_TOKEN` is gone as a result, since `TokenType::AccessToken` now covers it, mirroring the `GRANT_TYPE_*` constants. It is used in the error messages via `Display`.

**`.as_deref()` on line 166, the `openid` force, and the `from_user()` block** — all done. `from_user()` now consumes the attrs by value like `for_token_exchange()` does.

**Two things the rebase turned up, both worth a look:**

1. #1658 added `retain_supported_grant_types()` built on `RE_GRANT_TYPES`, which this branch removes. It now filters on `GrantType::from_str()`. Your test for it passes unchanged, and the enum is the single source of truth for what "supported" means.

2. `EphemeralClientRequest::grant_types` had to stay a `Vec<String>` rather than becoming a `Vec<GrantType>` like every other grant-type field. The CIMD document is fetched from a remote source and `ignore_unknown_auth_flows` strips unsupported grants *after* deserialization, so making it the enum would have `serde` reject claude.ai's document before the strip could run: it would have silently disabled the opt-in you just merged. There is a comment on the field explaining why that one is the exception. `validate_vec_grant_type` is kept for it and now also filters through the enum.

The DCR strictness test posts raw JSON now, because the invalid payload it asserts on can no longer be built as a Rust value. That also pins the behaviour at the layer that enforces it.

Verified: `cargo clippy --all-features -- -D warnings` clean, `cargo fmt --check` clean, unit tests green, and the full integration suite green on **both** hiqlite and Postgres, including `test_dcr_rejects_unsupported_grant` and `test_resource_survives_authorize_refresh` from the PRs that landed today.

sebadob:
> actor_token_type / requested_token_type — I want to flag a distinction here, because I think the suggestion does not quite hold as written. Those are token type URNs, urn:ietf:params:oauth:token-type:access_token, which is a different URN namespace from the grant types (urn:ietf:params:oauth:grant-type:token-exchange). GrantType cannot represent them, so reusing it would not compile.

Yes of course, my bad, sorry.

> EphemeralClientRequest::grant_types had to stay a Vec<String> rather than becoming a Vec<GrantType> like every other grant-type field. The CIMD document is fetched from a remote source and ignore_unknown_auth_flows strips unsupported grants after deserialization, so making it the enum would have serde reject claude.ai's document before the strip could run

That makes perfect sense.

---

After the latest merges to `main`, the Changelog now has a conflict.

src/common/src/constants.rs
```diff
@@ -76,7 +76,7 @@ pub static DEVICE_KEY_LENGTH: u8 = 64;
 /// Max serialized length of a client's custom `claims` JSON object.
 pub const CLIENT_CLAIMS_MAX_LEN: usize = 1024;
 pub static EVENTS_LATEST_LIMIT: u16 = 100;
-pub static GRANT_TYPE_DEVICE_CODE: &str = "urn:ietf:params:oauth:grant-type:device_code";
+/// RFC 8693 token type identifier for an OAuth 2.0 access token.
```
sebadob: That comment is obsolete now and wrongly comments the `UPSTREAM_AUTH_CALLBACK_TIMEOUT_SECS`.

src/jwt/src/claims.rs
```diff
@@ -141,6 +145,16 @@ pub struct JwtAccessClaims<'a> {
     pub custom_flattened: Option<HashMap<String, serde_json::Value>>,
 }
 
+/// RFC 8693 §4.1 `act` (actor) claim. Identifies the party acting on behalf of the `sub`.
+/// Nests recursively, so a chain of delegation keeps the whole history: the outermost `act`
+/// is the current actor, and each nested `act` is the one it in turn acts for.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ActClaim {
+    pub sub: String,
```
sebadob: Have not tested it, but we should be able to use borrowed data here, right?

Since there is one other change necessary before we can merge, you could have a look. I am not sure because of the self-referencing boxed value, but please check if this is possible:

```rust
pub struct ActClaim<'a> {
    pub sub: &'a str,
    #[serde(borrow, skip_serializing_if = "Option::is_none")]
    pub act: Option<Box<ActClaim<'a>>>,
}
```

If this works, we can save some allocations here.

sebadob (changes requested):

bartekus:
Both addressed in `3ff23bf1`, and the branch is rebased onto current `main` so the CHANGELOG conflict is gone.

**The obsolete comment** was mine: it was the doc comment on `TOKEN_TYPE_ACCESS_TOKEN`, which last round's `TokenType::AccessToken` replaced. Deleting the constant left the comment sitting on the next line down. Removed.

**Borrowing `ActClaim`** works, exactly as you wrote it. No lifetime fight at all, which surprised me a little, so here is why it holds:

`grant_token_exchange` already owns the `buf` the claims are deserialized out of, it declares it before the `act` is built and never touches it again afterwards, and `TokenSet::for_token_exchange` is awaited before that scope ends. So the borrow the actor claims hand out lives long enough on its own; the only change needed beyond the struct was `act: Option<ActClaim<'_>>` on the two `token_set` signatures and dropping the `.to_string()`.

The nesting is the part you flagged, and `#[serde(borrow)]` does carry through the `Box`. Since that is not obvious from reading it, I pinned it with a test rather than leaving it to the integration suite, which only ever asserts one level deep:

```rust
let range = bytes.as_ptr_range();
for sub in [outer.sub, middle.sub, inner.sub] {
    assert!(range.contains(&sub.as_ptr()), "`{sub}` was copied instead of borrowed");
}
```

A three-link chain round trips and every `sub` in it points into the input buffer, so the whole chain is allocation-free, not just the outermost link. The test also pins that an absent `act` stays omitted rather than becoming `null`, which matters now that a re-issued token can carry a chain.

One caveat worth naming, because it is inherent to borrowed deserialization rather than to this change: `from_slice` into a `&str` fails if that JSON string carries an escape sequence. It cannot here, since an `act.sub` is either a user ID or a client ID, and `common.sub` in the same struct is already `Option<&'a str>` and has been all along. So this adds no new constraint; it inherits the one that is already there.

Verified: `cargo clippy --all-features -- -D warnings` clean, `cargo fmt --check` clean, unit tests green, and the full integration suite green on both hiqlite and Postgres with the borrowed chain in place, including `test_token_exchange`.

sebadob (approved):
Thanks

PR #1653, opened 2026-07-18
feat(clients): add per-client favicon branding

### LINKED ISSUE #1648: Allow custom favicons for branded clients
Client-branded login pages still use the hard-coded Rauthy favicon.

Allow a per-client favicon through the branding API/UI, or reuse the client logo.

A reverse proxy or custom Rauthy build, both of which add unnecessary operational complexity.

mrchypark:
Fixes #1648.

## Summary

- add independent per-client favicon upload, preview, and delete controls
- reuse `client_logos` with `LogoRes::Favicon` instead of adding a separate table
- keep client logos and favicons independent for lookup, caching, replacement, and deletion
- expose the public favicon endpoint while requiring `Clients.Update` for mutations
- update the OIDC authorize page favicon in production and DEV UI
- align logo and favicon deletion controls in the client branding UI

## Validation

- `cargo fmt --check`
- frontend formatting and Svelte checks
- frontend production build
- targeted client favicon integration test compilation
- regression coverage for SVG favicon uploads not replacing the cached client logo


sebadob:
> This is a Draft while the maintainer direction on https://github.com/sebadob/rauthy/issues/1648 is pending

What direction do you need? Allowing custom favicons makes sense. 

Even though still marked as draft, I took a very quick look into the changes. We don't need a new table for this. Especially because it's basically a copy of the already existing `client_logos` which we should reuse. The only change necessary would be an addon to `LogoRes::Favicon`.

mrchypark:
Thanks — I agree. I’ve updated the implementation to reuse  with  instead of introducing a separate table, and aligned migration/setup changes accordingly.

src/api/src/clients.rs
```diff
@@ -336,35 +336,64 @@ pub async fn get_client_logo(
...
-            .insert_header((CONTENT_TYPE, content_type))
+        HttpResponse::Ok()
+            .insert_header((CONTENT_TYPE, logo.content_type))
             .insert_header((CONTENT_SECURITY_POLICY, csp))
-            .body(data))
+            .body(logo.data)
+    }
+}
+
+async fn multipart_image(
```
sebadob: That function name is not very descriptive. A better name would be something like `extract_multipart_data`, `get_multipart_data`, or something similar. Always add a hint to functions what it actually does.

src/api/src/clients.rs
```diff
@@ -387,42 +416,102 @@ pub async fn put_client_logo(
...
-                content_type = Some(mime.clone());
-            }
-            None => {
-                return Err(ErrorResponse::new(
-                    ErrorResponseType::BadRequest,
-                    "content_type is missing",
-                ));
-            }
-        }
+    let (buf, content_type) = multipart_image(payload).await?;
```
sebadob: This is very picky, but when extracting such data from a function, `data` or `bytes` makes more sense as a name than `buf`. `buf` is a buffer that only makes sense inside that function.

sebadob (changes requested):
CSS is broken on the branding page. It also makes no sense to have a delete button for the Favicon but not for the Logo. Both should work in the same way, so please add a delete button for the client logo as well.

Image

---

When running the DEV UI, it does not work as expected. I used the `rauthy` client on localhost to test the favicon. The normal Client logo works fine (as it did before), but the favicon stays the same. It shows the correct one when looking at the client config, but the Login page still shows the default.

That's not a deal breaker, since it affects DEV only, but it would be nice to see it working during there as well.

Apart from that, it seems to work nicely after a first test.

src/api/src/clients.rs
mrchypark: Agreed. I renamed it to `extract_multipart_data` in 06d6342a.

src/api/src/clients.rs
mrchypark: Agreed. I renamed the extracted payload to `data` in 06d6342a.

mrchypark:
Thanks, I agree with the review. I applied the requested layout, logo deletion, naming, and DEV favicon changes in 06d6342a.

sebadob (approved):
Thanks

sebadob:
I just had a look at `main` with all the latest changes, and your `test_client_favicon_is_independent_from_logo` test is broken.

Even though it requires a lot more CI time, I will add tests to future pipelines to catch things like this.

Edit:

It's broken in line 430, probably because of missing content type or something like that.

mrchypark:
Thanks for catching this. I agree—the helper was declaring the SVG bytes as image/png. I updated it to use image/svg+xml and verified the favicon flow in #1669.

PR #1655, opened 2026-07-18
Fix: omit `sub` claim when absent instead of emitting `null`

### LINKED ISSUE #1646: client_credentials tokens emit `"sub": null` — RFC 7519 requires StringOrURI; PyJWT-based resource servers (all FastMCP MCP servers) reject them
### Area

oidc / client_credentials

### Describe the bug

Access tokens issued via `client_credentials` carry **`"sub": null`** rather than omitting the claim. [RFC 7519 §4.1.2](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.2) makes `sub` OPTIONAL, but says that when present its value MUST be a `StringOrURI`. `null` is neither, so a spec-compliant validator rejects the whole token.

PyJWT ≥ 2.10 enforces this, which means **no PyJWT-based resource server can validate a Rauthy `client_credentials` token** — including every FastMCP-based MCP server (FastMCP is the most common Python MCP framework, and MCP servers are exactly the audience for `client_credentials`).

Actual token from `grant_type=client_credentials`:

```json
{
  "iat": 1784237454, "nbf": 1784237454, "exp": 1784239254,
  "iss": "https://rauthy.example.com/auth/v1/",
  "jti": "SKPUoOwidvwK",
  "aud": ["rs-test", "https://mcp.example.com/mcp"],
  "sub": null,
  "typ": "Bearer",
  "azp": "rs-test",
  "scope": "openid"
}
```

### Steps to reproduce

```bash
curl -sS -X POST https://<rauthy>/auth/v1/oidc/token \
  -d grant_type=client_credentials -d client_id=<id> -d client_secret=<secret>
```

Then validate the returned `access_token`:

```python
import jwt   # PyJWT >= 2.10
jwt.decode(token, key, algorithms=["RS256"], audience=..., issuer=...)
# jwt.exceptions.InvalidSubjectError: Subject must be a string
```

This fails even with `options={"verify_aud": False, "verify_iss": False}` — the subject check is unconditional, so signature-only verification fails too. An `authorization_code` token from the same client validates fine (it has a real `sub`).

### Expected behavior

Omit `sub` entirely when there is no subject (`client_credentials` has no end user), rather than serializing it as `null`. Alternatively set it to the `client_id`, which is what several ASes do — but omission is the safer, strictly-correct reading of RFC 7519.

Concretely: `#[serde(skip_serializing_if = "Option::is_none")]` on the `sub` field of the access-token claims struct.

### Actual behavior

`"sub": null` is emitted, and RFC-compliant validators reject the token outright.

### Version

0.36.0 (`ghcr.io/sebadob/rauthy:0.36.0`), Hiqlite backend.

### Notes

Found while testing Rauthy as the AS for MCP servers (cf. #1562, #1643, #1644, #1645). Worth noting this is the same failure shape as [ory/hydra's `"contacts": null`](https://github.com/ory/hydra) on its DCR endpoint — a present-but-null field where the spec requires a typed value; strict clients reject the response and operators end up rewriting JSON at the proxy. Happy to send a PR if you'd like.

impuls42:
`client_credentials` tokens have no subject, but `sub` on `JwtCommonClaims` was the one `Option` claim still missing `#[serde(skip_serializing_if = "Option::is_none")]`, so it serialized as `"sub": null`.

As discussed in #1646, `null` isn't strictly invalid, but omitting it is cleaner and matches every other optional claim there. `id_token` and user access tokens always set `sub`, so the only tokens that change are the subject-less ones.

Regression test in `handler_1646_sub_null.rs`: fetch a `client_credentials` token, assert the decoded payload has no `sub` key. It fails before the change (key present as `null`) and passes after. There's also a unit test on the serialization. Full suite green.

Closes #1646.

### DIFF
--- a/src/jwt/src/claims.rs
+++ b/src/jwt/src/claims.rs
@@ -20,6 +20,7 @@ pub struct JwtCommonClaims<'a> {
     pub jti: Option<&'a str>,
     #[serde(borrow)]
     pub aud: Audience<'a>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub sub: Option<&'a str>,
     // pub nonce: Option<&'a str>,
     pub typ: JwtTokenType,
@@ -559,6 +560,26 @@ mod tests {
                 "id custom_flattened leaked reserved claim `{reserved}`"
             );
         }
+    }
+
+    // Issue #1646: a subject-less token (e.g. `client_credentials`) must OMIT `sub`
+    // rather than emit `"sub": null`. RFC 7519 §4.1.2 requires `sub`, when present, to
+    // be a StringOrURI; `null` violates that and PyJWT >=2.10 rejects such tokens.
+    #[test]
+    fn common_claims_omit_sub_when_none_present_when_some() {
+        // sub = None -> the `sub` key must not appear at all
+        let mut claims = common();
+        claims.sub = None;
+        let v = serde_json::to_value(&claims).unwrap();
+        assert!(
+            v.get("sub").is_none(),
+            "`sub` must be omitted (not null) when None, got: {v}"
+        );
+
+        // sub = Some(..) -> the `sub` key must be present as that string
+        claims.sub = Some("some-subject-id");
+        let v = serde_json::to_value(&claims).unwrap();
+        assert_eq!(v.get("sub"), Some(&json!("some-subject-id")));
     }
 
     // (a) No custom value can produce shadowed JSON: a flattened map carrying a reserved


src/bin/tests/handler_1646_sub_null.rs
```diff
@@ -0,0 +1,72 @@
...
+// A `client_credentials` token has no end-user subject. Before the fix the `sub` claim was
+// serialized as an explicit JSON `null` (`"sub": null`), which trips strict OIDC/JWT consumers
+// that require `sub` to be a string when present. The fix marks the claim
+// `#[serde(skip_serializing_if = "Option::is_none")]`, so the key is now ABSENT entirely.
+//
+// Without the fix this test fails: `claims.get("sub")` would return `Some(Value::Null)` instead
+// of `None`, so `assert!(claims.get("sub").is_none())` panics.
+// -----------------------------------------------------------------------------------------
+#[tokio::test]
+async fn test_1646_client_credentials_sub_absent() -> Result<(), Box<dyn Error>> {
```
sebadob: We really don't need this integration test. Even the unit test below is something I would not have added, but that's ok to keep. The only reason it was serialized as `null` is because the omit was forgotten when it was made optional in the past. There is no reason to ever remove this in the future, which would be the only way to break something.

My issue with this one here is that when we start adding tests like `handler_someRFC_whatever*`, integration tests become super messy, and also no one can remember RFC numbers.  
Apart from that, each additional file here also increases the test compilation time a lot. Which is actually the biggest deal, and which is why the other files have so many tests with the goal to keep the number of files low. The integration test setup actually needs a rework to only have a single file that calls into different modules. It's a huge amount of work necessary already to compile all of them.

sebadob (changes requested):

impuls42:
Done. Dropped the integration test and kept only the unit test. Fair point on the `handler_<RFC>_*` sprawl and the compile-time hit from extra files.

sebadob (approved):

sebadob:
Thanks

PR #1656, opened 2026-07-18
Fix: serve RFC 8414 metadata at the path-insertion URL for the `/auth/v1` issuer

### LINKED ISSUE #1643: RFC 8414: metadata URL for a path-ful issuer (/.well-known/oauth-authorization-server/auth/v1) 301s to the SPA — claude.ai can't discover CIMD
### Area

oidc / discovery

### Describe the bug

Rauthy's issuer contains a path (`https://<host>/auth/v1/`). Per [RFC 8414 §3.1](https://www.rfc-editor.org/rfc/rfc8414#section-3.1), when the issuer has a path component, the metadata URL is formed by inserting `/.well-known/oauth-authorization-server` **between the host and the path**:

```
issuer:   https://rauthy.example.com/auth/v1/
metadata: https://rauthy.example.com/.well-known/oauth-authorization-server/auth/v1
```

PR #1561 (thank you — it's what made me try this) added the alias at the **root** (`/.well-known/oauth-authorization-server`), which is the correct URL only for a path-less issuer. The RFC-compliant URL for Rauthy's own issuer currently **301-redirects to the SPA** and returns `text/html`:

```
$ curl -sS -i https://rauthy-spike.example/.well-known/oauth-authorization-server/auth/v1
HTTP/2 301
location: /auth/v1/
# -> follows to text/html;charset=utf-8  (the login SPA, not metadata)
```

#1561's rationale notes that ChatGPT's connector backend reads CIMD support from the RFC 8414 endpoint. ChatGPT appears to probe the **root** form, so the root alias fixed it there. **claude.ai probes the RFC-correct path-insertion form**, gets HTML, concludes the AS has no CIMD support, and falls back to Dynamic Client Registration.

For comparison, Keycloak 26.7 (issuer `https://host/realms/master`) serves the path-insertion form directly:

```
$ curl -sS -o /dev/null -w '%{http_code}\n' https://kc/.well-known/oauth-authorization-server/realms/master
200
```

### Expected behavior

`GET /.well-known/oauth-authorization-server/auth/v1` returns the same JSON metadata body as `/auth/v1/.well-known/openid-configuration` (200, `application/json`), rather than redirecting to the UI.

### Actual behavior

301 → `/auth/v1/` → `text/html` login SPA. Clients that implement RFC 8414 path-insertion cannot discover `client_id_metadata_document_supported`.

### Steps to reproduce

```
curl -sS -o /dev/null -w '%{http_code}\n' https://<rauthy>/.well-known/oauth-authorization-server/auth/v1   # 301 -> HTML
curl -sS -o /dev/null -w '%{http_code}\n' https://<rauthy>/.well-known/oauth-authorization-server            # 200
curl -sS -o /dev/null -w '%{http_code}\n' https://<rauthy>/auth/v1/.well-known/oauth-authorization-server    # 200
```

Verified against a claude.ai custom connector: with only the root + path-appended aliases present, Claude never attempts CIMD and goes straight to `POST /auth/v1/clients_dyn`. Confirmed by the absence of any `"Rauthy Client v0.36.0"` fetch of the client-metadata document in the document host's access log.

### Version

0.36.0 (`ghcr.io/sebadob/rauthy:0.36.0`), Hiqlite backend, single replica.

### Notes

Serving all three forms seems harmless — the root alias keeps ChatGPT working, and the path-insertion form is what RFC 8414 actually specifies for this issuer. Related: #1562, #1561.

impuls42:
Rauthy's issuer carries a path (`https://<host>/auth/v1/`). Per RFC 8414 §3.1 the metadata URL for a path-ful issuer inserts the well-known segment between host and path: `/.well-known/oauth-authorization-server/auth/v1`. Rauthy only exposed the root alias `/.well-known/oauth-authorization-server` (added for ChatGPT, which probes the root form), so the path-insertion URL fell through to the SPA catch-all and 301'd to HTML. A client that follows RFC 8414 to the letter (claude.ai does) can't discover `client_id_metadata_document_supported` and falls back to Dynamic Client Registration.

This serves the same metadata body at the path-insertion URL, in both the no-slash and trailing-slash forms (the issuer path ends in `/`, and there's no `NormalizePath` middleware). It's registered at the two app-root builder sites, not the `/auth/v1`-scoped one, where it would resolve to a nonsensical `/auth/v1/.well-known/oauth-authorization-server/auth/v1`. `/auth/v1` is a constant base path, so the routes are fixed.

You mentioned in the issue that there's already a root-level mechanism for this. If you'd rather I reuse that than add the two explicit handlers, point me at it and I'll switch.

Regression test hits both slash forms and asserts 200, `application/json`, and a matching issuer. Full suite green.

Closes #1643.

### DIFF
--- a/src/api/src/oidc.rs
+++ b/src/api/src/oidc.rs
@@ -1268,3 +1268,29 @@ pub async fn get_well_known() -> Result<HttpResponse, ErrorResponse> {
 pub async fn get_well_known_oauth() -> Result<HttpResponse, ErrorResponse> {
     well_known_response().await
 }
+
+// RFC 8414 §3.1 path-insertion alias for the AS-metadata document. Because
+// rauthy's issuer carries a path component (`https://<host>/auth/v1/`), an
+// RFC-compliant client forms the metadata URL by INSERTING
+// `/.well-known/oauth-authorization-server` between host and issuer path, i.e.
+// `https://<host>/.well-known/oauth-authorization-server/auth/v1`. claude.ai
+// probes exactly this path-insertion form; without it the request falls through
+// to the SPA catch-all (301 -> HTML), so the client never reads
+// `client_id_metadata_document_supported` and falls back to Dynamic Client
+// Registration. The `/auth/v1` segment is rauthy's fixed API mount prefix (see
+// the hardcoded issuer in `rauthy_config.rs`), so the route can be a constant.
+// Body is identical to `get_well_known_oauth`. See issue #1643.
+//
+// Both the trailing-slash and no-slash forms are served: the issuer path is
+// `/auth/v1/` (with a trailing slash), so a strict client may preserve it, while
+// claude.ai probes the no-slash form. rauthy runs no `NormalizePath` middleware,
+// so each exact path needs its own route.
+#[get("/.well-known/oauth-authorization-server/auth/v1")]
+pub async fn get_well_known_oauth_rfc8414() -> Result<HttpResponse, ErrorResponse> {
+    well_known_response().await
+}
+
+#[get("/.well-known/oauth-authorization-server/auth/v1/")]
+pub async fn get_well_known_oauth_rfc8414_trailing() -> Result<HttpResponse, ErrorResponse> {
+    well_known_response().await
+}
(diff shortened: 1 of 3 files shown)


sebadob:
Thanks.

Issue #1654, opened 2026-07-18
`auth_time` derived from mutable `user.last_login` can report a time later than the user's actual last authentication


impuls42:
### Area

oidc

### Describe the bug

The `auth_time` claim is derived from the mutable `user.last_login` field, which is updated by events that are **not** a fresh user authentication (notably the `refresh_token` grant). As a result the `authorization_code` grant can emit an `auth_time` that is *later* than the moment the user actually authenticated for that session.

`authorization_code` grant — `src/service/src/oidc/grant_types/authorization_code.rs`:

```rust
AuthTime::given(user.last_login.unwrap_or_else(|| Utc::now().timestamp())),
```

`user.last_login` is written in many places, including on a token refresh — `src/service/src/oidc/validation.rs`:

```rust
// set last login
user.last_login = Some(Utc::now().timestamp());
```

(other writers: `authorize.rs` login, `password.rs`, `webauthn.rs`, `pam.rs`, `auth_providers.rs`.)

So `user.last_login` tracks "most recent activity of any kind for this user", not "when the user last actually authenticated". Meanwhile the `refresh_token` grant itself handles `auth_time` correctly — it carries `claims.auth_time` forward rather than reading `user.last_login` — which makes the `authorization_code` path the inconsistent one.

### Scenario

1. User logs in at `T0`. `last_login = T0`, token `auth_time = T0`. Correct.
2. A `refresh_token` grant happens at `T1` (routine token rotation, possibly on another device/session). `validation.rs` sets `last_login = T1`. The refreshed token's own `auth_time` is still correctly `T0` (carried from claims), but the stored `last_login` is now `T1`.
3. The UI triggers a silent re-auth via `POST /oidc/authorize/refresh` at `T2` (no actual re-authentication — the session is reused). A new auth code is issued and exchanged; the `authorization_code` grant reads `user.last_login` (`= T1`, or later) and emits `auth_time = T1`.

The user has not authenticated since `T0`, but the issued token claims `auth_time = T1`.

### Impact

`auth_time` is used by relying parties for `max_age` / step-up-authentication decisions ("require the user to have authenticated within the last N seconds"). Because `last_login` can be advanced by a background token refresh, a stale session can present an `auth_time` fresher than the real authentication, potentially satisfying a `max_age` requirement without the user actually re-authenticating. That is the direction that matters for security (too-recent, not too-old).

### Expected behavior

[...]

PR #1657, opened 2026-07-18
Fix: carry RFC 8707 `resource` through the session-refresh authorize path

### LINKED ISSUE #1645: RFC 8707: `resource` is dropped on /oidc/authorize/refresh (LoginRefreshRequest has no resource field) — audience binding breaks for returning users
### Area

oidc / resource indicators (RFC 8707)

### Describe the bug

`resource` is accepted on the full-login authorization path but **silently dropped on the silent re-auth path**, so RFC 8707 works on a user's *first* login and fails on every subsequent one.

The two request types are not symmetric — `LoginRefreshRequest` has no `resource` field:

```rust
// src/api_types/src/oidc.rs
pub struct LoginRequest {          // POST /oidc/authorize        (full login)
    pub email, password, pow, client_id, redirect_uri, scopes,
    pub state, nonce, code_challenge, code_challenge_method,
    pub resource: Option<String>,          // <-- present
}

pub struct LoginRefreshRequest {   // POST /oidc/authorize/refresh (existing session)
    pub client_id, redirect_uri, scopes,
    pub state, nonce, code_challenge, code_challenge_method,
                                           // <-- `resource` absent
}
```

Once a user has a valid Rauthy session, the UI posts to `/oidc/authorize/refresh` instead of `/oidc/authorize`. The `resource` from the original authorization request never reaches the auth code, so the subsequent token request is rejected:

```
WARN post_token{grant_type="authorization_code"}:grant_type_authorization_code{client_id="https://…/claude-connector.json"}:
     rauthy_service::oidc::grant_types::authorization_code: token request `resource` was not granted at the authorization request
ERROR post_token{grant_type="authorization_code"}: rauthy_handlers::oidc: the requested `resource` was not granted at the authorization request
```

### Steps to reproduce

Ephemeral (CIMD) client whose document declares `allowed_resources: ["https://mcp.example.com/mcp"]`, driven by a real claude.ai custom connector:

1. **Fresh browser (no Rauthy session)** → authorization request carries `resource=https://mcp.example.com/mcp` → login form → `POST /oidc/authorize` → token request **succeeds**, `aud` = `["<client_id>", "https://mcp.example.com/mcp"]`. ✅
2. **Repeat with the session still alive** → `POST /oidc/authorize/refresh` (no login form) → token request **fails** with `the requested resource was not granted at the authorization request`. ❌

Same client, same document, same `resource` — the only difference is which authorize endpoint the UI used.

### Expected behavior

`/oidc/authorize/refresh` accepts `resource` and validates/grants it identically to `/oidc/authorize`, so audience-bound tokens keep working for returning users.

### Actual behavior

`resource` is dropped on the silent re-auth path; every returning user loses audience binding and `POST /oidc/token` rejects the request. Since MCP clients send `resource` on every connection, in practice RFC 8707 only works for a user's very first authorization and breaks thereafter.

### Version

0.36.0 (`ghcr.io/sebadob/rauthy:0.36.0`), Hiqlite, single replica, `[ephemeral_clients] enable = true`.

### Notes

Found while testing Rauthy as the AS for a claude.ai MCP custom connector (cf. #1562). The fix looks like adding `resource: Option<String>` to `LoginRefreshRequest` and threading it into the same `AuthorizeData` path that `/oidc/authorize` already uses — happy to send a PR if you'd like. Related: #1643, #1644.

impuls42:
`resource` is accepted on `/oidc/authorize` but dropped on `/oidc/authorize/refresh`: `LoginRefreshRequest` has no `resource` field and `post_authorize_refresh` hard-codes `resource: None`, so it never reaches the new auth code. Audience binding works on a user's first login and silently breaks on every later one once a session exists (the `/token` call then fails with "the requested resource was not granted at the authorization request").

The fix adds the field and threads it through the same `AuthorizeData` -> `finish_authorize` path the full-login flow already uses, so it's validated the same way via `client.validate_resource_request`. The `refresh_token` grant is untouched.

It also tightens the `resource` validators to reject a `#` fragment. RFC 8707 §2 wants an absolute URI without one, the shared `RE_URI` allowed it, and the field docs already said "without a fragment". I flagged this in the issue thread.

One caveat: the UI still needs to forward `resource` on the `/authorize/refresh` POST for the full end-to-end path. The backend now accepts and carries it.

Regression test in `handler_1645_resource_refresh.rs`: log in with a `resource` and check `aud`, then do a silent re-auth via `/authorize/refresh` with the same `resource` and check the re-issued token still has it in `aud`. The last assertion fails without the fix. Validator unit tests cover a valid resource, a malformed one, and one with a fragment.

Closes #1645. (#1654 is the same authorize/refresh path for `auth_time`, handled separately in #1659.)

### DIFF
--- a/src/common/src/regex.rs
+++ b/src/common/src/regex.rs
@@ -71,6 +71,10 @@ pub static RE_STREET: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9À-ÿ-.\s]{0,48}$").unwrap());
 pub static RE_URI: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9,.:/_\-&?=~#!$'()*+%@]+$").unwrap());
+// Like `RE_URI` but WITHOUT `#`, so a value cannot contain a fragment. Used for RFC 8707
+// `resource` indicators, which MUST be an absolute URI without a fragment (RFC 8707 §2).
+pub static RE_RESOURCE: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9,.:/_\-&?=~!$'()*+%@]+$").unwrap());
 pub static RE_USER_NAME: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^[a-zA-Z0-9À-ɏ-'\s\x{3041}-\x{3096}\x{30A0}-\x{30FF}\x{3400}-\x{4DB5}\x{4E00}-\x{9FCB}\x{F900}-\x{FA6A}\x{2E80}-\x{2FD5}\x{FF66}-\x{FF9F}\x{FFA1}-\x{FFDC}\x{31F0}-\x{31FF}]{1,32}$").unwrap()
 });
--- a/frontend/src/api/types/authorize.ts
+++ b/frontend/src/api/types/authorize.ts
@@ -40,6 +40,9 @@ export interface LoginRefreshRequest {
     /// Validation: PATTERN_CODE_CHALLENGE
     code_challenge?: string;
     code_challenge_method?: CodeChallengeMethod;
+    /// RFC 8707 resource indicator forwarded from the authorization request.
+    /// Validation: PATTERN_RESOURCE
+    resource?: string;
 }
 
 export interface RequestResetRequest {
--- a/frontend/src/lib/admin/clients/ClientConfig.svelte
+++ b/frontend/src/lib/admin/clients/ClientConfig.svelte
@@ -12,6 +12,7 @@
         PATTERN_CONTACT,
         PATTERN_GROUP,
         PATTERN_ORIGIN,
+        PATTERN_RESOURCE,
         PATTERN_URI,
     } from '$utils/patterns';
     import {
@@ -405,14 +406,14 @@
             bind:values={allowedResources}
             label={ta.clients.allowedResources}
             errMsg={ta.validation.uri}
-            pattern={PATTERN_URI}
+            pattern={PATTERN_RESOURCE}
         />
         <p class="desc">{ta.clients.descDefaultAud}</p>
         <InputTags
             bind:values={defaultAud}
             label={ta.clients.defaultAud}
             errMsg={ta.validation.uri}
-            pattern={PATTERN_URI}
+            pattern={PATTERN_RESOURCE}
         />
 
         <div style:height=".5rem"></div>
--- a/frontend/src/routes/oidc/authorize/+page.svelte
+++ b/frontend/src/routes/oidc/authorize/+page.svelte
@@ -228,6 +228,9 @@
             payload.code_challenge = challenge;
             payload.code_challenge_method = challengeMethod;
         }
+        if (resource) {
+            payload.resource = resource;
+        }
 
         let res = await fetchPost<undefined | WebauthnLoginResponse>(
             '/auth/v1/oidc/authorize/refresh',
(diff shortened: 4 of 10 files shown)


src/common/src/regex.rs
```diff
@@ -71,6 +71,10 @@ pub static RE_STREET: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9À-ÿ-.\s]{0,48}$").unwrap());
 pub static RE_URI: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"^[a-zA-Z0-9,.:/_\-&?=~#!$'()*+%@]+$").unwrap());
+// Like `RE_URI` but WITHOUT `#`, so a value cannot contain a fragment. Used for RFC 8707
+// `resource` indicators, which MUST be an absolute URI without a fragment (RFC 8707 §2).
+pub static RE_URI_RESOURCE: LazyLock<Regex> =
```
sebadob: It makes no sense to have `URI` in the regex here. Name it only `RE_RESOURCE` instead. The only reason the `RE_URI` was used before is because it existed already.

Can you double check that the claim is not even more restrictive than a URI? Most special characters do not really make much sense here from a logical standpoint. If all of these are actually allowed, we can keep them though.

The bigger part about it is that the validation regexes for `Client` config need an update as well. In the `src/api_types/src/clients.rs` as well as the matching pattern in the UI, so we are consistent.

src/bin/tests/handler_1645_resource_refresh.rs
```diff
@@ -0,0 +1,225 @@
...
+// This test drives the full flow end-to-end:
+//   (a) create a client with `allowed_resources` = [RESOURCE]
+//   (b) full authorization_code login WITH `resource`, exchange -> assert `aud` contains RESOURCE
+//   (c) hit `POST /oidc/authorize/refresh` on the SAME authenticated session WITH the same
+//       `resource`, exchange the new code -> assert `aud` STILL contains RESOURCE
+//
+// Without the fix, step (c)'s token drops the resource and the final `aud` assertion fails.
+// -----------------------------------------------------------------------------------------
+#[tokio::test]
+async fn test_1645_resource_survives_authorize_refresh() -> Result<(), Box<dyn Error>> {
```
sebadob: Please add any additional tests that are necessary to `src/bin/tests/zzf_handler_resource_indicators.rs` instead of creating another file. Each additional file increases the work necessary to compile integration tests.

We also have quite a lot of redundancy here. All of that needs maintenance. Merge tests with the existing ones for `resource`.

sebadob (changes requested):

impuls42:
Reworked:

- Renamed `RE_URI_RESOURCE` -> `RE_RESOURCE`. On whether it should be stricter: it's `RE_URI` with `#` removed and nothing else. Every remaining character is one `RE_URI` already accepts (redirect_uris etc.), and a resource indicator is an absolute URI, so they're all legitimately valid. Going stricter (e.g. dropping the query chars `&?=`) would reject valid resources, since RFC 3986 permits a query in an absolute-URI. The only thing RFC 8707 §2 forbids is the fragment, which is exactly what we drop. Glad to tighten further if you'd prefer, but I left it as the URI set minus `#`.
- Updated the `Client` config side too: `allowed_resources` / `default_aud` now validate with a dedicated `validate_vec_resource` (RE_RESOURCE) instead of the fragment-allowing `validate_vec_uri`, with a matching `PATTERN_RESOURCE` in the admin UI.
- Merged the test into `zzf_handler_resource_indicators.rs` instead of a new file.
- Folded the SPA `resource` forwarding from #1662 into this PR and closed that one, so all the resource work is here now.

sebadob (dismissed):
Thanks.

sebadob:
> Glad to tighten further if you'd prefer, but I left it as the URI set minus #

No need to tighten then, When a full URI is allowed, we should accept that as well.

PR #1658, opened 2026-07-18
Fix: opt-in to accept CIMD documents that advertise unsupported grant types

### LINKED ISSUE #1644: claude.ai's CIMD document is rejected: grant_types validation has no urn:ietf:params:oauth:grant-type:jwt-bearer (DCR + ephemeral paths)
### Area

oidc / dynamic clients / ephemeral clients

### Describe the bug

claude.ai publishes a Client ID Metadata Document at `https://claude.ai/oauth/mcp-oauth-client-metadata`:

```json
{
  "client_id": "https://claude.ai/oauth/mcp-oauth-client-metadata",
  "client_name": "Claude",
  "client_uri": "https://claude.ai",
  "redirect_uris": ["https://claude.ai/api/mcp/auth_callback"],
  "grant_types": ["authorization_code", "refresh_token", "urn:ietf:params:oauth:grant-type:jwt-bearer"],
  "response_types": ["code"],
  "token_endpoint_auth_method": "none"
}
```

It declares **`urn:ietf:params:oauth:grant-type:jwt-bearer`**. Rauthy's `grant_types` validators don't accept that value, so Rauthy rejects the document on **both** paths — DCR and ephemeral/CIMD — with a 400. There are two different regexes:

- `validate_vec_grant_type` (DCR, `EphemeralClientRequest`/`DynamicClientRequest`):
  `^(authorization_code|client_credentials|urn:ietf:params:oauth:grant-type:device_code|password|refresh_token)$`
- the ephemeral/CIMD authorize path reports a narrower one (no `device_code`):
  `authorization_code|client_credentials|password|refresh_token`

### Steps to reproduce

DCR — Claude's exact payload:

```bash
curl -sS -X POST https://<rauthy>/auth/v1/clients_dyn -H 'Content-Type: application/json' -d '{
 "client_name":"Claude","client_uri":"https://claude.ai",
 "redirect_uris":["https://claude.ai/api/mcp/auth_callback"],
 "grant_types":["authorization_code","refresh_token","urn:ietf:params:oauth:grant-type:jwt-bearer"],
 "response_types":["code"],"token_endpoint_auth_method":"none"}'
# {"error":"BadRequest","message":"Payload validation error: ValidationErrors({\"grant_types\": ...
```

Identical payload minus `jwt-bearer` → **201**.

CIMD — pointing the authorize endpoint at Claude's real document:

```bash
curl -sS -G https://<rauthy>/auth/v1/oidc/authorize \
  --data-urlencode 'client_id=https://claude.ai/oauth/mcp-oauth-client-metadata' \
  --data-urlencode 'response_type=code' \
  --data-urlencode 'redirect_uri=https://claude.ai/api/mcp/auth_callback' \
  --data-urlencode 'scope=openid profile email' \
  --data-urlencode 'code_challenge=<...>' --data-urlencode 'code_challenge_method=S256'
# 400 — Payload validation error: grant_types
```

### Expected behavior

Unclear, and that's partly why I'm filing rather than sending a patch — you may reasonably decide this is claude.ai's bug. Options as I see them:

1. Accept-and-ignore unknown/unsupported `grant_types` in a fetched CIMD document (the document describes what the *client* can do; Rauthy already constrains what it will actually grant via `ephemeral_clients.allowed_flows`), rather than rejecting the whole document.
2. Add `jwt-bearer` to the accepted set (RFC 7523).
3. Reject only if the client actually *requests* the unsupported grant.

Option 1 seems most in the spirit of CIMD: `allowed_flows` is already the enforcement point, so a hard 400 on an advertised-but-unused grant type blocks an otherwise-valid client.

### Datapoint

Keycloak 26.7 (`--features=cimd`) also rejects this document, but for a different reason — `ClientRegistrationException: JWT authorization grant cannot be enabled in a public client` — since the doc pairs `jwt-bearer` with `token_endpoint_auth_method: none`. I verified that's the sole cause: same document minus `jwt-bearer` → 200. So arguably the root problem is upstream in claude.ai's document, and both ASes are collateral. Filing here so you can decide whether Rauthy wants to be lenient anyway.

### Version

0.36.0, Hiqlite backend. Related: #1562, #1643.

impuls42:
A CIMD document can advertise a grant type Rauthy doesn't implement. claude.ai lists `urn:ietf:params:oauth:grant-type:jwt-bearer` in its metadata but never actually requests it (it authenticates with `authorization_code` + `refresh_token`). Rejecting the whole document over an advertised but unused grant blocks an otherwise valid client.

Per your point in #1644, this is opt-in rather than unconditional leniency:

`ephemeral_clients.ignore_unknown_auth_flows` (default `false`; env `EPHEMERAL_CLIENTS_IGNORE_UNKNOWN_AUTH_FLOWS`). When it's on, `Client::ephemeral_from_url` strips unsupported grant types from a CIMD document before validation instead of rejecting it. Effective flows still come from `ephemeral_clients.allowed_flows`, so nothing unsupported is ever enabled. The flag only decides whether the document is accepted at all.

DCR and admin/bootstrap clients keep rejecting unknown grants unconditionally. Their advertised grants are stored verbatim as enabled flows, so an unknown one there is a real misconfiguration that should fail fast. That also covers your concern about silently ignoring a declared flow: on the ephemeral path the flow is never enabled in the first place, so no client ends up thinking it's using jwt-bearer.

Default behaviour is unchanged (reject everywhere).

Tests: the opt-in behaviour is covered by a unit test on the `retain_supported_grant_types` helper (strips the unsupported grant, keeps the rest). `handler_1644_grant_types.rs` pins the strict side, `/clients_dyn` with jwt-bearer still returns 400. That one locks in existing behaviour rather than reproducing a bug, since DCR was already strict. Checked against claude.ai's real document: with the flag on the ephemeral path accepts it and strips the grant, DCR with the same payload still 400s.

Closes #1644.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1476,6 +1476,21 @@ cache_lifetime = 3600
 # overwritten by: EPHEMERAL_CLIENTS_DANGER_ALLOW_UNVALIDATED_RESOURCE
 #danger_allow_unvalidated_resource = false
 
+# A dynamic client registration is always rejected when it advertises a grant
+# type Rauthy does not support. Some spec-valid CIMD clients (e.g. claude.ai)
+# advertise an unsupported grant (`urn:ietf:params:oauth:grant-type:jwt-bearer`)
+# in their metadata document without ever using it, which would otherwise cause
+# the whole document to be rejected. Setting this to `true` lets an ephemeral
+# (CIMD) client document through by sanitizing - stripping - the unsupported
+# grant types instead of rejecting it.
+#
+# This applies ONLY to ephemeral clients. Dynamic client registration (DCR) and
+# admin-managed clients always keep rejecting unknown grant types.
+#
+# default: false
+# overwritten by: EPHEMERAL_CLIENTS_IGNORE_UNKNOWN_AUTH_FLOWS
+#ignore_unknown_auth_flows = false
+
 [events]
 # The E-Mail address event notifications should be sent to.
 #
--- a/src/api_types/src/cust_validation.rs
+++ b/src/api_types/src/cust_validation.rs
@@ -8,20 +8,15 @@ use validator::ValidationError;
 
 #[inline]
 pub fn validate_vec_attr(value: &[String]) -> Result<(), ValidationError> {
-    let mut err = None;
-
     if value.is_empty() {
-        err = Some("'validate_vec_attr' cannot be empty when provided");
-    } else {
-        value.iter().for_each(|v| {
-            if !RE_ATTR.is_match(v) {
-                err = Some("^[a-z0-9-_/]{2,128}$");
-            }
-        });
+        return Err(ValidationError::new(
+            "'validate_vec_attr' cannot be empty when provided",
+        ));
     }
-
-    if let Some(e) = err {
-        return Err(ValidationError::new(e));
+    for v in value {
+        if !RE_ATTR.is_match(v) {
+            return Err(ValidationError::new("^[a-z0-9-_/]{2,128}$"));
+        }
     }
     Ok(())
 }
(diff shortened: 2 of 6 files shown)


config.toml
```diff
@@ -1476,6 +1476,21 @@ cache_lifetime = 3600
...
+# the whole document to be rejected. Setting this to `true` lets an ephemeral
+# (CIMD) client document through by sanitizing - stripping - the unsupported
+# grant types instead of rejecting it.
+#
+# This applies ONLY to ephemeral clients. Dynamic client registration (DCR) and
+# admin-managed clients always keep rejecting unknown grant types.
+#
+# default: false
+# overwritten by: EPHEMERAL_CLIENTS_IGNORE_UNKNOWN_AUTH_FLOWS
+#ignore_unknown_auth_flows = false
```
sebadob, line 1492: That block also needs to be added to `book/src/config/config.md`

src/bin/tests/handler_1644_grant_types.rs
```diff
@@ -0,0 +1,52 @@
...
+// as a dead flow. `urn:ietf:params:oauth:grant-type:jwt-bearer` is not supported by Rauthy and
+// must yield a 400. (The ephemeral/CIMD opt-in ACCEPT path - which *strips* unknown grants when
+// `ephemeral_clients.ignore_unknown_auth_flows` is enabled - fetches a hosted CIMD document and
+// is exercised by a separate live/real-env test; it is not driven here.)
+//
+// Without the fix (strict `validate_vec_grant_types`), the jwt-bearer grant would pass
+// validation and this request would be accepted (201) instead of rejected (400).
+// -----------------------------------------------------------------------------------------
+#[tokio::test]
+async fn test_1644_dcr_rejects_unsupported_grant() -> Result<(), Box<dyn Error>> {
```
sebadob: Please add the test to `src/bin/tests/handler_auth.rs` to keep the number of files low until we have the whole integration test setup reworked.

src/api_types/src/cust_validation.rs
```diff
@@ -84,22 +84,25 @@ pub fn validate_vec_contact(value: &[String]) -> Result<(), ValidationError> {
...
-        });
+        return Err(ValidationError::new(
+            "'flows_enabled' cannot be empty when provided",
+        ));
     }
-
-    if let Some(e) = err {
-        return Err(ValidationError::new(e));
+    for v in value {
+        if !RE_GRANT_TYPES.is_match(v) {
```
sebadob, line 87: Not a big deal, but If you do a `for` loop here you can `break` early on the first error. Actually, this can be simplified even more by having a direct `return`, which would get rid of the whole `if let Some(e) = err` thing`.

I know that the other cust validators need an update as well. I just noticed it. Haven't looked at them in a while. Would be nice if you could just do that as well.

sebadob (changes requested):

impuls42:
Addressed all three:
- documented `ignore_unknown_auth_flows` in `book/src/config/config.md`
- moved the test into `handler_auth.rs`
- simplified the grant-type validator with an early return, and gave the other `validate_vec_*` validators the same treatment (dropped the `let mut err` / `for_each` / `if let Some(e)` pattern).

sebadob (dismissed):
Thanks

sebadob:
The `grant_type`s will create a conflict with #1652 that might need to be resolved depending on what's merged first.

impuls42:
Makes sense. My suggestion, but I'm happy either way:

If it's easy to land the smaller ones first, #1655 / #1657 / #1658 / #1659 are all small and ready. #1652 already rewrites these files, so it would then just re-add `token-exchange` to `RE_GRANT_TYPES` and the accepted set on top of the for-loop validators from here. One small edit on their side.

If you'd rather land #1652 first, I'll rebase #1657 and #1658 onto it and resolve the grant_types overlap (and the resource one, since #1652 touches `clients.rs` / `oidc.rs` / `ClientConfig.svelte` / the resource test too) myself. Either direction I'll take care of our side so it's not extra work for you or @bartekus.

sebadob:
> If it's easy to land the smaller ones first

Absolutely, yes. I just noticed the conflict because I reviewd the other one before.

PR #1661, opened 2026-07-19
UI: keep the password reset / new account form inside short landscape viewports

### LINKED ISSUE #1660: Password reset / new-account page: Save button unreachable on iPad landscape (form taller than the page's fixed height budget)
## Environment

- Rauthy v0.36.0 (`ghcr.io/sebadob/rauthy:0.36.0`), Hiqlite
- Page: `/auth/v1/users/{id}/reset/{code}?type=new_user` (the `password_reset` variant is affected the same way)
- Field report device: iPad, Safari, **landscape**; portrait works

## What happens

During a real onboarding, a new user on an iPad in landscape could not reach the
**Save** button while setting their initial password — it stays hidden behind/below
the on-screen keyboard and nothing on the page scrolls far enough to reveal it.
Rotating to portrait is the only workaround.

## Where the problem lives (v0.36.0)

The page can never scroll as a document; every level is pinned to the layout viewport:

| File | Line | Rule |
|------|------|------|
| `frontend/src/css/global.css` | `body` block | `height: 100dvh; display: flex; overflow: auto` (nothing ever overflows it) |
| `frontend/src/lib/ContentCenter.svelte` | `<style>` | `height: 100%; display: flex; align-items: center` |
| `frontend/src/routes/users/{id}/reset/reset/+page.svelte` | 368–373 | `.container { max-height: calc(100dvh - 2.5rem); overflow-y: auto }` — the only scrollable element |
| `frontend/src/lib/PasswordPolicy.svelte` | 170–173 | single-column `ul` — 7 rows, ~200 px of the form's height |

The expanded password form is **691 px** tall (EN). Headless-Chromium measurements at
width 1180 (iPad landscape):

| viewport h | document scroll slack | `.container` scroll slack | Save top..bottom |
|-----------:|----------------------:|--------------------------:|-----------------:|
| 820        | 0                     | 0                         | 710..742         |
| 700 (≈ minus Safari chrome) | 0    | **31 px**                 | 665..**697**     |

So at iPad-landscape height the Save button sits **3 px from the viewport's bottom
edge** and the entire page owns **31 px** of scroll. On iOS the on-screen keyboard
shrinks only the **visual** viewport — `dvh` and the layout stay unchanged — so the
~390 px keyboard covers Save, and 31 px of travel cannot lift it out. Portrait works
because the ~1100 px layout viewport leaves the whole form above the keyboard.
(Desktop window-resizing never shows this: shrinking the window also shrinks `100dvh`,
which is why #1015's `max-height` fix looks correct everywhere except under an iOS
keyboard.)

## Preferred solution — make the form fit the page

Since the page deliberately owns its height (`100dvh`, centered), the robust fix is to
make the form's tall state fit realistic landscape heights instead of relying on the
inner scroller (which iOS keyboards defeat):

- `PasswordPolicy.svelte`: render the policy `ul` as a responsive grid
  (`repeat(auto-fit, minmax(14rem, 1fr))`) — two columns on wide pages, unchanged
  single column in narrow hosts like the account page.
- Reset page: tighten its own vertical rhythm (`h1`/`p` margins, `.typeChoice`,
  `.generate`).

Measured effect: form height **691 → 581 px** (EN); the minimum viewport height where
the whole form (incl. Save) is visible without any scrolling drops **~731 → ~621 px**,
which covers the tablet-landscape band (iPad Air ~700, iPad mini ~640). Verified over
a 20-viewport phone/tablet portrait+landscape matrix (Save reachability, no top
clipping, no horizontal overflow, no overlap of the absolute theme/language widgets
with the form): no regressions.

I'm submitting a PR implementing exactly this (CSS-only, two files).

## Alternate solution — let the page flow

Replace the fixed-height chain with normal document flow (`body { min-height: 100dvh;
height: auto }`, `align-items: safe center` or child `margin: auto` in
`ContentCenter`, drop the `.container` clamp). With content in normal flow, the
browser can scroll the focused field and Save into view even with the keyboard up.
Bigger blast radius (every page using `ContentCenter`), and `safe center` support in
Safari is still shaky — hence not the preferred option here.

## Possible follow-up (out of scope)

On narrow phone portrait (e.g. 375×667) the form is still taller than the viewport
(725 px after the fix) and uses the scroll path; a full no-scroll fit there would need
content changes (collapsible policy list / shorter intro copy).

CognitiveSand:
Fixes #1660.

## Problem

On tablets in landscape (layout height ~620–730 px once browser chrome is
subtracted) the expanded password form on
`/auth/v1/users/{id}/reset/{code}` is taller (691 px, EN) than the page's height
budget. The document itself never scrolls (`body`/`Main`/`ContentCenter` are pinned to
`100dvh`) and the `.container` clamp leaves ~31 px of inner scroll, so the Save button
sits ~3 px from the viewport's bottom edge. On iOS the on-screen keyboard shrinks only
the visual viewport — not `dvh` — so the keyboard covers Save with no scroll path left.
Real-world effect: a new user on an iPad in landscape cannot save their initial
password (rotating to portrait is the workaround). Full analysis and measurements in
#1660.

## Change (CSS-only, two files)

- `frontend/src/lib/PasswordPolicy.svelte` — policy `ul` becomes a responsive grid:
  `repeat(auto-fit, minmax(14rem, 1fr))`. Two columns on wide pages; narrow hosts
  (account page) stay single-column and visually unchanged.
- `frontend/src/routes/users/{id}/reset/reset/+page.svelte` — tighter vertical rhythm
  scoped to this page: `h1` 1rem→0.5rem, `p` 0.5rem→0.25rem, `.typeChoice`
  1rem→0.5rem, `.generate` 0.66rem→0.25rem. The existing `max-height` clamp stays as
  the fallback for even shorter viewports.

## Measured effect

Form height 691 → **581 px** (EN); minimum viewport height with Save fully visible and
zero scrolling: **~731 → ~621 px**. Checked over a 20-viewport matrix (Galaxy S8+,
iPhone SE/15 Pro/15 Pro Max, Pixel 7, iPad mini/Air/Pro 12.9 — portrait and landscape —
plus chrome-reduced and keyboard-strip heights), asserting per viewport: Save reachable
(directly or via the container's scroll), no top clipping, no horizontal overflow, no
overlap between the absolute theme/language widgets and the form. No regressions;
the tablet-landscape band that produced the field report now fits without scrolling.
Happy to also contribute the matrix script if you want it in the repo.

## Notes

- `prettier --check` passes on both files; `svelte-check` reports only the two
  pre-existing `src/wasm/*` module errors (generated artifacts not present in a fresh
  clone without the Rust toolchain).
- Narrow phone portrait (375×667) still uses the scroll path (form is 725 px there);
  a full no-scroll fit would need content changes (e.g. collapsible policy list) —
  left out deliberately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TkiUm4ngAaGoUwuV5n3mVU

### DIFF
--- a/frontend/src/lib/PasswordPolicy.svelte
+++ b/frontend/src/lib/PasswordPolicy.svelte
@@ -170,6 +170,11 @@
     ul {
         margin-left: 1rem;
         list-style: none;
+        /* Two columns when the host page is wide enough (halves the block's
+           height on short landscape viewports), single column otherwise. */
+        display: grid;
+        grid-template-columns: repeat(auto-fit, minmax(14rem, 1fr));
+        column-gap: 1rem;
     }
 
     li {
--- a/frontend/src/routes/users/{id}/reset/reset/+page.svelte
+++ b/frontend/src/routes/users/{id}/reset/reset/+page.svelte
@@ -372,6 +372,17 @@
         overflow-y: auto;
     }
 
+    /* Keep the whole form (incl. the Save button) inside short landscape
+       viewports (e.g. iPad landscape ~700px, where the on-screen keyboard
+       shrinks only the visual viewport and hides anything below it). */
+    .container h1 {
+        margin: 0.5rem 0;
+    }
+
+    .container p {
+        margin: 0.25rem 0;
+    }
+
     .err {
         margin-top: 0.5rem;
         max-width: 20rem;
@@ -388,7 +399,7 @@
     }
 
     .generate {
-        margin-bottom: 0.66rem;
+        margin-bottom: 0.25rem;
     }
 
     .policy {
@@ -397,7 +408,7 @@
     }
 
     .typeChoice {
-        margin-bottom: 1rem;
+        margin-bottom: 0.5rem;
         display: flex;
         gap: 0.5rem;
     }


frontend/src/lib/PasswordPolicy.svelte
```diff
@@ -170,6 +170,11 @@
     ul {
         margin-left: 1rem;
         list-style: none;
+        /* Two columns when the host page is wide enough (halves the block's
+           height on short landscape viewports), single column otherwise. */
+        display: grid;
+        grid-template-columns: repeat(auto-fit, minmax(14rem, 1fr));
+        column-gap: 1rem;
```
sebadob, line 177: That screws up the existing account dashboard for no reason when there is more than enough space in every direction. Also working with fixed `14rem` here can have very unexpected results in combination with i18n.

sebadob (changes requested):

PR #1662, opened 2026-07-19
Fix: forward RFC 8707 `resource` on the session-refresh login (SPA, follow-up to #1657)

impuls42:
Follow-up to #1657 (issue #1645), covering the frontend half.

#1657 made the backend accept and carry `resource` on `POST /oidc/authorize/refresh`, but the login SPA only forwards `resource` on the full-login request, not on the silent re-auth. So once a session exists, a returning user's authorize goes through `/authorize/refresh` without the resource, the re-issued auth code has no resource, and the token exchange then fails with "the requested `resource` was not granted at the authorization request" (or, if the client does not re-request it, the token silently loses its audience binding).

The full-login path already does this a few lines below; this just adds the same conditional to the refresh payload:

```js
if (resource) {
    payload.resource = resource;
}
```

plus the `resource` field on the `LoginRefreshRequest` type, matching `LoginRequest`.

Reproduced against a real client (claude.ai as an MCP OAuth client with an RFC 8707 resource): first login worked, every returning-session login lost the `…/mcp` audience and the resource server rejected the token. Logging out (forcing the full-login path) confirmed the difference, and this change makes the refresh path behave the same.

Stacked on #1657, so until that merges the diff here also shows its commit; I will rebase once it lands. No backend change. The backend contract is already covered by the integration test on that branch (`handler_1645_resource_refresh.rs`, which posts `resource` to `/authorize/refresh` and asserts the exchanged token keeps it in `aud`); rauthy has no SPA test harness for the login page, so the frontend side is verified by the live reproduction above.

### DIFF
--- a/frontend/src/api/types/authorize.ts
+++ b/frontend/src/api/types/authorize.ts
@@ -40,6 +40,9 @@ export interface LoginRefreshRequest {
     /// Validation: PATTERN_CODE_CHALLENGE
     code_challenge?: string;
     code_challenge_method?: CodeChallengeMethod;
+    /// RFC 8707 resource indicator forwarded from the authorization request.
+    /// Validation: PATTERN_URI
+    resource?: string;
 }
 
 export interface RequestResetRequest {
--- a/frontend/src/routes/oidc/authorize/+page.svelte
+++ b/frontend/src/routes/oidc/authorize/+page.svelte
@@ -228,6 +228,9 @@
             payload.code_challenge = challenge;
             payload.code_challenge_method = challengeMethod;
         }
+        if (resource) {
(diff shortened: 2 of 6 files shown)


sebadob:
This PR contains a lot more than the actual issue it is about. It also contains all the changes from your other PR about resource. Do you want to unify them? This would be less messy.

Maybe just push the additional changes from this one into the other where I commented already some things. Almost everything from this PR is a duplicate when reviewing.

impuls42:
Folded these SPA changes into #1657 as you suggested, so all of the `resource` work lives in one PR now. Closing this one to avoid the duplication.

Issue #1617, opened 2026-07-01
Server-side idp_hint 302 auto-forward on /oidc/authorize


bartekus:
# RFC: server-side `idp_hint` 302 auto-forward on `/oidc/authorize`

## Summary

When a relying party deep-links a user straight at a known upstream identity provider
via `GET /auth/v1/oidc/authorize?...&idp_hint=<provider_id>`, Rauthy currently forwards
to that provider from the SPA, after the login page has already booted and rendered.
The result is a visible flash of the Rauthy login card before the browser is redirected.
This RFC proposes a server-side 302 that forwards to the upstream provider before any
SPA is served, when (and only when) `idp_hint` names a known and enabled provider, and
falls through to the existing SPA render otherwise. It reuses the existing
`login_start` redirect builder. It also surfaces one real blocker (upstream callback
completion depends on two client-persisted secrets) that needs a maintainer ruling
before a PR is worth writing.

Verified against HEAD `f6223305`, 2026-07-01.

## 1. Problem statement

Today the forwarding is done client side. `frontend/src/routes/oidc/authorize/+page.svelte:63`
reads `idp_hint` from the query string, and the `$effect` at
`+page.svelte:153-161` waits until the provider list has loaded, matches the hint against
`providers.find(p => p.id === idpHint)`, and calls `providerLogin(provider.id)`
(`+page.svelte:414-448`). By the time that effect fires, the SPA has already mounted and
painted the login card. For a plain deep link into a single known IdP, the user sees:

1. blank page,
2. Rauthy login card with the provider button,
3. a redirect to the upstream provider.

Step 2 is a pointless flash. It is most visible on cold loads and slow links, and it
undercuts IdP-initiated-style deep links where the intent is "send me straight to
Provider X". A server-side 302 emitted from `get_authorize` removes step 2 entirely: the
first response the browser receives is already the redirect to the upstream authorize URL.

Concrete UX scenario: a portal links users to
`https://rauthy.example.com/auth/v1/oidc/authorize?client_id=app&redirect_uri=...&response_type=code&scope=openid&idp_hint=<24charProviderId>`.
With the server-side path, that navigation lands the user directly on the upstream login
page. No Rauthy card is ever shown.

## 2. Current mechanism (why a naive 302 is not a drop-in)

The client path is not just "POST then follow Location". `execProviderLogin`
(`frontend/src/utils/login.ts:13-42`) does three things that matter here:

[...]

PR #1664, opened 2026-07-22
Fix: make an upstream auth provider callback single-use

bartekus:
The standalone fix you asked for in #1617, ahead of the `idp_hint` work.

## The bug

`login_finish` deletes the `AuthProviderCallback` when the `state`, the XSRF token or the PKCE verifier fails to validate. Once all three pass, nothing deletes it: not the success path, and not any of the error paths below the validation either. It lives on in the cache until `CACHE_TTL_AUTH_PROVIDER_CALLBACK` expires it 300 seconds later.

So a callback that has already been used stays valid for the rest of that window. The doc comment on the function already claimed otherwise.

## What it is worth in practice

Bounded, and worth being straight about. Reaching the callback needs the encrypted `HttpOnly` cookie, and a replay re-submits the same upstream `code`, which any conforming provider rejects as already redeemed under RFC 6749 §4.1.2. So this is not a live account-takeover path today.

What it does mean is that the single-use invariant the code documents is not the invariant the code has, and the gap is entirely dependent on upstream providers behaving. It matters more for #1617: on the server-side path the callback cookie is the whole credential on the finish leg, since there is no XSRF echo from a client that never ran, so a five minute replay window becomes a real part of the threat model rather than a redundancy behind one.

## The fix

Delete it the moment it validates, which is exactly the point where this request has proven it is the one entitled to consume it. That covers the success path, the account-federation early return, and every error below the validation in one place instead of adding a fourth, fifth and sixth delete call.

Nothing downstream needs the callback to still exist: `slf` is already in hand, and the deletion cookie the success path returns is unchanged.

## One thing worth flagging

My first version moved the id out with `mem::take` to avoid the clone. That compiles, passes the whole suite, and silently breaks every ATProto login: `extract_user_at_proto()` matches the state ATProto hands back against `self.callback_id` (`auth_providers.rs:872`), which would then be an empty string. It is a clone with a comment saying why, so nobody, including me, turns it back into a take later. The allocation is a rounding error next to the HTTP round trip to the upstream IdP on the same path.

## Testing

No integration test, and I want to be explicit rather than quiet about it. Driving this needs a seeded `AuthProvider` and an upstream to call, which is the same gap `zzc_backchannel_logout.rs:27-30` works around and which you described in #1617 as needing a mocked provider through the `/dev` endpoints. That harness belongs with the `idp_hint` PR, where it is needed anyway, and it will cover this as a side effect: the replay assertion is one extra request on the round-trip case. Bundling it here would mean building it twice.

Verified: `cargo clippy --all-features -- -D warnings` clean, `cargo fmt --check` clean, full integration suite green on hiqlite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ all three stayed in the cache until its 300 second TTL ran out, whether the logi
 failed, which left it replayable for the rest of that window by anyone holding the encrypted
 callback cookie. This also makes the behaviour match what the function documented all along.
 
-[#1660](https://github.com/sebadob/rauthy/pull/1660)
+[#1664](https://github.com/sebadob/rauthy/pull/1664)
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,14 @@
 
 ## Unreleased
 
-### Security
+### Bugfix
 
 #### Upstream Auth Provider Callbacks are Single-Use
 
 A callback for an upstream auth provider login is deleted as soon as it has been validated now.
 Before, only a failed `state`, XSRF token or PKCE verifier check deleted it. A callback that passed
 all three stayed in the cache until its 300 second TTL ran out, whether the login then succeeded or
-failed, which left it replayable for the rest of that window by anyone holding the encrypted
-callback cookie. This also makes the behaviour match what the function documented all along.
+failed. This also makes the behaviour match what the function documented all along.
 
 [#1664](https://github.com/sebadob/rauthy/pull/1664)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+### Security
+
+#### Upstream Auth Provider Callbacks are Single-Use
+
+A callback for an upstream auth provider login is deleted as soon as it has been validated now.
+Before, only a failed `state`, XSRF token or PKCE verifier check deleted it. A callback that passed
+all three stayed in the cache until its 300 second TTL ran out, whether the login then succeeded or
+failed, which left it replayable for the rest of that window by anyone holding the encrypted
+callback cookie. This also makes the behaviour match what the function documented all along.
+
+[#1660](https://github.com/sebadob/rauthy/pull/1660)
+
 ### Changes
 
 #### Session MFA Upgrade after Passkey Registration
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+### Security
```
sebadob: This is not a security issue, it's more a bugfix / cleanup thing.

The `AuthProviderCallback` does not hold any security-related information. It contains an XSRF token, but that lives on the client side anyway, which is the only place from where you could replay something. So that only makes sure the request comes from the exact same browser context. 

The security-related parts are in `ProviderCallbackRequest`, mainly the `code` inside it. That lives on the client side, and we have no control over it. It even shows up in the browser history.

This makes it more a cleanup / memory efficiency fix rather than a security fix. I would move it under `Bugfix`.

src/service/src/oidc/auth_providers/login_finish.rs
```diff
@@ -66,6 +68,16 @@ pub async fn login_finish<'a>(
         ));
     }
 
+    // The callback is validated at this point, which makes this the one request that is allowed
+    // to consume it. Deleting it here rather than at the end makes it single-use for real: the
+    // success path never deleted it at all and relied on the TTL, and the error paths below this
+    // line never did either, both of which left a validated callback replayable for the rest of
+    // its `CACHE_TTL_AUTH_PROVIDER_CALLBACK`.
+    //
+    // The id is cloned rather than moved out of `slf`, because `extract_user_at_proto()` still
```
sebadob: Remove the comment about why cloning the `id`. Cloning values is necessary when they are used later. I would only add such an explaining comment if you actually `mem::take()` something out of a bigger `struct` that is re-used later on, where it's not directly obvious that not the full thing is needed.

Also, I would remove all that other explaining part which only makes real sense in the context of this issue. I would only keep 

```
The callback is validated at this point, so we can safely clean up the cache.
```

I like to keep the amount of comments in code low. It's very hard to review code bases that have comments above almost all steps in a function. Only in cases like these where the whole logic is very complex and you must not make any mistakes, it makes sense to have a bit more.

In general, code should be self-documenting. Only when doing something very complex or unexpected, I like to know the thought behind it. Comments need maintenance as well, and they often get out of date.

sebadob (changes requested):

bartekus:
Both applied in `edbf60f8`.

Moved under `Bugfix` and dropped the replay framing from the entry text. You are right that the XSRF token is the only thing in there and it lives client-side anyway, so nothing in `AuthProviderCallback` is worth protecting on its own; the `code` in `ProviderCallbackRequest` is the part that matters and this fix does not touch it.

The inline comment is now the single line you quoted. I also cut the doc comment on the function down to one line while I was in there, on the same reasoning: the only part worth stating is that the callback is deleted on validation, which is exactly what the old one got wrong when it said "in case of any error".

sebadob (approved):
Thanks

PR #1667, opened 2026-07-24
fix: avoid persistent cache state with PostgreSQL

caniko:
## What changed

When PostgreSQL is selected as Rauthy's durable database, force Hiqlite cache storage to memory-only. Hiqlite remains available for request caching and coordination, but its cache WAL and snapshots are no longer persisted.

## Why

PostgreSQL is the source of truth in this mode. A persistent Hiqlite cache can survive a PostgreSQL cluster rollover and return objects that no longer exist in the active database. During recovery, that can expose stale users, providers, or other cached objects.

Native Hiqlite mode keeps its existing cache-storage setting.

## Scope

This intentionally does not remove the Hiqlite runtime cache/control plane. PostgreSQL-only without Hiqlite startup is a larger architectural change and should be handled as a separate follow-up.

## Validation

- `cargo fmt --all -- --check` passed.
- Added unit tests for PostgreSQL forcing cache persistence off and native Hiqlite preserving the configured setting.
- `git diff --check` passed.
- `cargo test -p rauthy-data database::tests` was attempted in the Nix development shell but the clean checkout could not compile because generated `templates/html/*` artifacts are absent.
- `just build-ui` was attempted after `npm ci`, but the checkout also lacks generated WASM modules and the available Nix Rust sysroot does not include `wasm32-unknown-unknown`; no generated files were fabricated.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -675,7 +675,8 @@ log_sync = "interval_200"
 # In-memory only significantly increases the throughput though, so it
 # depends on your needs, what you should prefer.
 #
-# default: true
+# default: true for Hiqlite. PostgreSQL mode always forces this to false so
+# the cache cannot outlive the PostgreSQL database it is derived from.
 # overwritten by: HQL_CACHE_STORAGE_DISK
 cache_storage_disk = true
 
--- a/src/data/src/database.rs
+++ b/src/data/src/database.rs
@@ -66,16 +66,16 @@ pub struct DB;
 
 impl DB {
     /// Builds the NodeConfig and starts the Hiqlite node
-    pub async fn init(config: hiqlite::NodeConfig) -> Result<(), ErrorResponse> {
+    pub async fn init(mut config: hiqlite::NodeConfig) -> Result<(), ErrorResponse> {
         if HIQLITE_CLIENT.get().is_some() {
             panic!("DB::init() must only be called once at startup");
         }
 
-        // Note: At the time of writing, I have not included the option to not start the DB
-        // layer when the feature is enabled, as this would mean many adoptions and additional
-        // checks in the Hiqlite code.
-        // This means, even if Postgres is configured, this will still create and start the Hiqlite
-        // DB, but it simply won't do anything else than heartbeats between the nodes.
+        // PostgreSQL is the source of truth in this mode. Keep the Hiqlite cache available for
+        // request-level caching and coordination, but do not persist its WAL or snapshots: a
+        // persistent cache can outlive a PostgreSQL cluster rollover and return objects which no
+        // longer exist in the active database.
+        configure_cache_storage(&mut config, is_postgres());
         let client = hiqlite::start_node_with_cache::<Cache>(config).await?;
 
         if is_postgres() {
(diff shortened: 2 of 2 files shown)


sebadob:
> When PostgreSQL is selected as Rauthy's durable database, force Hiqlite cache storage to memory-only. 

That's a terrible idea. The current setup exists for a reason.

> Hiqlite remains available for request caching and coordination, but its cache WAL and snapshots are no longer persisted.

That's not how it works. Snapshots are always written to disk in any case without an additional feature you need to set for `hiqlite`, which Rauthy does not do. Only the WAL can be kept in-memory with the config setting. But, even though technically you can achieve higher write throughput, you also have higher memory requirements. The throughput is usually not needed, but using less memory is something everyone wants.

> PostgreSQL is the source of truth in this mode.

Not entirely true either. Some things live inside the cache only, like callback states or auth codes, and other things. This is the main reason why it's persistent by default. Authentication codes (and other things) would not outlive a restart otherwise.

> A persistent Hiqlite cache can survive a PostgreSQL cluster rollover and return objects that no longer exist in the active database.

If you apply a backup, you need to remove the cache dirs right now, that's true. I have an easier fix for that on my TODO, but not implemented yet. Enforcing an ephemeral cache is definitely not the solution. These are solutions that would work:

1. Parse an env var at startup to force a full `cache_clear()`. That function already exists. Some caches are cleared between restarts already. You could parse something like `RAUTHY_FULL_CACHE_CLEAR` at startup and if `true`, clear all existing caches.
2. Add an API endpoint to be able to clear caches. That might come in handy for debugging in some cases. That would absolutely require authentication though.

---

Apart from all of that, theoretically it's possible to not even start Hiqlite when using Postgres, but that would require thousands of lines of reworked code, and changed caching behavior. The end result would be a lot less responsive application, a lot more network traffic, and more resource usage.

caniko:
Thousands of lines sounds very dramatic... Why don't we have an abstraction of the specific SQL DB used so that any DB can take hiqlite's place?

sebadob:
> Thousands of lines sounds very dramatic

It is very dramatic. The cache is accessed in lots of places. An abstraction over cache GET, PUT, DELETE and CLEAR would be the first step to make this work. Then migrate everything to use it. That should be already at least a few hundred lines. 

The next step would be to move all data that now lives inside the cache to the DB layer, but conditionally. This means keeping the current code for hiqlite-based deployments in place, and adding everything necessary to optionally push it to the DB instead. This needs additional migrations and queries, and it needs an additional abstraction over the current methods to decide which route to take.

Then you either fully drop all other cached data like prerendered HTML and such, which would waste more resources at runtime, or add a completely new local cache for each instance with something like `cached`, but again, only conditionally.

And then you need to migrate all in-cluster real-time communication as well. This was based on PG listen / notify in the past. Yes, you could do that, also only conditionally and in addition.. All of these decisions have been made for good reason. The Raft-based Hiqlite also comes with a stronger consistency than Postgres (without clustering and forcing synchronous commits on multiple instances).

> Why don't we have an abstraction of the specific SQL DB

I had that in the past, and I got rid of it for a reason. I used `sqlx` back then with their `Any` driver, and it broke so many times out of the blue. Some things even that badly that they did not even care about my bug reports, and things never got fixed, only ignored. When you allow multiple DB's you are also limited writing queries, because they must work on all options. Let's say you wanted to make MySQL possible, then you have even more branching and conditions everywhere, because it does not even understand `ON CONFLICT ...` or `RETURNING`, and you would need to handle DB deadlocks.

Also, more abstractions means slower code. This is always the case. I don't care about any other DB than Hiqlite or Postgres. Postgres already only exists for backwards-compatibility, since it's the DB the project started with years ago. If I would start it from scratch today, I would not even add Postgres support. There is just no need to use anything else than Hiqlite, especially if you want to keep things simple.

Hiqlite does not only provide the DB, but also the cache, the real-time communication and access to some (very few) distributed counters. This means either you use Postgres for all of that (and make everything much slower), or you go completely nuts and use special deployments for each case, like Postgres for the DB, Redis for the cache, and something like Nats for in-cluster traffic. All of that is completely unnecessary.

I simply don't see the need or any good reason to do this. Only because you can do things like that and abstract everything doesn't mean it's a good idea.

Edit:

I totally forgot, when you move the Cache into Postgres, you also need to manually handle TTL and cleanup as well, which means additional checks and schedulers, or you really need another deployment like Redis that does that for you.

caniko:
That is very illuminating, thank you. I am closing to continue the discussion at the issue level, I was not aware that postgres was a backwards compatibility layer.

sebadob:
Did not have to be closed. The implementation of a way to clear the cache on-demand makes a lot of sense, and I have that on my TODO anyway. During development, there is already a block that clears most caches with each startup. I would probably rework that slightly and user an `EnumIter` for future stability, but other than that, it could be re-used if the ENV var is set.

caniko:
I closed it in favor of #1668, we can of course open it again if you think it is worth discussing in this PR. It seemed to me that you preferred to take it in another direction than what I had proposed.

PR #1677, opened 2026-07-24
feat(fido-mds): ship and seed the FIDO MDS dataset

bartekus:
Refs #1618. This is PR 1 of the split we agreed on there: the dataset and its plumbing, with nothing reading it yet. Enforcement (the ceremony change and the actual filters) follows in later PRs, so this one is non-breaking and existing logins are untouched.

Ready for review. It is rebased onto current `main` (v0.36.2) and I am done rewriting the branch, so nothing here should reset your read state; any review feedback will land as follow-up commits on top rather than a force-push. The two decisions at the bottom are flagged for your call during review, not gates before it: if you want either one different, say so and I will change it in place.

## What it does

- Adds a `rauthy-data` module (`fido_mds`) that transforms the signed MDS blob into a compact prepared form and persists it.
- Adds three tables (hiqlite + postgres migrations) and a nullable `passkeys.aaguid` column with an `ON DELETE SET NULL` foreign key. Nothing writes that column yet.
- Ships the prepared dataset as one embedded asset and seeds it on startup when the tables are still empty, so both a fresh instance and one upgrading into this version get it.
- Adds the offline `fido-mds-prep` bin target that regenerates the asset, driven by `just fido-mds-prep`.

## The transform, and the two traps behind it

Everything correctness-sensitive is in `fido_mds/raw.rs`, covered by `fido_mds/tests.rs`. Two rules that look obvious are wrong against the real blob, so the tests pin them:

- **`statusReports` is not chronologically ordered.** The one REVOKED entry in the blob lists `REVOKED` *before* its older `NOT_FIDO_CERTIFIED`, so a "latest report" revocation check keeps the one authenticator it exists to drop. Revocation is therefore "any report", and the four `*_COMPROMISE` / `USER_VERIFICATION_BYPASS` statuses drop the entry too.
- **Certification level is never in the last report.** The level and a bare `FIDO_CERTIFIED` are separate reports on the same entry, so the level is derived as the max across all reports, stored as one ordered value (`NotCertified < Certified < L1 < ... < L3Plus`).

Run against the live blob (no. 271), the transform yields exactly the numbers from the issue: 330 entries kept, 240 distinct root certs (down from 705 references).

## Schema shape

- `fido_mds_certs` (hash PK, DER bytes): the deduplicated roots.
- `fido_mds_entries` (aaguid PK as 16 bytes, description, three `u32` bitmasks, the ordered `cert_level`). Enum order is wire-stable with `Unknown` pinned to bit 0, so the columns can widen past 32 bits later; postgres stores them as `bigint`.
- `fido_mds_entry_certs`: the `n:m` join.
- `passkeys.aaguid`: nullable, `ON DELETE SET NULL` to `fido_mds_entries`. I verified on SQLite that adding this column to a populated `passkeys` works and that deleting an entry nulls the referencing passkey and cascades away its join rows, which is the reject-but-still-show behaviour you wanted.

## Two calls I did not want to make silently

1. **The shipped dataset is an embedded asset, not inline SQL.** A regenerated several-hundred-KiB migration per release would accumulate forever, so the schema is a normal migration and the *data* is `include_bytes!`'d and applied through the same insert path a refresh will use. Say the word if you would rather it be a data migration proper.
2. **The prep bin lives in the `rauthy` package** as a second `[[bin]]` with `default-run = "rauthy"`, so `cargo run` is unchanged and the transform stays single-sourced in `rauthy-data`. The tradeoff is that building `--bin fido-mds-prep` compiles the server's dependency set, but those are already built for the server, so running the recipe reuses them.

## Verification

- `cargo clippy -- -D warnings` and `cargo fmt --check` clean across the workspace, re-run after the rebase onto v0.36.2.
- The six transform tests pass; the SQLite FK-cascade behaviour above is verified.
- I have not run the `zz*` integration suite against a live pair of DBs for the seed path yet; the migrations are modelled directly on existing ones and the SQLite one is verified, but flag if you want me to add a seed assertion to the suite.

Still open on the issue: whether the four filters are global or per-client (you leaned global, which does not affect this PR), and the `Ed25519` question is unrelated.


sebadob:
I have not even taken a look at this, because it's marked as a draft. Reviewing drafts is actually double the work. E.G. after another force-push, lots of files are marked as never read again, which means I need to read everything at least twice.

However, I will take a first quick look later.

bartekus:
Fair, and that is my mistake: I marked it draft to invite feedback on two decisions before building PR 2 on top, without thinking about what repeated force-pushes cost you on the reading side. Fixed now.

It is out of draft and rebased onto current `main` (v0.36.2). I did the rebase first and flipped it to ready second, deliberately, so the state you start reading from is the final one. I am done rewriting this branch: anything that comes out of your review lands as a follow-up commit, not a force-push.

The rebase was only `CHANGELOG.md` (my entry now sits under `## Unreleased` above v0.36.2) and `Cargo.lock`; no code conflicted, and the migration numbers `V27` / `32_` are still unclaimed. `cargo clippy -- -D warnings`, `cargo fmt --check`, and the six transform tests are green on the rebased tip.

The two decisions I wanted your reaction to are in the PR body under "Two calls I did not want to make silently" (embedded asset rather than an ever-growing data migration, and the prep bin in the `rauthy` package via `default-run`). They are not gates on your review: read it as normal, and if you want either one different I will change it in place.

migrations/hiqlite/32_fido_mds.sql
```diff
@@ -0,0 +1,39 @@
+CREATE TABLE fido_mds_certs
+(
+    hash BLOB NOT NULL
+        CONSTRAINT fido_mds_certs_pk
+            PRIMARY KEY,
+    der  BLOB NOT NULL
```
sebadob, line 6: I would rename this column into `ca_der` or `cert_der` to make it very clear what kind of data it is, since it's just bytes.

src/bin/src/fido_mds_prep.rs
```diff
@@ -0,0 +1,75 @@
...
+const DEFAULT_SOURCE: &str = "https://mds.fidoalliance.org/";
+const DEFAULT_OUT: &str = "assets/fido_mds/dataset.bin";
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let mut source = DEFAULT_SOURCE.to_string();
+    let mut out = PathBuf::from(DEFAULT_OUT);
+
+    let mut args = std::env::args().skip(1);
+    while let Some(arg) = args.next() {
```
sebadob, line 24: It makes a lot of sense to make use of `clap` here to get rid of all of this manual parsing. It's in the workspace already anyway, and it makes adding future features like e.g. parsing the DB to extract specific AAGUIDs a lot easier.

src/bin/src/fido_mds_prep.rs
```diff
@@ -0,0 +1,75 @@
...
+                     --source  MDS blob URL or local file (default: {DEFAULT_SOURCE})\n  \
+                     --out     output path (default: {DEFAULT_OUT})"
+                );
+                return Ok(());
+            }
+            other => return Err(format!("unknown argument: {other}").into()),
+        }
+    }
+
+    let jwt = if source.starts_with("http://") || source.starts_with("https://") {
```
sebadob, line 44: We should probably deny downloads via `http`, because in this case the payload could be tampered with, and we would lose the passive validation of the dataset. We should also force an `https://` in front in the `else` below.

src/bin/Cargo.toml
```diff
@@ -5,11 +5,18 @@ edition.workspace = true
 authors.workspace = true
 license.workspace = true
 description.workspace = true
+# keeps `cargo run` on the server binary now that a second bin target exists
```
sebadob, line 8: We can get rid of this comment, only makes sense in the context of this PR.

src/data/src/fido_mds/db.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::fido_mds::PreparedDataset;
+use crate::rauthy_config::RauthyConfig;
+use hiqlite::Params;
+use hiqlite::macros::params;
+use rauthy_common::is_hiqlite;
+use rauthy_error::ErrorResponse;
+use tracing::info;
+
+/// The dataset shipped with the image, produced by the `fido-mds-prep` tool.
+static EMBEDDED: &[u8] = include_bytes!("../../../../assets/fido_mds/dataset.bin");
```
sebadob, line 17: Change the name to something more meaningful like `MDS_DATASET`. In the code logic below, we don't care if it's embedded, but the name would make more sense.

src/data/src/fido_mds/db.rs
```diff
@@ -0,0 +1,121 @@
...
+
+const SQL_CERT: &str = "INSERT INTO fido_mds_certs (hash, der) VALUES ($1, $2)";
+const SQL_ENTRY: &str = "INSERT INTO fido_mds_entries \
```
sebadob, line 20: Using the `r#"INSERT ..."#` syntax makes a bit cleaner Strings.

src/data/src/fido_mds/db.rs
```diff
@@ -0,0 +1,121 @@
...
+    (aaguid, description, key_protection, attachment_hint, attestation_types, cert_level) \
+    VALUES ($1, $2, $3, $4, $5, $6)";
```
sebadob, line 22: We probably need an `ON CONFLICT DO UPDATE SET ...` here, right?

src/data/src/fido_mds/db.rs
```diff
@@ -0,0 +1,121 @@
...
+            .await?
+            .remove(0)
+            .get("count")
+    } else {
+        DB::pg_query_rows(sql, &[], 1).await?.remove(0).get("count")
+    };
+    Ok(count)
+}
+
+/// Insert a full dataset. Certs and entries are written before the join rows so the foreign keys
```
sebadob, line 61: I would say we can get rid of this comment. The function name already says what it's doing, and there is no benefit mentioning FK logic that is necessary inside the function.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+//! Nothing reads this data yet. Enforcement (rejecting a passkey whose authenticator is not in, or
+//! not certified strongly enough by, this dataset) is a separate, opt-in step.
+
+pub mod db;
+mod raw;
+
+#[cfg(test)]
+mod tests;
+
+use serde::{Deserialize, Serialize};
```
sebadob, line 19: I usually like the following order for imports, just for consistency:

- `use`es
- `mod`s
- `static`s
- `const`s
- rest

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// The bincode config used for the shipped dataset and any cached form.
+const BINCODE: bincode::config::Configuration = bincode::config::standard();
```
sebadob, line 22: You cannot know this, but there is a future TODO to migrate away from `bincode`, since it's not being maintained anymore. There are global serialization functions in `src/common/src/utils.rs` we should re-use. The more we use these upfront, the more we probably can reduce the migration issues and code changes when we finally do it. This is also using the "legacy" config on purpose. If we are lucky, we might be able to migrate to `wincode` without any issues at all.

src/data/src/fido_mds/db.rs
```diff
@@ -0,0 +1,121 @@
...
+/// resolve.
+async fn insert(dataset: &PreparedDataset) -> Result<(), ErrorResponse> {
```
sebadob, line 63: Making this an `impl PreparedDataset` with `async fn insert(&self)` would be a bit cleaner.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// A fully transformed MDS dataset, ready to be written to the database as-is.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct PreparedDataset {
```
sebadob, line 26: Naming it `MdsDataset` makes more sense when you call it from outside of this file. Otherwise, the name would be very opaque and only make sense with `fido_mds::` in front, which we don't want to use when we call our own structs.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+    /// The monotonic MDS blob number (`no`). A higher value is newer.
+    pub blob_no: i64,
+    /// The `nextUpdate` date the blob advertises, kept verbatim for the scheduler.
+    pub next_update: String,
```
sebadob, line 30: This always returns a formatted date, right? We could then parse it as `chrono::Utc` or `chrono::NaiveDate` for type safety, or we convert it to a UTC timestamp when parsing, which would make it an `i64` which is more efficient and never needs additional parsing. Rename it `next_update_ts` and it's clear that it's a unix timestamp in secods.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub entries: Vec<PreparedEntry>,
+    /// Every distinct root certificate referenced by `entries`, deduplicated by hash.
+    pub certs: Vec<PreparedCert>,
+}
+
+impl PreparedDataset {
+    #[inline]
+    pub fn serialize(&self) -> Result<Vec<u8>, rauthy_error::ErrorResponse> {
```
sebadob, line 38: We don't need the crates in advance when they are our own ones, so we can use `ErrorResponse` directly.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        bincode::serde::encode_to_vec(self, BINCODE).map_err(|err| {
```
sebadob, line 39: As mentioned above already, we should re-use the global serialization functions to make a future migration away from `bincode` easier. I have not looked into it yet, but `wincode` claims to be 100% compatible as drop-in replacement, but only for the legacy config I think.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+            })
+    }
+
+    /// Transform a raw MDS JWT (as downloaded, `header.payload.signature`) into a prepared dataset.
+    ///
+    /// The signature is not verified: the blob is fetched over TLS from the alliance and shipped as
+    /// trusted, pre-packaged data. Entries are kept only when they carry an AAGUID, have never been
+    /// revoked or reported compromised, and ship at least one root certificate.
+    #[inline]
+    pub fn from_jwt(jwt: &str) -> Result<Self, rauthy_error::ErrorResponse> {
```
sebadob, line 65: This could be an `impl FromStr for PreparedDataset` inside `raw`, which would make it a lot cleaner. You can then `.parse()` directly on the raw JWT data without cross-module calling here. Things like these are easier to review inside a browser for instance, when you don't have to jump between files too much.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+        raw::transform(jwt)
+    }
+}
+
+/// One authenticator, keyed by its 16-byte AAGUID.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct PreparedEntry {
```
sebadob, line 72: Would change the name to `MdsEntry` to make it clear outside of the context of this module, since it's `pub`. The same for most other `pub` structs and enums below. We usually want to avoid using the `module::*` when calling our own structs and enums to have more readable code.

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub aaguid: [u8; 16],
```
sebadob, line 73: We should be able to use `Uuid` here, right? That would be type safe. Or is there any reason not to?

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub description: String,
+    /// Bitmask of [`KeyProtection`].
+    pub key_protection: u32,
```
sebadob, line 76: For all the bitmasked values, I would use their type directly, like `KeyProtection`, to make it misuse resistant, or is there a reason I am missing right now why it's a bad idea?

src/data/src/fido_mds/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+    Unknown = 1 << 0,
+    Software = 1 << 1,
+    Hardware = 1 << 2,
+    Tee = 1 << 3,
+    SecureElement = 1 << 4,
+    RemoteHandle = 1 << 5,
+}
+
+impl KeyProtection {
+    fn from_mds(s: &str) -> Option<Self> {
```
sebadob, line 134: A `impl FromStr for KeyProtection` is probably more clear, so we can `.parse()` on any String value. The same for the others below. In addition, you can `impl From<u32>` for parse a bitmask into the proper variant.

src/data/src/fido_mds/raw.rs
```diff
@@ -0,0 +1,221 @@
...
+            None => {
+                mask |= unknown;
+                warn!("Unknown FIDO MDS {field} value {v:?} for {description:?}");
+            }
+        }
+    }
+    mask
+}
+
+fn parse_aaguid(s: &str) -> Result<[u8; 16], ErrorResponse> {
```
sebadob, line 114: We should be able to make use of `Uuid` here so we don't need to do anything manual, right? It even has `serde` impls out of the box.

src/data/src/fido_mds/raw.rs
```diff
@@ -0,0 +1,221 @@
...
+    })?;
+    bytes.try_into().map_err(|_| {
+        ErrorResponse::new(
+            ErrorResponseType::BadRequest,
+            format!("FIDO MDS AAGUID {s:?} is not 16 bytes"),
+        )
+    })
+}
+
+pub(super) fn transform(jwt: &str) -> Result<PreparedDataset, ErrorResponse> {
```
sebadob, line 129: As mentioned in the other comment already, changing this into an `impl FromStr for PreparedDataset` would be more clear without the need for any docs or guessing.

src/data/src/fido_mds/raw.rs
```diff
@@ -0,0 +1,221 @@
...
+        ErrorResponse::new(
+            ErrorResponseType::BadRequest,
+            format!("Error parsing FIDO MDS payload: {err}"),
+        )
+    })?;
+
+    // deduplicate certs across all entries, keyed by their DER hash; BTreeMap keeps the output
+    // deterministic so a regeneration from the same blob produces a byte-identical asset
+    let mut cert_map: BTreeMap<[u8; 32], Vec<u8>> = BTreeMap::new();
+    let mut entries = Vec::new();
```
sebadob, line 145: Not really critical, but we can initialize `with_capacity()`. We know how many entries we have at this point. The skipped ones will be not too many, and auto-growing a `Vec<_>` will over-allocate as well under the hood, so we don't really waste memory.

sebadob (changes requested):
`assets/fido_mds/dataset.bin` should not be added to the repo, but be downloaded during `just setup` and before `build` as well. This would also guarantee that it's impossible to inject a malicious file into a build pipeline.

Since it's basically doing nothing from the perspective of a user, I would probably not mention it in the changelog. I will most probably only release a new version with all necessary changes in place.

bartekus:
All of it is addressed, in two follow-up commits on top of the tip you read, plus a
merge of current `main`. No force-push, as promised.

## The dataset is no longer in the repo

`assets/fido_mds/dataset.bin` is gone from version control and gitignored. `just setup`
fetches it, and `just build` refetches it before a release image, so a release always
ships a current dataset rather than whatever was committed last.

One thing had to come with it. `fido-mds-prep` links `rauthy-data`, so a bare
`include_bytes!` on the missing asset is a bootstrap cycle: the tool that creates the
asset cannot be built without the asset. A fresh clone would not compile, and neither
would the Code Style workflow, which never runs `just setup`. So `src/data/build.rs`
copies the asset into `OUT_DIR`, and the allowance for a missing one is bounded to
exactly the case that needs it:

- missing or empty, debug profile: an empty placeholder and a `cargo::warning`. Seeding
  it is a no-op that warns at startup.
- missing or empty, release profile: a hard error naming `just fido-mds-prep`. A release
  binary that silently ships no dataset is the thing worth preventing, and the recipe
  builds the tool in the debug profile, so the cycle still has a way out.
- unreadable for any other reason: a hard error in every profile. Only `NotFound` means
  "not fetched yet"; a permissions problem is a broken tree, not a bootstrap.

Verified end to end in one build cache: a tree with no dataset builds `fido-mds-prep`,
the tool writes an asset, and the rebuilt server embeds and seeds that asset rather than
the placeholder. A malformed asset fails startup loudly rather than seeding nothing.

On the threat model I want to be precise, because I think the change is worth making but
not for quite the reason given. Moving the dataset out of the repo does defend against a
malicious commit to this repository, and that is a real gain. What it does not do is make
injection impossible: it replaces an artifact that appears in a diff and can be reviewed
with a build-time fetch whose integrity rests on TLS. TLS is not nothing here, and I do
not want to undersell it: the connection is authenticated against the platform trust
store, so a network attacker cannot substitute the payload in transit. But authenticating
the server is not the same as verifying the blob. The MDS blob is itself a signed JWT and
this PR does not check that signature, so nothing here would detect a payload served by a
compromised or misconfigured alliance host, and a `--source` pointing at a local file is
trusted as far as whoever put the file there.

Verifying the blob's signature would close that, and I think it is worth doing. I am not
proposing a trust anchor in this message, because the right anchor depends on what the
blob's own `x5c` chain actually terminates in and I was rate-limited out of inspecting a
live blob before writing this. I would rather come back with the chain in front of me
than guess at it. Either way it is its own PR, not this one.

What I did do is close the transport holes that were actually open. `http://` is refused,
any other scheme is refused, and the redirect policy refuses to follow a hop off `https`,
which the default client would have taken silently.

The changelog entry is gone.

## The tool

`clap` now, with its own `about`. While testing it I hit something worth flagging: the
metadata service rate-limits hard, and a release build that fetches the dataset would
simply fail on a `429`. So the download retries with backoff and checks that what came
back is JWT-shaped. If you would rather `just build` not reach the network at all, say so
and I will make it reuse an existing asset and fail only when there is none.

## Naming and types

`MdsDataset` / `MdsEntry` / `MdsCert` / `MdsCertLevel`, `MDS_DATASET`, and the raw
deserialization structs are `Raw*`. `aaguid` is a `Uuid` from `webauthn_rs::prelude`,
which is what the rest of the webauthn code already uses, so the manual hex parsing is
gone. The three bitmasks are newtypes over their flag enums reached through `contains()`
and `bits()`, so they cannot be swapped for each other or for a bare integer; a plain
`KeyProtection` would not work because the MDS field is a set, not one value.
`next_update` is `next_update_ts: i64`. `from_jwt()` is `impl FromStr for MdsDataset`,
and every MDS string value parses through `FromStr` on its enum.

`MdsCertLevel: FromStr` turned out to be worth more than tidiness: the max-across-reports
derivation collapses to `filter_map().max()`, because `Certified` already orders below
every level, so the separate "saw a bare FIDO_CERTIFIED" flag disappears. Same result,
and there is a test pinning the ordering.

Serialization goes through `rauthy_common::utils`, so the move off `bincode` is one place.

## SQL

`fido_mds_certs.der` is `cert_der`. The three inserts are raw strings with
`ON CONFLICT DO UPDATE` (the join table is all primary key, so `DO NOTHING`).
`insert()` and `seed_embedded()` are `impl MdsDataset` and the `db` module is private.
Import order and the comments that only restated the function name are fixed, and the
entries `Vec` is `with_capacity`.

The migrations had to move: `32_` and `V27__` were both claimed while this was open, so
they are `35_fido_mds.sql` and `V30__fido_mds.sql`. Neither of mine has shipped.

## Verification

- `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.
- Eight transform tests, including new ones pinning the cert-level ordering and the
  `Unknown` bit position, since both are persisted.
- `just test-hiqlite` and `just test-postgres`: 154 passed, 0 failed on each. Migration
  35 and V30 apply and the dataset seeds on both.
- Restarting on an existing database re-applies nothing and inserts nothing.
- The conflict clause is exercised for real on **both** engines, not only inferred from
  an empty-table seed: re-applying the whole dataset over the populated tables completes
  with no constraint error, leaves the counts at 340 / 245 / 716, and on Postgres a
  passkey planted against one of those entries keeps its `aaguid` and its join row across
  the upsert. That is the refresh behaviour we agreed on.
- On the real hiqlite-produced schema: the migration applies to a populated `passkeys`
  table leaving `aaguid` null with `PRAGMA foreign_key_check` clean, and deleting an entry
  nulls the referencing passkey and cascades its join rows while keeping the passkey row.

One thing I could not do: run the live download twice back to back to show byte-identical
output, because the service was rate-limiting by then. Determinism is covered by the
round-trip test and by the `BTreeMap` plus sort, not by a live double-run.

On `just pre-pr-checks`, so you know exactly what did and did not run here: `fmt`,
`is-clean`, `test-hiqlite` and `test-postgres` all ran, and so did
`clippy --all-targets -- -D warnings`. I did not run `build-ui` locally. My commits touch
no frontend source (the frontend files in the diff arrive through the merge of `main`),
so I did not think it earned a UI build, but I would rather say so than let CI be the
first place it comes up. The two database suites ran on the first of the two follow-up
commits; the second only changes the build script, and after it I did boot the server and
apply the migrations and seed on both engines as part of the checks above, so both
database paths were exercised at this tip even though the full suites were not re-run.

sebadob (approved):
Thanks

sebadob:
> Verifying the blob's signature would close that, and I think it is worth doing. I am not
proposing a trust anchor in this message, because the right anchor depends on what the
blob's own x5c chain actually terminates in and I was rate-limited out of inspecting a
live blob before writing this. I would rather come back with the chain in front of me
than guess at it. Either way it is its own PR, not this one.

I agree. It should really not be that much overhead to do anyway. The question then is just how we want to handle the pubkey. If we fetch this as well from the same source, we gain nothing at all. We would need to add it to the repo and only update it manually. But then I don't know how often they might change it and do key rotation. This has the potential to become annoying, but it's worth checking out.

---

If I am not mistaken, the only other DB migration we would need to complete the picture in the end is the new `aaguid` column on `passkeys`, right? I reworked them quite a bit already and we accept Resident Keys now. When the big Hiqlite PR is being merged, and we have a new release there, the next Rauthy release is just around the corner. I have one bigger local branch for Rauthy with all the `Duration`  config vars rework and renaming lots of them to make them uniform, and then that's probably almost all that makes it into the next release. If we can also push the missing parts for the DB, we could provide the new MDS feature with a patch later on, even when this is not fully finished until then.

sebadob:
Sorry, just saw the `aaguid` is already in the migrations, my bad. This means we could actually start requesting them already, even if we don't validate yet.

PR #1682, opened 2026-07-27
enable PKCE for confidential dynamic clients

alari:
DCR clients register confidential by default, but MCP / OAuth 2.1 clients still send a `code_challenge`, so `/authorize` rejects them with `'code_challenge' not enabled for this client`. Enable PKCE for every dynamic client, not only public ones.

### DIFF
--- a/src/data/src/entity/clients.rs
+++ b/src/data/src/entity/clients.rs
@@ -1394,6 +1394,20 @@ impl Client {
                 ))
             }
         } else if code_challenge.is_some() || code_challenge_method.is_some() {
+            // A dynamic client cannot say at registration time whether it is going to use
+            // PKCE, which is why confidential dynamic clients have no `challenge` configured.
+            // When such a client decides to send one anyway, it is accepted as an optional
+            // addon instead of being rejected: the challenge is saved with the auth code and
+            // the `code_verifier` is validated during the token exchange as usual. It is
+            // still not required - a confidential dynamic client without a challenge keeps
+            // working like before.
+            if self.is_dynamic()
+                && code_challenge.is_some()
+                && code_challenge_method.as_deref() == Some("S256")
+            {
+                return Ok(());
+            }
+
             trace!("'code_challenge' not enabled for this client");
             Err(ErrorResponse::new(
                 ErrorResponseType::BadRequest,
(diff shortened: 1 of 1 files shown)


sebadob (changes requested):
This is not going to happen in this way, as it would break lots of clients. Rauthy is very strict. If PKCE is enabled, it enforces it. This means a confidential client MUST always provide a PKCE challenge if configured, it is not optional. 

The only proper solution for this would be a new config variable that allows optional PKCE values for confidential DCR clients only. You cannot implement a fix that only works for one specific client. It must work for all of them. Usually, clients will not send a PKCE challenge when they are confidential. It will very slightly boost the security, but it's also kind of redundant. PKCE was created to make public client logins more safe. In terms of security, you don't really need it.

alari:
Redone as the config value you asked for: `dynamic_clients.confidential_pkce_optional` (env `DYN_CLIENT_CONFIDENTIAL_PKCE_OPTIONAL`), `false` by default, so nothing changes for anyone who does not opt in — confidential DCR clients keep getting no PKCE at all, exactly as today.

With it set, PKCE becomes optional for that one class and nothing else:

- registration gives a confidential dynamic client `challenge: S256`, so a `code_challenge` is no longer refused;
- `validate_code_challenge()` accepts a *missing* challenge only when the client is both confidential and dynamic and the value is set — every other client keeps the strict "enabled means required" behaviour;
- a challenge that is sent is verified exactly as before, and the token exchange keys off what the auth code actually carries, so omitting the challenge cannot downgrade a request that did send one. A confidential client still authenticates with its secret at the token endpoint either way.

Public dynamic, ephemeral and static clients are untouched. Documented in `config.toml` and the reference config; rebased onto current `main`, force-pushed. `cargo fmt`, `cargo clippy -- -D warnings` and the test suite are green here.

sebadob (changes requested):
Looks better now, but still has an issue: It introduces a breaking change.

The behavior when to require PKCE should not be changed. The default should stay like it is, since that's how clients usually work. This PR would make it necessary to change the default config to make most dynamic clients work while it should be the other way around.  You need to change line `1396` the `else if code_challenge.is_some() || code_challenge_method.is_some() {` branch depending on if it's a dynamic client.

PKCE for a confidential, dynamic client should be an optional addon if the client sends a challenge. It should not be required by default.

alari:
Redone without the config value. Registration is untouched — a confidential dynamic client still gets no `challenge` — and the only change is in the `else if` branch of `validate_code_challenge()` you pointed at: a `code_challenge` sent by a dynamic client is accepted as an optional addon instead of being rejected. It rides the existing machinery from there — saved with the auth code, `code_verifier` checked during the token exchange — and a dynamic client that sends nothing keeps working exactly as today, so defaults and every other client class are unchanged.

S256 only: `plain` from a confidential client adds nothing over not using PKCE at all, so it stays rejected.

Rebased onto current `main`.

sebadob (approved):
That made the PR a lot simpler. Thanks!

PR #1685, opened 2026-07-29
i18n: translate missing French strings ('authorize' + 'admin' claims)

cdewaele:
While diagnosing a mixed French/English page on my own Rauthy deployment (login -> "password forgotten"), I found that the entire `authorize` block in `frontend/src/i18n/common/fr.ts` (28 keys) had been left as an exact copy of `en.ts`, i.e. never translated.

A key-by-key diff against `en.ts` confirms this was the only gap in `common/fr.ts`:
- 236/236 keys present on both sides (no missing/extra keys).
- Every other identical FR/EN value is legitimate: cognates/technical terms (`ID`, `MFA`, `Type`, `Info`, `Minutes`, `Test`, `Pagination`, `Total`) or a URL (`fidoLink`).

This PR translates the `authorize` block, following the conventions already used elsewhere in `fr.ts` (lowercase `e-mail`, neutral/vouvoiement tone).

**Update:** fixes the CI formatting failure from the first commit, and follows up on @sebadob's comment that other untranslated keys likely exist elsewhere. The same key-by-key diff method applied to `admin/en.ts` vs `admin/fr.ts` found 5 more strings left as an exact English copy, all in the "claims at root" feature:
- `clients.claimsAtRoot`, `clients.claimsAtRootWarning`, `clients.claimsDesc`
- `scopes.claimsAtRoot`, `scopes.claimsAtRootWarning`

Same result: every other identical FR/EN value in `admin/fr.ts` is a legitimate cognate or technical term kept in English by existing convention in this file (`token`, `claims`, `JWT`, etc. — see e.g. `descAuthMethod`).

`npm run format-check` passes locally.


sebadob:
This is not a bug, it's intentional, as I don't speak any french. Auto-translations are usually a nightmare, which is why I copy&paste the english parts if I need to add new translations, and then some (native) speaker needs to update them. I have no idea if things like lowercase e-mail are the default in french. Someone else added these.

It's very likely that there are more keys that are in english.

sebadob:
Thanks

PR #1686, opened 2026-07-30
allow-list RFC 8707 resources for dynamic and ephemeral clients

alari:
MCP clients send an RFC 8707 `resource`, and a lot of them either self-register over DCR or come in as CIMD documents I don't control, so they can't carry `allowed_resources` themselves. On 0.36.1 that means `/authorize` always ends in `invalid_target` for them: dynamic clients get rejected outright, and a CIMD doc that lists no resources only gets through with `danger_allow_unvalidated_resource`.

This adds `allowed_resources` to `[dynamic_clients]` and `[ephemeral_clients]`, same shape as the `allowed_scopes` you already apply to these clients. It's seeded onto the client at registration; a CIMD doc that declares its own still wins. Empty by default, so nothing changes unless an operator sets it: dynamic clients stay rejected and ephemeral ones still need the danger flag. When it is set, a self-registered client can only request what's on the list.

The trusted party you noted a dynamic client lacks is the operator's config here, same as it already is for scopes and flows on these clients.

Also fixes `create_dynamic` inserting 22 columns vs `Client::create`'s 24, so `allowed_resources` and `default_aud` were silently dropped for dynamic clients regardless of this change.

Happy to split the insert fix out or adjust the shape if you'd rather have it another way.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1112,6 +1112,17 @@ enable = true
 # overwritten by: DYN_CLIENT_ALLOWED_SCOPES - single String, \n separated values
 allowed_scopes = ['openid', 'profile', 'email', 'groups']
 
+# RFC 8707 resource indicators a dynamically registered client may request. A
+# dynamic client cannot declare `allowed_resources` itself, so without this it can
+# never request a `resource`; an MCP client, which must send one, would otherwise
+# get `invalid_target`. Resolved from the live config on every request and never
+# stored with the client. Entries are matched verbatim. An empty list keeps the
+# default deny.
+#
+# default: []
+# overwritten by: DYN_CLIENT_ALLOWED_RESOURCES - single String, \n separated values
+#allowed_resources = []
+
 # The default scopes separated by ' ' for dynamic clients.
 #
 # default: ['openid']
@@ -1494,6 +1505,16 @@ cache_lifetime = 3600
 # default: false
 # overwritten by: EPHEMERAL_CLIENTS_DANGER_ALLOW_UNVALIDATED_RESOURCE
 #danger_allow_unvalidated_resource = false
+
+# RFC 8707 resource indicators allowed when an ephemeral client document declares
+# none of its own (a CIMD document you do not control cannot declare them). Keeps
+# the default deny without the blunt `danger_allow_unvalidated_resource`. A document
+# that declares its own resources still wins. Resolved from the live config on every
+# request. Entries are matched verbatim.
+#
+# default: []
+# overwritten by: EPHEMERAL_CLIENTS_ALLOWED_RESOURCES - single String, \n separated values
+#allowed_resources = []
 
 # A dynamic client registration is always rejected when it advertises a grant
 # type Rauthy does not support. Some spec-valid CIMD clients (e.g. claude.ai)
--- a/src/data/src/entity/clients.rs
+++ b/src/data/src/entity/clients.rs
@@ -960,8 +960,11 @@ impl Client {
 
     /// Validates an RFC 8707 `resource` request value against this client's policy: it
     /// must match one of the client's configured `allowed_resources`. The entries are
-    /// matched verbatim, so an operator decides what a valid value looks like. Ephemeral
-    /// clients without their own allow-list are only permitted when
+    /// matched verbatim, so an operator decides what a valid value looks like. Dynamic
+    /// clients, and ephemeral clients whose document declares no allow-list of its own,
+    /// are matched against `dynamic_clients.allowed_resources` /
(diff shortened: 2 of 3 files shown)


src/data/src/entity/clients.rs
```diff
@@ -250,9 +250,10 @@ $18, $19, $20, $21, $22, $23, $24)"#;
 INSERT INTO clients (id, name, enabled, confidential, secret, secret_kid, redirect_uris,
 post_logout_redirect_uris, allowed_origins, flows_enabled, access_token_alg, id_token_alg,
 auth_code_lifetime, access_token_lifetime, scopes, default_scopes, challenge, force_mfa,
-client_uri, contacts, backchannel_logout_uri, restrict_group_prefix)
+client_uri, contacts, backchannel_logout_uri, restrict_group_prefix, allowed_resources,
+default_aud)
```
sebadob: Dynamic clients MUST NOT have any allowed or default resources saved. This would make a priv esc and account takeover a no-brainer.

src/data/src/entity/clients.rs
```diff
@@ -1711,7 +1719,17 @@ impl From<EphemeralClientRequest> for Client {
             restrict_group_prefix: None,
             claims: None,
             claims_at_root: false,
-            allowed_resources: value.allowed_resources.map(|r| r.join(",")),
+            // A CIMD document may not list `allowed_resources`, and it may not be ours
+            // to edit. Fall back to the operator's ephemeral_clients.allowed_resources;
+            // a document that declares its own resources still takes precedence.
+            allowed_resources: value.allowed_resources.map(|r| r.join(",")).or_else(|| {
+                let r = RauthyConfig::get()
```
sebadob: This is a very naive approach, as changing the confrig would not update the DB. NEVER save any allowed resources for dynamic or ephemeral clients into the DB. They need to be resolved dynamically during authentication from the configured config at that moment in time.

A CIMD document MUST NOT list these. This would be an absolute security nightmare with probably a CVE of 10.

src/data/src/entity/clients.rs
```diff
@@ -1886,6 +1904,17 @@ impl Client {
             client_uri: req.client_uri,
             contacts: req.contacts.map(|c| c.join(",")).filter(|c| !c.is_empty()),
             backchannel_logout_uri: req.backchannel_logout_uri,
+            // A dynamically registered client cannot declare `allowed_resources`
+            // itself, so seed it from config. Empty by default, which leaves these
+            // clients unable to request a `resource` (see validate_resource_request).
+            allowed_resources: {
+                let r = RauthyConfig::get()
```
sebadob: The same as above - never do this.

config.toml
```diff
@@ -1112,6 +1112,16 @@ enable = true
 # overwritten by: DYN_CLIENT_ALLOWED_SCOPES - single String, \n separated values
 allowed_scopes = ['openid', 'profile', 'email', 'groups']
 
+# RFC 8707 resource indicators a dynamically registered client may request. A
+# dynamic client cannot declare `allowed_resources` itself, so without this it can
```
sebadob, line 1116: "Cannot" is the wrong term - it "must never".

config.toml
```diff
@@ -1112,6 +1112,16 @@ enable = true
...
+# never request a `resource`; an MCP client, which must send one, would otherwise
+# get `invalid_target`. Seeded onto every dynamic client at registration. Entries
+# are matched verbatim. An empty list keeps the default deny.
```
sebadob: We don't care of specific clients like "MCP" here. This only adds unnecessary text.

sebadob (changes requested):
A dynamic or ephemeral client MUST NEVER be able to define resources on their own. This would be a huge security vulnerability. It was 100% on purpose to deny this this. 

Allowing some very specific resources via config is the only option, but never save these values to the DB. Always resolve them dynamically from the currently existing config for those clients.

alari:
Redone the way you asked — nothing is written to the DB anymore. Registration is untouched; `validate_resource_request()` resolves the lists from the currently existing config on every request: a dynamic client is matched against `dynamic_clients.allowed_resources`, an ephemeral client whose document declares no `allowed_resources` of its own against `ephemeral_clients.allowed_resources` (a document that does declare its own still wins, and `danger_allow_unvalidated_resource` keeps working as before). Both lists are empty by default, so nothing changes until an operator sets them, and removing an entry takes effect on the next request.

The `create_dynamic` column change is gone too — with nothing persisted there was nothing left for it to fix.

Rebased onto current `main`.

sebadob (approved):
Thanks!

PR #1692, opened 2026-08-08
doc: add point to ensure /app/data is empty on disaster recovery

trefzer:
I did a disaster recovery which did recover, then tried to apply some WAL files which failed, so DB was recreated with initial state. So I assume it's save to cleanup the app/data directory before a recovery.

### DIFF
--- a/book/src/config/backup.md
+++ b/book/src/config/backup.md
@@ -84,13 +84,14 @@ The process is really simple:
 
 1. Have the cluster shut down. This is probably the case anyway, if you need to restore from a
    backup.
-2. Provide a backup file name on S3 storage with the `HQL_BACKUP_RESTORE` ENV value with prefix
+2. Ensure the directory /app/data is empty
+3. Provide a backup file name on S3 storage with the `HQL_BACKUP_RESTORE` ENV value with prefix
    `s3:object_name` (encrypted), or a file on disk (plain sqlite file) with the prefix
    `file:/path/to/backup`.
-3. Start up Rauthy
-4. Check the logs and wait for the backup to be finished
-5. After a successful restore, Rauthy will start its normal operation
-6. Make sure to remove the `HQL_BACKUP_RESTORE` env value.
+4. Start up Rauthy
+5. Check the logs and wait for the recovery to be finished
+6. After a successful restore, Rauthy will start its normal operation
+7. Make sure to remove the `HQL_BACKUP_RESTORE` env value.
 
 ```admonish danger 
 After a successful restore, you **MUST** remove the env var again! If you don't do it, Rauthy will 


sebadob:
> then tried to apply some WAL files which failed

Just to understand you correctly, you tried to apply WAL files manually, or what exactly do you mean? They should be cleaned up automatically, and if that does not happen, it's a logic bug in `hiqlite`.

trefzer:
Oh I was unclear:

after some experiment (going forward and backward with versions, killing the container etc) with a test instance I was in a state that I could not start the rauthy instance anymore because it either said that it could not find migration 28 (for version 35.x), or it said that the DB is too old (vor version 36.x). (Nothing to worry the instance that was heavily abused).

So I decided this is a good time to test now the disaster recovery.
I copied the backup file and set the environment to find the file. (but did not clean the volume containing the database).

Starting the container then:
- successfully restored the backup
- applied some things from the volume which failed
- mentioned something about an empty database
- runs all migrations
- created a new initial userid (which was indeed working)

I then stoped stopped the container, cleaned the volume. Which successfully restored the database as expected.

Sorry to not have saved the exact logs, but it's a test VM, that only keeps logs for an hour ;(

sebadob:
> (...) it either said that it could not find migration 28 

That is to be expected when you try to downgrade to a version that is missing DB migrations. E.g. when v0.36 updates the DB and applies migrations, you can only downgrade by applying a backup. This check makes sure that the DB can never be in an inconsistent state, or a different one the current code and logic is working for. It of course cannot detect manual DB changes.

> (...) or it said that the DB is too old

If that happened for the same DB that had at least v28 migration applied, I am not really sure how you got into that state. 

The new DB version is checked and possibly upserted 10 seconds after startup. This means if you start a new version that applies migrations, then you kill it within the first 10 seconds, and you try to downgrade, you will see the "missing migration" error even when the DB version is still set to the old one. However, that should not keep you from upgrading to the newer version that already applied the other DB migrations, because it should still be on the version that you came from originally.

Btw this explains why the instances produced an error in your other issue when you tried to upgrade. The first scheduler run for the app version check runs after 10 seconds for the first time.

> applied some things from the volume which failed

If that was about WAL logs, that would be a bug in `hiqlite`. It will check it over there. Usually, it should delete them after changing to a DB backup. However, if it was about WAL files, it should have `panic`ed as well because of an inconsistent state, so I am not sure what exactly failed in your case. Logs would have been very helpful for sure.
Anyway, I will have a look at this. It should not be necessary to manually clean up the data dir, but it of course does not hurt either to do it manually.

sebadob:
I just double checked the `hiqlite` code for restoring backups, and it cleans up WAL files as well:

```rust
debug!("Removing old data");
let _ = fs::remove_dir_all(&path_db).await;
let _ = fs::remove_dir_all(&path_snapshots).await;
let _ = fs::remove_dir_all(&path_lock_file).await;
let _ = fs::remove_dir_all(&path_logs).await;
```

This makes me wonder what your failed "applied something" was. Making sure the `data` dir is empty is not necessary. It cleans up all the files automatically, which should make it unnecessary to add this step to the docs. Rauthy logs would be helpful if you can ever reproduce this issue. Backup restore is even part of the Hiqlite integration tests, so it would be very weird if this fails.

Edit:

It actually makes sense to clean up the dir with an on-disk cache, as I am not sure if Rauthy catches such a case and does this automatically. Currently, the cache files are not cleaned up automatically during backup restore. It would probably make sense to add this to the `hiqlite` code as well.

PR #1696, opened 2026-08-12
fix: Security hardening for OIDC, sessions, clients, users, auth, encryption and outbound fetches

reneleonhardt:
## Summary

Consolidated security hardening for the OIDC/OAuth flows, session and logout lifecycle, client registration, secret handling and outbound fetches. Reviewed against the codebase with a dependency-aware map; every change is scoped and small. Maximized cheap black-box and white-box unit tests cover the important changed production paths; fully verified live end-to-end coverage is outside this requested change. Includes the fixes from the first review round (grace-window refresh claims, uniform password-grant error, end-only wildcard matching, per-IP credential-stuffing set, distributed counters, fail-loud encryption migration).

## Changes (logical concerns)

1. Atomic one-time claims — use hiqlite's atomic `get_remove()` for auth-code, device-code and PoW claims; the temporary `dlock` dependency and its round trips are removed.
2. Token issuance, refresh and logout hardening:
- the password grant answers credential failures (unknown, disabled, expired, password-expired user, wrong password) with one identical `Unauthorized` inside the grant handler, while preserving the password-reset flow and passing through non-credential errors;
- issued-token cleanup only removes rows at least 1 s past expiry (the old 60 s early-deletion margin made userinfo / introspection / token-exchange reject still-valid access tokens);
- refresh tokens (regular + device) are claimed atomically: the row `exp` moves to `now + refresh_token_grace_time`, so concurrent refreshes within the grace window still succeed, while a replay after it revokes the whole token family (RFC 6819) — device refresh tokens are only revoked when `token_revoke_device_tokens` is enabled.
3. Session lifecycle and force-logout races:
- `Init`-state sessions are only valid on the explicit login endpoints, even when `session_validate_ip` is disabled (the exception list was silently bypassed);
- the middleware last-seen refresh is update-only (`touch_last_seen`), so an in-flight request can no longer resurrect a force-logged-out session or clobber a concurrent login;
- `Session::upsert()` refuses expired sessions, so no promotion path (`finish_authorize`, auth-code grant, forward-auth callback, webauthn finish) can resurrect an expired session — the per-site checks were removed.
4. Redirect and logout URI safety:
- dynamic registration: `redirect_uris` must be absolute https (http loopback allowed, RFC 8252), wildcards rejected (a registered `https://*` was an open redirect / code leak), `backchannel_logout_uri` removed from the request entirely (a server-side fetch from an unauthenticated registration would be SSRF);
- wildcard redirect matching requires a host/path boundary (`https://app.example.com*` no longer matches `https://app.example.com.evil.com`), the wildcard must be the last character (`strip_suffix`, cheap `ends_with('*')` gate first) and a base ending at a boundary accepts any continuation — in redirect, post-logout and logout id-token-hint validation (the latter also had its check inverted).
5. Registration and password-reset redirects — `redirect_uri` is allow-listed against the configured client URIs (post-reset Location header was attacker-controllable; the registration check was inverted); the match is inline and tolerates a configured base with a trailing `/`.
6. Brute-force protection and throttling:
- credential-stuffing detection keeps a per-IP set of attempt hashes with a scan-window TTL (the per-attempt-key + full-cache-snapshot counting was reverted per review);
- the failed-login counter uses hiqlite's distributed counters and is reset only by a successful login; escalation persists across blacklist windows.
7. Constant-time comparisons — CSRF tokens, API keys, magic-link binding/CSRF, forward-auth CSRF, upstream-provider callback XSRF/PKCE/state, PAM remote password and dynamic-client registration tokens (consistent with the existing client-secret check).
8. Atomic encryption migration — client/API-key secret migration never silently clobbers a concurrent change: a failed conditional update is re-read and retried; if the race persists, the migration fails loudly. Legacy NULL key-id rows still migrate.
9. Atproto SSRF hardening — user-supplied identifiers are validated before server-side resolution: literal loopback / private / link-local addresses and `localhost` are rejected (including ports, brackets, IPv4-mapped IPv6 and `169.254.169.254`).

## Verification

- `cargo fmt --all -- --check` — clean
- `cargo clippy -- -D warnings` (exact CI form) — 0 warnings
- `cargo test --workspace --lib` — 87 passed / 0 failed; `--all-features --lib` — 87 passed / 0 failed
- `cargo test --workspace --all-targets --no-run` — all workspace targets compile; `src/bin/tests` require a live instance (see Reproduce)

## Reproduce

Bin integration tests (`src/bin/tests`) run against a live instance on port 8081: a fresh debug build (release rejects the tests' empty User-Agent), `ADMIN_FORCE_MFA=false` (the bootstrap admin has no WebAuthn), bootstrap `init_admin@localhost`/`123SuperSafe` and `init_client` with the secret hardcoded in the tests. Previously verified: `test_api_keys` passes end-to-end and each test passes in isolation. The full-suite login-delay escalation from tests that intentionally fail logins remains a documented limitation, reproduced on main and not a regression of this branch.

## Out of scope / decided

- Hostname-based SSRF in outbound fetches (static-client backchannel logout, provider JWKS): closed for dynamic registration by removing the field; a fetch-time safe layer (resolve-verify-pin) remains a possible follow-up.
- Backchannel logout tokens are not single-use: they are stateless, exp-limited and replaying one only re-triggers an idempotent logout, so a jti blacklist adds a rejection path without security value (per review).
- webauthn / MFA-mod / magic-link claim windows: left as-is (idempotent or capability-equivalent).
- CI integration tests under `src/bin/tests` require a live instance and are not part of the workspace lib suite.


src/api/src/users.rs
```diff
@@ -524,6 +512,40 @@ pub async fn post_users_register_handle(
         .finish())
 }
 
+/// Validates a registration / password-reset `redirect_uri` against the configured client
+/// URIs. The given URI must start with a configured client URI, followed by a path / query /
+/// fragment boundary (or be an exact match). The boundary check prevents host confusion such
+/// as `https://example.com.evil.com` passing for a configured `https://example.com`.
+#[inline]
+async fn validate_redirect_uri_allowed(redirect_uri: &str) -> Result<(), ErrorResponse> {
```
sebadob: The function name is a bit unclear / misleading without reading the comment. Change it to something like `validate_reg_redirect_uri`. You can read the comment with an active LSP, but not when reviewing inside a browser. Usually, a `recirect_uri` is always about the one during the auth flow.

src/api/src/users.rs
```diff
@@ -524,6 +512,40 @@ pub async fn post_users_register_handle(
...
+        ErrorResponseType::BadRequest,
+        "given `redirect_uri` not allowed",
+    ))
+}
+
+/// `true` if `given` starts with `base` and continues at a path / query / fragment boundary
+/// (or matches exactly). A bare prefix would let `https://example.com.evil.com` pass for a
+/// configured `https://example.com` — host confusion.
+#[inline]
+fn redirect_uri_matches_base(base: &str, given: &str) -> bool {
```
sebadob: This validation fails if the given base already has a tailing `/`, even though such a prefix is safe. I also don't see a reason to extract it into a separate function, and the name is misleading. It does not "match", it only validates the prefix. I would really keep the logic inside the function above, as it does not make much sense on its own.

src/data/src/entity/api_keys.rs
```diff
@@ -415,7 +443,8 @@ impl ApiKey {
             ));
         }
 
-        if self.secret.as_slice() == sha256!(secret.as_bytes()) {
+        // constant-time comparison to avoid leaking the secret via timing
```
sebadob: No need for this comment. The function name says everything already. Most of the constant time comparisons are actually just sugar coating. These comparisons are insanly fast (usually done in nanoseconds) and it's practically impossible to do any timing attacks because of the way slower network requests and even async context changes. I only added these after they were flagged in the security review, but it was impossible to measure any differences.

src/data/src/entity/clients.rs
```diff
@@ -1522,6 +1545,77 @@ impl Client {
...
+
+/// Wildcard redirect-URI match with a host/path boundary.
+///
+/// A registered `https://app.example.com/callback*` matches any URI starting with
+/// `https://app.example.com/callback` (path boundary inherent). A bare-host wildcard such as
+/// `https://app.example.com*` must NOT match `https://app.example.com.evil.com`: the
+/// requested URI must continue at a path / query / fragment boundary (or match exactly),
+/// which keeps the host confusion class closed for static clients as well.
+pub fn wildcard_prefix_match(registered: &str, requested: &str) -> bool {
+    let Some((prefix, _)) = registered.split_once('*') else {
```
sebadob: This rework drops the `uri.ends_with('*')` validation.

Also, could be made faster by using `strip_suffix()`. The `split_once()` iterates over the whole data from the beginning, while strip suffix starts at the end, which is the only place where a wildcard is allowed anyway.

Also, suffix strip adds in the above-dropped validation again.

src/data/src/entity/clients.rs
```diff
@@ -1522,6 +1545,77 @@ impl Client {
...
+        return false;
+    };
+    match requested.strip_prefix(prefix) {
```
sebadob, line 1639: The same issue as in the other redirect function - it does not work when the base has a trailing `/`.

src/data/src/entity/clients.rs
```diff
@@ -1316,7 +1340,7 @@ impl Client {
         let loopback = RauthyConfig::get().vars.access.rfc_8252_enable
             && (self.is_dynamic() || self.is_ephemeral());
         let has_any = self.get_redirect_uris().iter().any(|uri| {
-            (uri.ends_with('*') && redirect_uri.starts_with(uri.split_once('*').unwrap().0))
+            wildcard_prefix_match(uri, redirect_uri)
```
sebadob: This degrades performance. The `uri.ends_with('*')` is very cheap, the whole wildcard check is not. This is also completely dropped inside the new `wildcard_prefix_match()` function, which even degrade security.

The same issue in line 1372.

src/data/src/entity/clients.rs
```diff
@@ -1859,14 +1953,35 @@ impl Client {
...
+        //   keep the capability. See fork-todo.md decision 1.
+        validate_no_backchannel_logout_uri(&req.backchannel_logout_uri)?;
+
+        let mut redirect_uris = Vec::with_capacity(req.redirect_uris.len());
+        for uri in &req.redirect_uris {
+            validate_dyn_redirect_uri(uri)?;
+            redirect_uris.push(uri.clone());
+        }
+
+        let backchannel_logout_uri = None;
```
sebadob: When forcing it to `None` anyway, it would be way cleaner to just remove it completely from the request payload. Then you can also remove the whole checking function above.

src/data/src/entity/clients.rs
```diff
@@ -1885,7 +2000,7 @@ impl Client {
             force_mfa: false,
             client_uri: req.client_uri,
             contacts: req.contacts.map(|c| c.join(",")).filter(|c| !c.is_empty()),
-            backchannel_logout_uri: req.backchannel_logout_uri,
+            backchannel_logout_uri,
```
sebadob: Makes no sense to even include it when it's forced to `None` anyway. Simply remove that like and let `..Default::default()` take care of it.

src/data/src/entity/clients.rs
```diff
@@ -2012,6 +2164,44 @@ mod tests {
         assert!(only_unknown.is_empty());
     }
 
+    #[test]
+    #[test]
+    fn test_wildcard_prefix_match() {
```
sebadob, line 2236: Does not include a single test that ensures that a wildcard must only ever appear at the very end.

src/data/src/entity/clients_dyn.rs
```diff
@@ -113,7 +113,11 @@ impl ClientDyn {
     }
 
     pub fn validate_token(&self, bearer: &str) -> Result<(), ErrorResponse> {
-        if self.registration_token_plain()? != bearer {
+        // constant-time comparison to avoid leaking the registration token via timing
```
sebadob: Unnecessary comment, the function name says everything already.

src/data/src/entity/cred_stuff_detect.rs
```diff
@@ -24,31 +24,39 @@ impl CredStuffDetect {
 
     #[inline]
     async fn handle_trigger(ip: IpAddr, email: &str, password: &str) -> Result<(), ErrorResponse> {
+        // Each distinct (email, password) pair is stored under its own cache key with a
+        // scan-window TTL. This makes the counting race-free (previously a shared BTreeSet
+        // was read-modify-written, losing updates under parallel requests from the same IP,
+        // which let the blacklist threshold be bypassed), while keeping the de-duplication
```
sebadob: We don't care about what code did in the past inside a code comment. This only ever makes sense in the context of this PR.

src/data/src/entity/cred_stuff_detect.rs
```diff
@@ -24,31 +24,39 @@ impl CredStuffDetect {
...
-                ip,
-                &hashes,
+                key,
+                &(),
                 Some(vars.scan_window as i64),
             )
             .await?;
 
+        // Count all distinct attempts for this IP within the scan window.
+        let snapshot = client.get_snapshot::<_, ()>(Cache::CredStuffDetect).await?;
```
sebadob: We are not going to do this. The `get_snapshot` is insanely expensive, which is why the design beforehand was way superior, even though there was a slight change for a race-condition. Such a race-condition and maybe some dropped requests are not a big deal at all. Cred stuffing will lead to a huge amount of requests and will be caught anyway. Copying the whole cache snapshot for each request is absolutely insane. 

The PoW requirement and max hash threads limitations reduce such a chance even way further, apart from the cache lookups being very, very fast. In addition, they are being answered from a single KV handler thread. I guess it's very hard to even construct such a race condition programmatically, even without all the network latency in between.

If your really never wanted to have the theoretical possibility that a request might slip through, this needs a different design, like e.g. using distributed counters or something like that.

src/data/src/entity/failed_login_counter.rs
```diff
@@ -4,16 +4,61 @@ use rauthy_error::ErrorResponse;
...
+            .await?;
+        let count = Self::next_count(current);
+        DB::hql()
+            .put(
+                Cache::App,
+                Self::cache_idx(&ip),
+                &count,
+                Some(Self::TTL_SECS),
+            )
+            .await?;
```
sebadob: I don't see any good reason to drop distributed counters here, which are the perfect use-case for something like "a counter". You "fixed" a design like this above for the cred stuff detection while introducing it here instead of an atomic operation. 

We don't need a TTL here, it's even degrading security and throttling. It will lead to incorrect warnings about failed logins from a given IP. The counter is reset when there was a successful login, so no need to have a TTL.

src/data/src/entity/issued_tokens.rs
```diff
@@ -21,12 +20,15 @@ pub struct IssuedToken {
 impl IssuedToken {
     pub async fn cleanup_expired() -> Result<(), ErrorResponse> {
         let sql = "DELETE FROM issued_tokens WHERE exp < $1";
-        let now_plus_1 = Utc::now().add(chrono::Duration::minutes(1)).timestamp();
+        // Do NOT delete rows early: `validate_not_revoked` rejects tokens whose row is gone,
+        // so deleting anything with `exp < now + X` would make still-valid access tokens fail
+        // on userinfo / introspection / token exchange for up to X seconds before expiry.
+        let now = Utc::now().timestamp();
```
sebadob: The fix here would be to change to `now_minus_1` instead. The only reason why the 1 minute was added was to fix clock skew issues. It should only be minus instead to make sense.

src/data/src/entity/pow.rs
```diff
@@ -28,10 +28,14 @@ impl PowEntity {
...
     }
 
-    /// Checks re-usages of PoWs and prevents a future re-use
+    /// Checks re-usages of PoWs and prevents a future re-use.
+    /// The get+delete is serialized per challenge via a distributed lock, so two concurrent
+    /// requests cannot both consume the same challenge.
     pub async fn check_prevent_reuse(challenge: String) -> Result<(), ErrorResponse> {
         let client = DB::hql();
 
+        let _lock = client.lock(format!("pow_{challenge}")).await?;
```
sebadob: A way better solution for this would be to add a `get_remove` function to hiqlite, which returns the removed data. This would be way quicker, require less resources and network roundtrips.

src/data/src/entity/refresh_tokens.rs
```diff
@@ -168,6 +168,23 @@ impl RefreshToken {
         Ok(slf)
     }
 
+    /// Atomically claims this refresh token for one-time use. Returns `true` if the claim
+    /// succeeded (the token was still valid and this caller won), `false` if it was already
+    /// claimed or expired. The conditional UPDATE (`exp > now`) makes the claim atomic, so
+    /// two concurrent refreshes with the same token cannot both succeed, and a stolen token
+    /// cannot be replayed after legitimate use.
```
sebadob: You are commenting function-internal logic here which does not make sense for the function itself. The only thing that a caller needs to know is that it's atomic. There is no need to explain why this is the case. If you wanted to explain it, a comment for the query inside makes more sense.

I would also either add `#[must_use]` in such a case, or not even return a `bool` at all, and error with NotFound or Forbidden directly, which would be misuse resistant.

src/data/src/entity/refresh_tokens_devices.rs
```diff
@@ -140,6 +140,19 @@ impl RefreshTokenDevice {
         Ok(slf)
     }
 
+    /// Atomically claims this device refresh token for one-time use. See `RefreshToken::claim`.
+    pub async fn claim(id: &str, now: i64) -> Result<bool, ErrorResponse> {
```
sebadob: The same as above. Don't return a `bool`, but an Error directly, if it failed, to make it misuse resistant.

src/data/src/entity/sessions.rs
```diff
@@ -741,7 +771,8 @@ impl Session {
                 "CSRF Token not present in HTTP Header",
             ));
         }
-        if self.csrf_token.eq(csrf?) {
+        // constant-time comparison to avoid leaking the token via timing
```
sebadob: No need for that comment.

src/middlewares/src/principal.rs
```diff
@@ -139,7 +139,9 @@ async fn get_session_from_cookie(req: &ServiceRequest) -> Result<Option<Session>
                 // only update the last_seen, if it is older than 10 seconds
                 if session.last_seen < now - 10 {
                     session.last_seen = now;
-                    session.upsert().await?;
+                    // Update-only on purpose: must not resurrect a session that was
+                    // concurrently force-logged-out (invalidate_all / invalidate_for_user).
```
sebadob: This comment can be removed. Only makes sense in the context of this PR, and the `touch_last_seen()` already has the same explanation.

src/service/src/oidc/grant_types/authorization_code.rs
```diff
@@ -95,8 +96,12 @@ pub async fn grant_type_authorization_code(
         ));
     }
 
-    // get the oidc code from the cache
+    // get the oidc code from the cache. A distributed lock serializes redemption of the
+    // SAME code, making single-use strict: two concurrent requests with one code cannot both
+    // pass (the cache get/delete alone is not atomic). The lock is held only for the
+    // find-validate-delete section below, before any token is issued.
     let idx = req_data.code.as_ref().unwrap().to_owned();
+    let _lock = DB::hql().lock(format!("auth_code_{idx}")).await?;
```
sebadob: Same as above already. Adding a `get_remove()` function to hiqlite makes way more sense and is superior in every way.

src/service/src/oidc/grant_types/device_code.rs
```diff
@@ -25,6 +26,22 @@ pub async fn grant_type_device_code(peer_ip: IpAddr, payload: TokenRequest) -> H
         }
         Some(dc) => dc,
     };
+
+    // Serialize polling / verification of the SAME device code via a distributed lock: the
+    // find-check-delete of the device code is not atomic on its own, so two concurrent polls
+    // could both pass the `verified_by` check and each issue a full TokenSet. Holding the
+    // lock for the whole poll makes the grant strictly single-use.
+    let _lock = match DB::hql().lock(format!("device_code_{device_code}")).await {
```
sebadob: Another candidate for `get_remove()`.

src/service/src/oidc/grant_types/password.rs
```diff
@@ -81,10 +81,23 @@ pub async fn grant_type_password(
     }
 
     // This Error must be the same if user does not exist AND passwords do not match to prevent
-    // username enumeration
-    let mut user = User::find_by_email(String::from(email)).await?;
-    user.check_enabled()?;
-    user.check_expired()?;
+    // username enumeration. Returning the original NotFound / Forbidden errors here would leak
+    // whether an account exists and whether it is disabled or expired.
+    let mut user = match User::find_by_email(String::from(email)).await {
```
sebadob: Apart from code repitition, this is the wrong place to do it. It not only makes the code less readable, but leaves out other potential ones, and it also very brittle. A single typo can leak internals. 

There is a `TODO` in `src/api/src/oidc.rs` -> `post_token()` for that.

src/service/src/oidc/authorize.rs
```diff
@@ -219,6 +219,17 @@ pub(crate) async fn finish_authorize(
     user_needs_mfa: Option<&mut bool>,
     provider_mfa_login: Option<ProviderMfaLogin>,
 ) -> Result<AuthStep, ErrorResponse> {
+    // Guard against resurrecting a session that was concurrently invalidated (global logout
+    // via `Session::invalidate_all`, user logout / password reset via `invalidate_for_user`).
+    // Without this, `session.set_authenticated` / `upsert` below would restore the original
+    // `exp` and keep the session alive after a force logout.
+    if session.exp <= Utc::now().timestamp() {
```
sebadob: This is the wrong place to do this check. It would also not fix any race condition, as it could theoretically still happen between this line and the `set_authenticated()`. This check only makes sense directly inside `session.set_authenticated()`, which will also automatically have make correct in all situations.

src/service/src/oidc/bcl_logout_token.rs
```diff
@@ -152,6 +153,26 @@ impl LogoutToken<'_> {
         let mut buf = Vec::with_capacity(256);
         jwk.validate_token_signature(logout_token, &mut buf)?;
 
+        // Replay protection: an intercepted backchannel logout token must not be usable twice
+        // (a replay would re-trigger session termination / backchannel fan-out). Cache the `jti`
+        // for the token's validity window + clock skew; a second presentation is rejected.
+        let ttl = RauthyConfig::get().vars.backchannel_logout.token_lifetime as i64
```
sebadob: Some issue with this one:

1. A token cannot be intercepted. This attack vector is a left-over from the very old days when tokens were used over plain connections. Rauthy does not even allow this in production.
2. Tokens are stateless, which is why they have an `exp` claim. They are not sessions. If we would need replay protection, they would not need an `exp` claim, since they were one-time use anyway. So yes, we can blacklist the `jti`, but it does not improve security.
3. Triggering another backchannel logout would actually increase the security, and make sure twice, that someone is actually safely logged out.
4. The `App` cache is logically completely wrong. If you really wanted to add this, add a `Cache` variant for it.
5. Since your other changes are about atomicity so much, and it makes sense to add a `get_remove()` to `hiqlite` anyway for other situations, I would also add a `replace()`, which in addition to `put()`, returns `Option<_>` with the old value in case a key existed before. This would make it actually "safe" and a lot cleaner.

src/service/src/oidc/logout.rs
```diff
@@ -70,16 +70,13 @@ pub async fn get_logout_html(
...
+        // configured URI (exact or wildcard with host/path boundary), otherwise reject.
         let valid_redirect = uri_vec.as_ref().unwrap().iter().any(|uri| {
-            if uri.ends_with('*') && target.starts_with(uri.split_once('*').unwrap().0) {
-                return true;
-            }
-            if target.eq(uri) {
-                return true;
-            }
-            false
+            rauthy_data::entity::clients::wildcard_prefix_match(uri, &target) || target.eq(uri)
```
sebadob: Here, again, you need to add the check to the extracted function to make sure it actually ends with a `*`. You can also drop the crate and module for all Rauthy-owned ones, and use the function directly. At most keep `clients::*` for a tiny reference.

src/service/src/oidc/validation.rs
```diff
@@ -60,6 +62,17 @@ pub async fn validate_auth_req_param(
     Ok((client, header))
 }
 
+/// Revokes the whole refresh-token family for a user after a detected replay: all refresh
+/// tokens (regular + device), all issued tokens and all sessions. RFC 6819 §5.2.1.3: on
+/// token reuse the authorization server SHOULD revoke the grant family.
+async fn revoke_refresh_family(user_id: &str) -> Result<(), ErrorResponse> {
+    RefreshToken::invalidate_for_user(user_id).await?;
+    RefreshTokenDevice::invalidate_for_user(user_id).await?;
```
sebadob: We usually don't want to revoke device tokens. They live on devices with limited capabilities, and authenticating them can be a huge pain, like for headless IoT devices. This should depend on `token_revoke_device_tokens` config var.

src/service/src/oidc/validation.rs
```diff
@@ -128,11 +141,14 @@ pub async fn validate_and_refresh_token(
...
     client.validate_user_groups(&user)?;
 
-    // validate that it exists in the db and invalidate it afterward
+    // validate that it exists in the db and atomically claim it for one-time use.
+    // The old token is expired by the claim (rotation), so a stolen token cannot be replayed
+    // after legitimate use and two concurrent refreshes with the same token cannot both win.
+    // Note: this supersedes the previous grace-period extension
+    // (`refresh_token_grace_time` no longer keeps the old token usable after a refresh).
     let now = Utc::now().timestamp();
-    let exp_at_secs = now + RauthyConfig::get().vars.lifetimes.refresh_token_grace_time as i64;
```
sebadob, line 133: There is no reason to drop the `refresh_token_grace_time`.

src/service/src/encryption.rs
```diff
@@ -32,15 +32,31 @@ pub async fn migrate_encryption_alg(new_kid: &str) -> Result<(), ErrorResponse>
...
+                client.secret.clone().unwrap(),
+                new_kid.to_string(),
+                &old_kid,
+            )
+            .await?
+        {
+            modified += 1;
+        } else {
+            error!(
+                "Skipped client '{}' during secret migration - secret was changed concurrently",
```
sebadob: That must never happen. Especially not silently with a simple log. Either error loudly, or have built-in automatic retry. When the migration succeeds, all encrypted DB entries MUST have been migrated.

src/service/src/encryption.rs
```diff
@@ -55,23 +71,32 @@ pub async fn migrate_encryption_alg(new_kid: &str) -> Result<(), ErrorResponse>
...
+        // atomic conditional update, see clients above
+        let old_kid = api_key.enc_key_id.clone();
+        if api_key
+            .save_migrated(secret_enc, access_enc, new_kid.to_string(), &old_kid)
+            .await?
+        {
+            modified += 1;
+        } else {
+            error!(
+                "Skipped API key '{}' during secret migration - secret was changed concurrently",
```
sebadob: The same as above. Never error silently and ignore in such a case.

src/service/src/forward_auth.rs
```diff
@@ -267,6 +267,17 @@ pub async fn get_forward_auth_client_callback(
 
     // update session metadata
     let mut session = Session::find(sid).await?;
+    // Guard against resurrecting a session that was concurrently invalidated (global logout
+    // via `Session::invalidate_all`, user logout / password reset via `invalidate_for_user`):
+    // the auth code may still be valid while its session has already been killed. Without
+    // this check, the `upsert` below would restore the original `exp` and keep the session
+    // alive after a force logout.
+    if session.exp <= Utc::now().timestamp() {
```
sebadob: Doing these checks in many places is cumbersome and error prone. I would add an `exp` guard to the query for `Session::find()` to have this automatic. Or even completely safe, actually make use of the new `touch_last_seen()`. In this case, there are more updates, to an `exp` guard could also be added to `upsert()`, but only after making 100% sure that the upsert is never used for any invalidation, which it should not. There are dedicated invalidation functions.

src/service/src/login_delay.rs
```diff
@@ -120,6 +120,12 @@ async fn build_send_event(
         .unwrap();
     IpBlacklist::put(peer_ip.to_string(), nbf_seconds as i64).await?;
 
+    // Reset the failure counter when the IP gets blacklisted: the blacklist TTL then acts as
+    // the decay window. Without this, a shared / NAT'd IP that suffered a brute-force burst
+    // would stay permanently throttled - once blacklisted, even a legit user behind the same
+    // IP can never log in successfully to reset the counter, keeping the IP locked forever.
+    FailedLoginCounter::reset(peer_ip.to_string()).await?;
```
sebadob: The failed login counter does not limit legitimate users. It's only applied after a failure. It's even reset after a successful login from an IP.

Cargo.toml
```diff
@@ -107,7 +107,7 @@ gethostname = "1"
 hex = { version = "0.4", features = ["serde"] }
 hickory-resolver = "0.26.1"
 hiqlite = { version = "0.14.0", features = [
-    "cache", "cast_ints", "counters", "dashboard", "listen_notify_local", "macros"
+    "cache", "cast_ints", "counters", "dashboard", "dlock", "listen_notify_local", "macros"
```
sebadob: I really want to avoid using `dlock`. It requires quite a few round trips and we can solve basically each situation without it.

We only need `get_remove()` and `replace()` functions for the cache to get cleaner code, no dlock logic, fewer network roundtrips and faster code.

sebadob (changes requested):
This PR was actually quite a pain to review because there were so many, non-trivial changes. Please split into smaller parts for future ones.

src/data/src/entity/pow.rs
reneleonhardt: Agreed. `get_remove()` and `replace()` are implemented in hiqlite's cache API; I'll switch the PoW removal over once the hiqlite release is out.

src/service/src/oidc/grant_types/authorization_code.rs
reneleonhardt: Same — the authorization code grant will use hiqlite's `get_remove()` once it's released.

src/service/src/oidc/grant_types/device_code.rs
reneleonhardt: Same for the device code grant.

Cargo.toml
reneleonhardt: Understood. The dlock use is a stopgap until `get_remove()`/`replace()` land in hiqlite — they are implemented; I will drop the dlock round trips as soon as a release provides them.

reneleonhardt:
What changed beyond the 32 comments:

- Reworked the atproto identifier resolution (loopback/private/link-local rejection incl. bracketed IPv6, explicit ports, IPv4-mapped literals, canonicalization) — stacked as its own commit, the originally reviewed atproto commit stays untouched.

Cargo.toml
reneleonhardt: Update: `get_remove()` / `replace()` are published in sebadob/hiqlite#362 — the dlock round trips can be dropped once it is released.

src/data/src/entity/clients.rs
reneleonhardt: Fixed in the follow-up `870eb00`: a registered base ending at a boundary (`/`, `?`, `#`) now counts as already bounded, so `https://app.example.com/app/*` matches `https://app.example.com/app/foo`. Regression cases for exactly this were added in `24393de`, including the intentional non-match of the bare `https://app.example.com/app`.

src/data/src/entity/clients.rs
reneleonhardt: Covered since `870eb00`: `strip_suffix('*')` plus the `ends_with('*')` gate enforce the end-only position, and the tests assert a mid-URI `*` (`https://example.com/*/cb`) never matches. `24393de` extends the suite with the trailing-slash-base cases from the comment above.

src/service/src/oidc/validation.rs
reneleonhardt: Not dropped — `41254dd` folded the claim into the row update and kept the config value as the window: `exp = now + refresh_token_grace_time`, now atomic and with family revocation on replay.

src/service/src/login_delay.rs
reneleonhardt: Agreed — the blacklist-time reset is removed in `d16bc81`: the counter now keeps escalating across blacklist windows (7 → 10 → 15 → 20 → 25), so a persistent attacker cannot restart the ladder after each block. The reset on a successful login stays as the decay for legitimate users (shared NAT), and the account-level `CredStuffDetect` covers the valid-credential stuffing vector.

src/service/src/forward_auth.rs
```diff
@@ -268,8 +268,7 @@ pub async fn get_forward_auth_client_callback(
     // update session metadata
     let mut session = Session::find(sid).await?;
     if session.state != SessionState::Auth {
-        // A Session is only set to `SessionState::Auth` AFTER a successful and complete
-        // auth code flow. Because of this, it is possible to get here with an `init` session.
+        // A Session is only set to `SessionState::Auth` AFTER a successful and complete auth code flow. Because of this, it is ...
```
sebadob: This comment change looks weird.

scripts/bin-tests/setup.sh
```diff
@@ -0,0 +1,105 @@
...
+# - The bootstrap admin must be `init_admin@localhost` / `123SuperSafe` and the
+#   `init_client` client must pre-exist with the EXACT secret the tests hardcode
+#   (clients.json supports {"Plain": "..."}).
+# - Port 8081 (hardcoded in tests/common.rs get_backend_url()).
+#
+# Known limitation: under a full `--no-fail-fast` run, the login-delay escalation
+# (src/service/src/login_delay.rs: >=5 failed logins -> sleep, >=7 -> 60s IP
+# blacklist) ratchets because some tests intentionally perform failed logins
+# (e.g. test_client_secret). Each test passes in isolation; the full-suite cascade
+# is a maintainer-tuning item, not a code defect.
```
sebadob: What is the idea behind this file? You could also 

```
just test-hiqlite
```

which does all of that already.

src/api/src/oidc.rs
```diff
@@ -1008,8 +1008,23 @@ pub async fn post_token(
...
-            // TODO return always the same error here as well, just like during authorize?
-            Err(err)
+            // One uniform error for every credential failure of the password grant:
+            // the caller must not learn whether the user exists, is disabled or
+            // expired, or the password was wrong. `PasswordExpired` is returned
+            // without verifying the password and would leak the account state, so
+            // it is masked too; `PasswordRefresh` (only reachable with the correct
+            // password) keeps its reset-email flow. Non-credential failures
+            // (client errors, rate limits, internal errors) pass through unchanged.
+            match &err.error {
```
sebadob, line 1011: I would remove the type matching. This is very brittle if we ever update an error inside the business logic in the future and forget to update this match.

src/api/src/users.rs
```diff
@@ -65,13 +65,7 @@ use validator::Validate;
...
 /// Returns all existing users
 ///
-/// This endpoint will switch from simply returning all existing users to a server side pagination
-/// dynamically, depending on the threshold set via `SSP_THRESHOLD`.
-/// If the response contains all existing users, the status code will be an HTTP 200.
-/// If the backend is in server side pagination mode, it will return an HTTP 206.
-///
-/// **Permissions**
-/// - rauthy_admin
+// / This endpoint will switch from simply returning all existing users to a server side pagination / dynamically, depen...
```
sebadob: This file (and others) has lots of weird comment truncations for no reason.

src/data/src/entity/api_keys.rs
```diff
@@ -267,6 +267,32 @@ WHERE name = $5"#;
 
         Ok(())
     }
+
+    // / Atomically migrates the encrypted secret + access rights to a new key, but only while / the row still uses `old_kid...
```
sebadob: The space after the first `//` will make this comment local only and remove it from the fn def.

Also, the same weird comment truncation has happened here, whatever it said before.

sebadob (changes requested):
I stopped the review in the middle of it. When you fix all the truncated comments, I need to go over all of these files again anyway.

scripts/bin-tests/setup.sh
reneleonhardt: Good point — this duplicates `just test-hiqlite` entirely: that recipe is a DEBUG build, serves `config-test.toml` on 8081, and that config already has `admin_force_mfa = false` (plus the bootstrap admin/client the suite hardcodes). The script was written before I knew about the justfile recipe, so there is nothing it adds over your flow. Happy to drop it from the PR if you want — just say the word.

sebadob:
Is this ready for another review?

reneleonhardt:
Yes, sorry.

### Review boundary: `2026-08-19T13:19:41Z` (`4972452082`)
- Pushed commits `3d586be`, `fce2843`, `868d49b`, `4459a38`, and `718878a` address trailing-slash wildcard coverage, persistent failed-login escalation, password-grant error normalization, redundant test setup, and refresh-claim error preservation.
- Compliance: all post-review maintainer concerns are addressed; no remaining non-compliance was found in these changes.
- Additional post-review changes: focused trailing-slash wildcard coverage and removal of redundant live-test setup; no other changes were needed after the deep review.

src/api/src/oidc.rs
```diff
@@ -1005,10 +1005,6 @@ pub async fn post_token(
         }
         Err(err) => {
             error!("{}", err.message);
-            if !has_password_been_hashed {
-                return Err(err);
-            }
-            // TODO return always the same error here as well, just like during authorize?
```
sebadob, line 1011: We are not only back to the state before (no unified error), but we also dropped the mandatory custom error message in case of `!has_password_been_hashed` which would break the UI.

src/data/src/entity/auth_providers.rs
```diff
@@ -577,8 +577,7 @@ impl AuthProvider {
             client_secret_basic,
             client_secret_post,
             scope,
-            // TODO add `scopes_supported` Vec and make them selectable with checkboxes in the UI
-            // instead of typing them in?
+            // TODO add `scopes_supported` Vec and make them selectable with checkboxes in the UI instead of typing them in?
```
sebadob: Why removing the line break here? Comments usually have them on purpose to stay within the max line bounds.

src/data/src/entity/clients.rs
```diff
@@ -705,6 +706,29 @@ VALUES ($1, $2, $3, $4)"#;
...
+    /// uses `old_kid`. Returns `true` when the update applied, `false` when the secret was
+    /// changed concurrently (then the migration must not clobber it). Used by the encryption
+    /// key migration (`service::encryption::migrate_encryption_alg`).
+    pub async fn update_secret_migrated(
+        &self,
+        secret: Vec<u8>,
+        new_kid: String,
+        old_kid: &str,
+    ) -> Result<bool, ErrorResponse> {
+        let sql = "UPDATE clients SET secret = $1, secret_kid = $2 WHERE id = $3 AND (secret_kid = $4 OR secret_kid IS NULL)";
```
sebadob: I am pretty sure this line is beyond the max line limit. Use multi-line string in that case and make them fit just like all the other queries.

src/data/src/entity/clients.rs
```diff
@@ -1561,6 +1586,69 @@ impl Client {
...
+        ));
+    }
+    Ok(())
+}
+
+/// Matches wildcard redirect URIs without crossing a host/path boundary.
+#[inline]
+pub fn wildcard_prefix_match(registered: &str, requested: &str) -> bool {
+    let Some(prefix) = registered.strip_suffix('*') else {
+        return false;
```
sebadob, line 1637: When we have this early return we don't need the specific `uri.ends_with('*')` checks above anymore. This line does the same thing, just twice.

src/data/src/entity/clients.rs
sebadob: This var still exists in the current PR, and it's also added to the `struct` below specifically, which is not necessary (as commented).

src/data/src/entity/refresh_tokens.rs
```diff
@@ -47,7 +47,7 @@ impl RefreshToken {
         scope: Option<String>,
         // TODO should we even save mfa for refresh tokens?
         //  even if the original token has been issued with mfa, the refresh
-        //  token not really is, because it can be given without user interaction.
```
sebadob, line 50: The additional space in front of `TODO` comments is on purpose for proper IDE highlighting of the whole TODO comment.

src/data/src/entity/refresh_tokens_devices.rs
```diff
@@ -92,17 +92,17 @@ impl RefreshTokenDevice {
...
-    //     if is_hiqlite() {
-    //         DB::hql().execute(sql, params!(now, user_id)).await?;
-    //     } else {
-    //         DB::pg_execute(sql, &[&now, &user_id]).await?;
-    //     }
-    //
-    //     Ok(())
-    // }
+    pub async fn invalidate_for_user(user_id: &str) -> Result<(), ErrorResponse> {
+        let now = Utc::now().timestamp();
```
sebadob: Since we are using second precision, it's safer to use now - 1 instead, just to really be 100% sure to prevent logic issues inside queries regarding `>` or `>=` conditions.

src/data/src/entity/sessions.rs
```diff
@@ -362,6 +362,14 @@ OFFSET $3"#;
     }
 
     pub async fn upsert(&self) -> Result<(), ErrorResponse> {
+        // Never write expired sessions back.
+        if self.exp < Utc::now().timestamp() {
```
sebadob, line 366: Have you double-checked that we never use the `upsert` in a situation where this could become an issue? It would be good to add a comment for the function mentioning this behavior.

src/data/src/entity/webauthn.rs
```diff
@@ -565,8 +565,7 @@ impl WebauthnAdditionalData {
     pub async fn delete(&self) -> Result<(), ErrorResponse> {
         match self {
             Self::Login(d) => d.delete().await,
-            // The service req data is not deleted here, but actually further down the road
-            // after the service req has been made.
+            // The service req data is not deleted here, but actually further down the road after the service req has been made.
```
sebadob: Removed the line break which made the comment exceed line length limit again.

src/data/src/entity/sessions.rs
```diff
@@ -416,6 +424,26 @@ SET user_id = $3, roles = $4, groups = $5, is_mfa = $6, state = $7, exp = $8, la
         Ok(())
     }
 
+    /// Refreshes `last_seen` without inserting or resurrecting a session.
+    pub async fn touch_last_seen(&self) -> Result<(), ErrorResponse> {
```
sebadob: It would be cleaner when we require `&mut self` here and update the timestamp as well. This would clean up code used in other files below a bit and removes the need to manually set it.

src/service/src/oidc/grant_types/authorization_code.rs
```diff
@@ -95,9 +97,8 @@ pub async fn grant_type_authorization_code(
         ));
     }
 
-    // get the oidc code from the cache
     let idx = req_data.code.as_ref().unwrap().to_owned();
-    let code = match AuthCode::find(idx).await? {
+    let code = match AuthCode::find(idx.clone()).await? {
```
sebadob: I don't think we ever use `AuthCode::find()` in a way that we don't want to delete the code afterwards. So we should be fine reworking it to the new atomic get_remove. We then can get rid of situations where we manually clean it up afterwards. If any of the validations below fail, the whole code should be depleted anyway.

That would make the code that uses an auth code a bit cleaner and easier to follow and maintain.

src/service/src/oidc/grant_types/authorization_code.rs
```diff
@@ -168,6 +169,18 @@ pub async fn grant_type_authorization_code(
         (None, granted) => granted.map(String::from),
     };
 
+    // Claim immediately before issuing tokens.
+    let code: AuthCode = match DB::hql().get_remove(Cache::AuthCode, idx).await? {
```
sebadob: This will be gone when the `find()` is reworked to be atomic, but apart from that, we never want to use DB / Cache calls outside of the `data` module.

src/service/src/oidc/grant_types/authorization_code.rs
```diff
@@ -183,12 +196,12 @@ pub async fn grant_type_authorization_code(
...
     .await?;
 
-    code.delete().await?;
-
     // update session metadata
     if let Some(sid) = code.session_id.clone() {
         let mut session = Session::find(sid).await?;
-        session.set_authenticated(&user).await?;
+        if session.exp > Utc::now().timestamp() {
+            session.set_authenticated(&user).await?;
```
sebadob: The `exp` check should definitely be moved into the `set_authenticated()` function. Preferrably even inside the DB query to avoid race conditions and parallel updates.

src/service/src/oidc/grant_types/device_code.rs
```diff
@@ -13,7 +15,7 @@ use std::net::IpAddr;
 use std::ops::{Add, Sub};
 use tracing::{debug, error, warn};
 
-/// Return a [TokenSet](crate::models::response::TokenSet) for the `device_code` flow
+// Return a [TokenSet](crate::models::response::TokenSet) for the `device_code` flow
```
sebadob: The additional `/` should not be removed to because of the fn doc.

src/service/src/oidc/grant_types/device_code.rs
```diff
@@ -127,10 +130,28 @@ pub async fn grant_type_device_code(peer_ip: IpAddr, payload: TokenRequest) -> H
             None
         };
 
-        if let Err(err) = code.delete().await {
-            // should really never happen - in cache only
-            error!(?err, "deleting DeviceAuthCode");
-        }
+        // Claim immediately before issuing tokens.
+        let code: DeviceAuthCode = match DB::hql()
+            .get_remove(Cache::DeviceCode, code.user_code().to_string())
```
sebadob: We usually always want to call the DB only inside the `data` module. Apart from that, we should do the same as for the `AuthCode` and probably have a `get_remove()` inside `find()`.

DB / Cache operations should be "hidden" inside `data`.

src/service/src/oidc/grant_types/password.rs
sebadob: As mentioned in the review before already. We don't want to return a unified error here, but in the calling API endpoint to really always catch all of them. This is not the right place to do it.

src/service/src/oidc/bcl_logout_token.rs
```diff
@@ -36,8 +36,7 @@ pub struct LogoutToken<'a> {
     pub sub: Option<&'a str>,
     pub sid: Option<&'a str>,
 
-    // The `nonce` MUST NOT exist in this token. We try to deserialize into an `Option<_>` for easy
-    // `.is_none()` validation.
+    // The `nonce` MUST NOT exist in this token. We try to deserialize into an `Option<_>` for easy `.is_none()` validation.
```
sebadob: Once again removed linebreak which makes the comment exceed line length.

src/service/src/oidc/logout.rs
```diff
@@ -70,16 +68,13 @@ pub async fn get_logout_html(
...
+        // boundary); the cheap `ends_with('*')` gate skips non-wildcard entries
         let valid_redirect = uri_vec.as_ref().unwrap().iter().any(|uri| {
-            if uri.ends_with('*') && target.starts_with(uri.split_once('*').unwrap().0) {
-                return true;
-            }
-            if target.eq(uri) {
-                return true;
-            }
-            false
+            (uri.ends_with('*') && wildcard_prefix_match(uri, &target)) || target.eq(uri)
```
sebadob: `uri.ends_with('*')` is checked inside `wildcard_prefix_match()` already.

src/service/src/oidc/validation.rs
```diff
@@ -46,8 +48,7 @@ pub async fn validate_auth_req_param(
                 "'code_challenge' is missing",
             ));
         } else {
-            // 'plain' is the default method to be assumed by the OAuth specification when it is
-            // not further specified.
+            // 'plain' is the default method to be assumed by the OAuth specification when it is not further specified.
```
sebadob: The removed linebreak again.

sebadob (changes requested):

sebadob (approved):
Thanks!

PR #1697, opened 2026-08-12
fix: Enforce strict single-use for auth codes, device codes and PoW challenges

reneleonhardt:
## Summary

The cache get/delete claim was not atomic: two concurrent requests with the same auth code / device code / PoW challenge could both pass and each get a full token set. Claim handling is now serialized per key via a distributed lock (hiqlite `dlock` feature):

- auth-code grant: lock around find-validate-delete; the code is deleted before any token is issued;
- device-code grant: the whole poll runs under the device-code lock;
- PoW reuse check: get+delete under the challenge lock.

Auth codes are cache-only (no atomic DB claim is possible), so the distributed lock is the piece that makes single-use strict. webauthn challenges and MFA-mod tokens stay claim-before-side-effect (idempotent / challenge-bound; kept by decision).

## Verification

`cargo fmt --check` clean; `cargo clippy -- -D warnings` (CI form) — 0 warnings; workspace lib tests: 80 passed / 0 failed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,7 +107,7 @@ gethostname = "1"
 hex = { version = "0.4", features = ["serde"] }
 hickory-resolver = "0.26.1"
 hiqlite = { version = "0.14.0", features = [
-    "cache", "cast_ints", "counters", "dashboard", "listen_notify_local", "macros"
+    "cache", "cast_ints", "counters", "dashboard", "dlock", "listen_notify_local", "macros"
 ] }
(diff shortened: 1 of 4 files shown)


sebadob:
If I remember correctly, this is fully included in #1696, right?

reneleonhardt:
Correct — the fix is fully included in #1696, in its final evolved form: the auth-code/device-code/PoW claims now use hiqlite's atomic `get_remove` (via hiqlite#362) instead of the dlock serialization this PR describes. Closing as superseded — all review stays on #1696. Thanks for the catch!

PR #1698, opened 2026-08-12
fix: Atomically claim refresh tokens and revoke the family on replay

reneleonhardt:
## Summary

The refresh grant did a non-atomic find-then-extend: two concurrent refreshes with one token could both pass, and a stolen token stayed usable for the grace window. Replace with an atomic conditional claim (`UPDATE … WHERE exp > now`) on both `refresh_tokens` and `refresh_tokens_devices`, so only the first use wins and the old token dies on rotation.

Behavior note: `refresh_token_grace_time` no longer keeps the old token usable after a refresh — clients must use the rotated token.

Replay handling: when the atomic claim fails, the presented token was already used — the standard replay / theft signal (RFC 6819 §5.2.1.3). Instead of only rejecting the request, the whole family is revoked: all refresh tokens (regular + device), all issued tokens and all sessions for the user, forcing a clean re-authentication. Tradeoff: a legit client that retries an old refresh token after losing the rotated response also triggers the revocation — clients must use the token from the last response.

## Verification

`cargo fmt --check` clean; `cargo clippy -- -D warnings` (CI form) — 0 warnings; workspace lib tests: 80 passed / 0 failed.

### DIFF
--- a/src/data/src/entity/refresh_tokens.rs
+++ b/src/data/src/entity/refresh_tokens.rs
@@ -168,6 +168,23 @@ impl RefreshToken {
         Ok(slf)
     }
 
+    /// Atomically claims this refresh token for one-time use. Returns `true` if the claim
+    /// succeeded (the token was still valid and this caller won), `false` if it was already
+    /// claimed or expired. The conditional UPDATE (`exp > now`) makes the claim atomic, so
+    /// two concurrent refreshes with the same token cannot both succeed, and a stolen token
+    /// cannot be replayed after legitimate use.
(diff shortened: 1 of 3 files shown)


sebadob:
This is also fully included in #1696, right?

reneleonhardt:
Correct — fully included in #1696, in its final evolved form (`41254dd`): claim with grace window plus gated device-token revocation (`token_revoke_device_tokens`). Closing as superseded — all review stays on #1696. Thanks for the catch!

PR #1700, opened 2026-08-13
perf: 11x - 36x faster password hashing

reneleonhardt:
## Summary

Same Argon2id cost as the old default, 11-36x faster per hash.
Each preset also has a hardening factor vs the [OWASP Argon2id minimum](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) (19 MiB / t=2) = total work m x t - a hardening factor, not a speed one.
- **11x speed, ~13.5x OWASP floor = `hardened`** - same m=128 MiB / t=4 security as main, ~25 ms vs 280 ms, 1x-2x memory.
- **36x speed, ~5x OWASP floor = `memory`** - m=64 MiB / t=3, ~9 ms, 0.5x memory (default preset).

Minimums (1 vCPU, lanes clamped): speed ~4.5x, hardened ~1.5x - the 11-36x headline is the 8-vCPU case, see the k8s table for every size. A config knob (`argon2_preset`) switches speed / memory / hardened; every value stays overridable.

## What changed

1. **Concurrent argon2 workers** - one per configured `max_hash_threads`; the bounded channel caps in-flight memory at `threads x m`.
2. **Default memory preset**; OWASP-endorsed, 3.4× the OWASP minimum - the [OWASP-low profile](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) (m=64 MiB / t=3), 1 in-flight (like main), 0.5x memory (existing hashes re-hash on next login).
3. **`argon2_preset` selector** - `speed | memory | hardened`; every field still overridable (explicit > preset > default).
4. **SIMD argon2-rust backend** (AVX-512/AVX2/SSE2/NEON) - lanes parallelize across cores, p=8 (best measured, clamped to host cores); same PHC format, verify path and memory; `is_argon2_uptodate` uses the crate's `decode_phc`.
5. **Real client IP resolved once per request** (`req.extensions()` cache).
6. **Bounded event pipeline** - channels + semaphore + non-blocking sends; saturated -> drop with a warning, never stall or grow unbounded.
7. **Bounded login-location checks** - per-login geo/notify capped; saturated -> skip.

## Config

```toml
[hashing]
argon2_preset = "memory"  # speed, hardened. default: memory
# argon2_p_cost = 1   # ~4.5x vs main, 1 core - CPU-minimal
# argon2_p_cost = 4   # ~19x  vs main, 4 cores (OWASP standard)
# argon2_p_cost = 8   # ~36x  vs main, up to 8 cores - auto-clamped (default for all presets)
```

## Presets (vs main: 280 ms/hash, 130 MiB)

m = memory (MiB), t = time cost, p = lanes (parallelism within one hash), threads = concurrent hashes (in-flight).

| scenario | config | in-flight | lanes/hash | speed vs main | peak mem | mem vs main |
|---|---|---|---|---|---|---|
| main (old scalar) | 128 / 4 | 1 | 1 | 1x | 130 MiB | 1x |
| hardened, `threads=1` | 128 / 4 / 8 / 1 | 1 | up to 8 | ~11x | 130 MiB | 1x |
| hardened (preset) | 128 / 4 / 8 / 2 | 2 | up to 8 | ~11x | 258.5 MiB | 2x peak |
| speed (preset) | 64 / 3 / 8 / 2 | 2 | up to 8 | ~36x | 130.5 MiB | 1x |
| memory (default) | 64 / 3 / 8 / 1 | 1 | up to 8 | ~36x | 66.2 MiB | 0.5x |

Security margin vs the [OWASP Argon2id minimum](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) (19 MiB / t=2) - hardening factors, not speed:

| preset | m / t | total work (m x t) | vs OWASP floor | speed vs main |
|---|---|---|---|---|
| OWASP minimum | 19 / 2 | 38 MiB-passes | 1x | ~175x (p=8) |
| `speed` | 64 / 3 | 192 MiB-passes | ~5x | ~36x |
| `memory` (default) | 64 / 3 | 192 MiB-passes | ~5x | ~36x |
| `hardened` | 128 / 4 | 512 MiB-passes | ~13.5x | ~11x |
| main (old default) | 128 / 4 | 512 MiB-passes | ~13.5x | 1x |

OWASP min at the official p=1 is ~28x; at p=8 (same lanes as the presets) ~175x - it is the fastest because it does ~7x less work than main, which is exactly why it is the minimum.

**The one number:** `hardened` + `max_hash_threads = 1` = same security, same memory, ~11x faster single-login.

## k8s per node

| preset | 1 vCPU | 2 vCPU | 4 vCPU | 8 vCPU |
|---|---|---|---|---|
| `speed` | 100m/1 256/512 · 4.5x | 100m/2 256/512 · 9.5x | 100m/4 256/512 · 19x | 100m/8 256/512 · 36x |
| `memory` | 100m/1 192/384 · 4.5x | 100m/2 192/384 · 9.5x | 100m/4 192/384 · 19x | 100m/8 192/384 · 36x |
| `hardened` | 100m/1 384/768 · 1.5x | 100m/2 384/768 · 3x | 100m/4 384/768 · 6x | 100m/8 384/768 · ~9x |

Cells: CPU req/lim · MEM req/lim (MiB) · speed vs main. CPU lim = `threads x min(p, cores)`, capped at host.
Request = guarantee, limit = cap: on a busy node you only get the cores you requested - request the speed you want to keep.

## Resource model

- **Security = m + t** (memory per hash + time cost) - fixed per preset; **p does not change them** (lanes share the same m budget).
- **Memory = threads x m** (in-flight hashes x per-hash m): 130.5 / 66.2 / 258.5 MiB peak.
- **CPU = threads x min(p, host cores)** at peak. Each in-flight hash spawns p lane threads (always); free cores decide the per-hash speed (idle vs busy).
- **Latency** = per-hash; `threads` never changes it, p and free cores do.
- `threads=1` = serialized logins like main (1 in-flight, 1x memory) but still ~11x per hash; `p=1` additionally serializes each hash (main's exact profile, CPU-minimal) - a CPU choice, not a security one.

## Verification

`cargo fmt --check` clean; `cargo clippy -- -D warnings` 0; `cargo test --workspace --lib` 90 passed / 0 failed (host) and 79 passed / 0 failed on Linux aarch64 (Colima). Pipeline benchmark wired into CI (criterion, baseline-gated).

### DIFF
--- a/.github/workflows/code_style.yaml
+++ b/.github/workflows/code_style.yaml
@@ -39,3 +39,28 @@ jobs:
       - name: Code Style UI
         working-directory: ./frontend
         run: npm run check
+
+  bench:
+    if: github.event.pull_request.draft == false
+    name: Password Hashing Bench
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/sebadob/rauthy-builder:20260519
+    steps:
+      - uses: actions/checkout@v4
+      # persist the criterion baseline between runs (ephemeral runners) so PRs compare
+      # against the last saved result of the base branch
+      - uses: actions/cache@v4
+        with:
+          path: target/criterion
+          key: criterion-baseline-${{ github.base_ref }}
+          restore-keys: criterion-baseline-
+      - name: Run hashing pipeline benchmark
+        run: |
+          if find target/criterion -maxdepth 3 -type d -name main 2>/dev/null | grep -q .; then
+            # baseline exists: compare (50% leniency against run-to-run noise) then re-save
+            cargo bench --bench hashing -- --save-baseline main --load-baseline main --baseline-lenient 50
+          else
+            # first run: record the baseline only
+            cargo bench --bench hashing -- --save-baseline main
+          fi
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -507,6 +507,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "argon2-rust"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9116a9e4fd543b4e7ae219c3d7f1259b59615105c0c27b3d75939a8b8cf8af86"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "arrayvec"
 version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4956,7 +4965,7 @@ version = "0.36.2"
 dependencies = [
  "actix-web",
  "actix-web-prom",
- "argon2",
+ "argon2-rust",
  "chrono",
  "cidr",
  "clap",
@@ -5026,7 +5035,7 @@ version = "0.36.2"
 dependencies = [
  "actix-web",
  "ammonia",
- "argon2",
+ "argon2-rust",
  "base64 0.22.1",
  "bincode",
  "brotli",
@@ -5060,7 +5069,7 @@ dependencies = [
  "actix-web",
  "actix-web-lab",
  "arc-swap",
- "argon2",
+ "argon2-rust",
  "askama",
  "async-trait",
  "atrium-api",
@@ -5152,7 +5161,7 @@ version = "0.36.2"
 dependencies = [
  "actix-multipart",
  "actix-web",
- "argon2",
+ "argon2-rust",
  "chacha20poly1305",
  "chrono",
  "cryptr",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -411,6 +411,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
+
+[[package]]
 name = "anstream"
 version = "1.0.0"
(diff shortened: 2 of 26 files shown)


sebadob (changes requested):
This PR contains multiple different changes that are not related at all. Can you please split it? Reviewing such big PRs is not fun.

Apart from that, after a very quick first look, we are not changing to `argon2-rust`. This is a brand new crate used by basically no one yet. It has only 300 downloads total, while `argon2` is very well established and used by millions.

It may make sense though to check out the `parallel` feature or others from `argon2`. If I am not mistaken, the `p_cost` does currently not improve performance. Back then when I added password hashing, it had no effect yet and was an open TODO for them. It's possible that they added it in the meantime.

Migrating the password hashing from the brittle channel limitation to a Semaphore was on my TODO since quite a while. This is a very welcome change. It should however now limit with a constant, but actually rexpect the configs `max_hash_threads`.

Edit:

I also don't see a real value in having hashing presets. Usually, you tune argon2 depending on your infrastructure:

- assign as max mem as you can afford (in combination with `max_hash_threads`)
- assign as many cores as you can afford (in combination with `max_hash_threads`)
- bump up `t_cost` until you reach a hashing time of 750-1000ms (makes the most sense for logins imo)

The current default is a safe starting point, that's all it should ever be. I would only lower the `m_cost` if you are really limited on memory.

PR #1701, opened 2026-08-19
fix(data): clean the cache WAL on minor upgrades and bump to 0.37.0-nightly.1

reneleonhardt:
## Summary

Follow-up to #1696, as agreed in the hiqlite#362 review. The hiqlite cache is
persisted in the WAL and restored on restart, so entries written by the previous
version (single-use claims, per-IP abuse counters, removed features) would
linger until the next compaction. The temp migration now clears the affected
namespaces once per minor upgrade.

Also removes the v0.36-cycle temp migrations (PAM group cleanup, group-admin
role warning) and bumps the version to `0.37.0-nightly.1`, raising the lowest
compatible DB version to `0.36.0`.

## Changes (single commit, no overlap with #1696)

1. Version-gated cache WAL cleanup on minor upgrades: `AuthCode`, `DeviceCode`,
   `PoW`, `DPoPNonce`, `AuthProviderCallback`, `ClientDynamic`,
   `ClientEphemeral`, `ClientSecret`, `CredStuffDetect`, `IpBlacklist`,
   `IpRateLimit`, `EmailRateLimit`. Runs only when the stored DB version
   predates the new minor (unit-tested), so restarts within a minor keep
   in-flight flows. Sessions, users and other long-lived caches are kept.
2. `DB::migrate()` returns the pre-upgrade DB version and threads it into the
   temp migrations; fresh installs skip the cleanup.
3. Version bump to `0.37.0-nightly.1`, `LOWEST_COMPATIBLE_VERSION` → `0.36.0`,
   version guard → 0.37, old temp migrations and their TODO markers removed.

## Testing

- Unit: new gate tests + full workspace unit suite green; clippy and rustfmt
  clean.
- Integration: started against a real v0.36.2 DB — cleanup logged, DB version
  upserted to `0.37.0-nightly.1`; a second start skips the cleanup.

## Notes

- hiqlite stays on the git patch until the stable release with `get_remove()`
  lands, as agreed.
- Only ephemeral single-use and abuse-counter state is purged; backend/session
  caches survive the upgrade.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = ["src/*"]
 exclude = ["rauthy-client"]
 
 [workspace.package]
-version = "0.37.0-nightly.1"
+version = "0.37.0-20260819"
 edition = "2024"
 rust-version = "1.95.0"
 authors = ["Sebastian Dobe <sebastiandobe@mailbox.org>"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = ["src/*"]
 exclude = ["rauthy-client"]
 
 [workspace.package]
-version = "0.36.2"
+version = "0.37.0-nightly.1"
 edition = "2024"
 rust-version = "1.95.0"
 authors = ["Sebastian Dobe <sebastiandobe@mailbox.org>"]
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -97,6 +97,7 @@ pub async fn run(
 
     let cache_data_dir = node_config.data_dir.to_string();
     let cache_storage_disk = node_config.cache_storage_disk;
+    // TODO remove this cache-WAL compatibility cleanup in the next minor version.
     rauthy_data::temp_migrations::prepare_cache_wal(&cache_data_dir, cache_storage_disk)
         .await
         .map_err(|err| std::io::Error::other(err.to_string()))?;
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -95,6 +95,12 @@ pub async fn run(
     // init BEFORE Hiqlite to avoid issues in case of misconfiguration
     rauthy_data::ipgeo::init_geo().await;
 
+    let cache_data_dir = node_config.data_dir.to_string();
+    let cache_storage_disk = node_config.cache_storage_disk;
+    rauthy_data::temp_migrations::prepare_cache_wal(&cache_data_dir, cache_storage_disk)
+        .await
+        .map_err(|err| std::io::Error::other(err.to_string()))?;
+
     DB::init(node_config)
         .await
         .expect("Error starting the database / cache layer");
@@ -127,6 +133,10 @@ pub async fn run(
         error!("Error during version migration: {err:?}");
         time::sleep(Duration::from_secs(1)).await;
     }
+
+    rauthy_data::temp_migrations::mark_cache_wal_current(&cache_data_dir, cache_storage_disk)
+        .await
+        .map_err(|err| std::io::Error::other(err.to_string()))?;
 
     UserPicture::test_config().await.unwrap();
 
--- a/src/bin/src/server.rs
+++ b/src/bin/src/server.rs
@@ -107,7 +107,7 @@ pub async fn run(
     tokio::spawn(password_hasher::run());
 
     debug!("Applying database migrations");
-    DB::migrate().await.expect("Database migration error");
+    let previous_db_version = DB::migrate().await.expect("Database migration error");
 
     debug!("Starting Events handler");
     EventNotifier::init_notifiers(tx_email).await.unwrap();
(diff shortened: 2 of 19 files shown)


src/bin/src/server.rs
```diff
@@ -121,7 +121,9 @@ pub async fn run(
     tokio::spawn(watch_health());
 
     // Loop, because you could get into a race condition when recovery a HA Leader after lost volume
-    while let Err(err) = version_migration::manual_version_migrations().await {
+    while let Err(err) =
+        version_migration::manual_version_migrations(previous_db_version.clone()).await
```
sebadob, line 132: No need to `clone()` the `previous_db_version`.

src/data/src/temp_migrations.rs
```diff
@@ -1,52 +1,78 @@
...
+pub async fn apply_temp_migrations(
+    previous_db_version: Option<Version>,
+) -> Result<(), ErrorResponse> {
+    if let Some(previous) = previous_db_version {
+        let app = DbVersion::app_version();
+        if needs_cache_wal_cleanup(&previous, &app) {
+            info!("Cleaning up cache WAL entries written by v{previous}");
+            let client = DB::hql();
+            for cache in CACHE_WAL_CLEANUP {
+                client.clear_cache(cache).await?;
```
sebadob: As I am thinking about it, I am not sure if we can do it this way. 

When we start with existing cache data, the latest snapshot will be applied, which will work fine, but then the WAL file will be replayed for entries not existing in the snapshot. When we deserialize them with the new ordering of the variants, I am not sure if it will only apply "bad" values, or if it could actually crash because of deserilaization issues.

I think it's safer (and easier) to simply delete the whole cache dirs `logs_cache` and `state_machine_cache`.

Cargo.toml
```diff
@@ -4,7 +4,7 @@ members = ["src/*"]
 exclude = ["rauthy-client"]
 
 [workspace.package]
-version = "0.36.2"
+version = "0.37.0-nightly.1"
```
sebadob: Instead of `nightly`, I usually append the date, so you know from when this is, like e.g.

```
0.37.0-20260819
```

sebadob (changes requested):

QexleOSS (commented):
Went through the diff. Clearing the ephemeral caches on a minor bump makes sense, old auth codes and rate counters shouldn't survive a version change.

Agreed with sebadob on the WAL concern. clear_cache only touches the live cache, the on-disk WAL with the old enum variants still replays at startup and after a reorder that can panic on deserialize instead of just applying stale data. Nuking logs_cache and state_machine_cache is the safer fix. If you go that route the CACHE_WAL_CLEANUP loop is dead code, drop it so it doesn't look like the API clear is what's protecting you.

The .clone() at server.rs:125 I think you actually need. The startup loop retries and previous_db_version is moved by value into apply_temp_migrations, so a second retry moves an already-moved Option and won't compile. To drop the clone take &Option<Version> in both functions.

(Heads up for the release notes) IpBlacklist gets wiped on every minor upgrade so any manually kept blocklist goes too. Probably fine but just flag it.

Only needs_cache_wal_cleanup has a test. The migrate() return and the fresh install None skip don't. A small unit test that migrate() returns the previous version would cover the new contract.

Not blocking.

sebadob:
> The .clone() at server.rs:125 I think you actually need. The startup loop retries and previous_db_version is moved by value into apply_temp_migrations, so a second retry moves an already-moved Option and won't compile. To drop the clone take &Option in both functions.

Right, I missed the loop here. Of course we need the clone then.

reneleonhardt:
## Review changes

- Fully complied: pre-start deletion of both cache directories on version change, dead-loop removal, required clone retention, date version, and blacklist/counter warning.
- Maximized unit-test coverage: all important changed production lines and branches are covered by black/white-box unit tests.
- Additional changes after deep Codemap review: moved reset before `DB::init()`, added an atomic version marker with fresh/current/invalid/old/disk-disabled tests, converted helper errors, and behavior-preserving warning fixes.

src/data/src/entity/users.rs
```diff
@@ -2138,13 +2138,13 @@ mod tests {
...
-                    String::from("The password has expired"),
-                ));
-            }
+        if let Some(exp) = user.password_expires
+            && exp < OffsetDateTime::now_utc().unix_timestamp()
+        {
+            return Err(ErrorResponse::new(
+                ErrorResponseType::PasswordExpired,
+                String::from("The password has expired"),
+            ));
```
sebadob, line 2147: When updating this already, please drop `time` in favor of `chrono` for consistency. We also don't need the `String::from()` anymore for the error message. This is a left-over from a pretty early time. No need to allocate here.

src/bin/src/server.rs
```diff
@@ -95,6 +95,12 @@ pub async fn run(
     // init BEFORE Hiqlite to avoid issues in case of misconfiguration
     rauthy_data::ipgeo::init_geo().await;
 
+    let cache_data_dir = node_config.data_dir.to_string();
+    let cache_storage_disk = node_config.cache_storage_disk;
+    rauthy_data::temp_migrations::prepare_cache_wal(&cache_data_dir, cache_storage_disk)
+        .await
+        .map_err(|err| std::io::Error::other(err.to_string()))?;
```
sebadob, line 103: Please add a `TODO` comment above it that reminds of the removal with the next minor version.

sebadob (changes requested):

sebadob (approved):
Thanks

PR #1703, opened 2026-08-22
RFC 8707: carry `resource` through upstream-provider logins (#1702)

### LINKED ISSUE #1702: RFC 8707: `resource` is dropped on upstream-provider logins — the auth code is minted with `resource: None`, and the token request then fails "not granted at the authorization request"
### Summary

When a user authenticates through an upstream auth provider (Google in my case) during `/oidc/authorize`, the RFC 8707 `resource` from the authorization request never reaches the auth code. The token request that follows — which (correctly) repeats the same `resource` — is then rejected:

```
WARN  token request `resource` was not granted at the authorization request
ERROR the requested `resource` was not granted at the authorization request
```

The same authorization request succeeds when the user logs in with password/passkey, or when an existing session goes through `/oidc/authorize/refresh` (the path #1645 fixed). Only the brokered path loses it.

### Where

- `src/service/src/oidc/auth_providers/login_finish.rs` builds the `AuthorizeData` for `finish_authorize` with a literal `resource: None` (there is even a comment: "brokered logins via an upstream IdP do not propagate RFC 8707 resource indicators yet").
- `AuthProviderCallback` (`src/data/src/entity/auth_providers.rs`) has `req_scopes`, `req_state`, `req_nonce`, `req_code_challenge`, … but no `req_resource`, so `login_start.rs` has nothing to store.
- `ProviderLoginRequest` (`src/api_types/src/auth_providers.rs`) has no `resource` field, unlike `LoginRequest` / `LoginRefreshRequest`, which validate it with `RE_RESOURCE`.
- Frontend: `frontend/src/routes/oidc/authorize/+page.svelte` forwards `resource` into the password and refresh payloads but not into the `ProviderLoginRequest` built in `providerLogin()`.

### Why it matters

MCP clients (claude.ai is the concrete one) send `resource` on both the authorization and the token request, as RFC 8707 / the MCP authorization spec require. A user whose first login runs through "Login with Google" cannot complete the connector setup at all; the client sees a token-endpoint error and shows "authentication failed". The only workaround is to log in to Rauthy with password/passkey first so the request takes the session path.

### Repro (v0.36.2)

1. A client with `allowed_resources: ["https://rs.example/mcp"]`; an upstream provider enabled.
2. `GET /oidc/authorize?...&resource=https://rs.example/mcp`, choose the upstream provider, complete its login (+ passkey if enforced).
3. `POST /oidc/token` with `grant_type=authorization_code&code=…&resource=https://rs.example/mcp` → 400 with the error above. Without the `resource` on the token request the exchange succeeds, but the access token then has no resource audience — the client is right to refuse that.

### Proposed fix

Carry it through like the other request parameters: `resource` on `ProviderLoginRequest` (validated with `RE_RESOURCE`, same as `LoginRequest`), `req_resource` on `AuthProviderCallback`, `login_start` stores it, `login_finish` passes `slf.req_resource` instead of `None`; the frontend adds `resource` to the provider-login payload next to `code_challenge`. The `allowed_resources` check already lives in `finish_authorize`/the code path, so no new validation beyond the regex.

Happy to send the PR.

impuls42:
Closes #1702.

A brokered login (upstream auth provider) built its `AuthorizeData` with a literal `resource: None`, so the auth code never carried the RFC 8707 resource indicator and the token request that repeats it was refused with `the requested "resource" was not granted at the authorization request`. Password, passkey and `/oidc/authorize/refresh` (#1645) logins already carried it — only the provider round trip lost it.

**Change**, the same shape as the other request parameters:
- `ProviderLoginRequest.resource: Option<String>` validated with `RE_RESOURCE` (identical rule and doc comment to `LoginRequest.resource`);
- `AuthProviderCallback.req_resource`, `#[serde(default)]` so a callback cached by the previous version still decodes during a rolling upgrade;
- `login_start` stores it, `login_finish` passes `slf.req_resource` to `finish_authorize` (the `allowed_resources` check already lives there, nothing new to validate);
- frontend: the authorize page adds `resource` to the provider-login payload next to the PKCE values, and the TS type gains the field.

**Tests**: two unit tests on `ProviderLoginRequest` mirroring the #1645 ones (a valid resource deserializes and validates; a fragment is rejected). `cargo fmt`, `cargo check` and `cargo clippy` clean on `rauthy-api-types`, `rauthy-data`, `rauthy-service`.

Seen in the wild with claude.ai's MCP connector: a user whose first Rauthy login runs through "Login with Google" cannot finish the connector setup; the same flow through a password/passkey login works.

### DIFF
--- a/src/data/src/entity/auth_providers.rs
+++ b/src/data/src/entity/auth_providers.rs
@@ -648,6 +648,11 @@ pub struct AuthProviderCallback {
     pub req_nonce: Option<String>,
     pub req_code_challenge: Option<String>,
     pub req_code_challenge_method: Option<String>,
+    /// RFC 8707 resource indicator from the authorization request (#1702).
+    /// `default` so a callback cached by the previous version still decodes
+    /// during a rolling upgrade.
+    #[serde(default)]
+    pub req_resource: Option<String>,
 
     pub provider_id: String,
 
--- a/frontend/src/api/types/auth_provider.ts
+++ b/frontend/src/api/types/auth_provider.ts
@@ -71,6 +71,8 @@ export interface ProviderLoginRequest {
     /// Validation: PATTERN_CODE_CHALLENGE
     code_challenge?: string;
     code_challenge_method?: CodeChallengeMethod;
+    /// Validation: PATTERN_RESOURCE — RFC 8707 resource indicator
+    resource?: string;
 
     // values for the callback from upstream
     /// Validation: PATTERN_URI
--- a/frontend/src/routes/oidc/authorize/+page.svelte
+++ b/frontend/src/routes/oidc/authorize/+page.svelte
@@ -456,6 +456,7 @@
             nonce: nonce,
             code_challenge: challenge,
             code_challenge_method: challengeMethod,
+            resource: resource || undefined,
             provider_id: id,
             pkce_challenge: '',
             pow: '',
--- a/src/api_types/src/auth_providers.rs
+++ b/src/api_types/src/auth_providers.rs
@@ -1,7 +1,7 @@
 use crate::cust_validation::validate_vec_scopes;
 use rauthy_common::regex::{
-    RE_ALNUM, RE_ATPROTO_HANDLE, RE_CLIENT_ID, RE_CLIENT_NAME, RE_CODE_CHALLENGE, RE_SCOPE_SPACE,
-    RE_URI,
+    RE_ALNUM, RE_ATPROTO_HANDLE, RE_CLIENT_ID, RE_CLIENT_NAME, RE_CODE_CHALLENGE, RE_RESOURCE,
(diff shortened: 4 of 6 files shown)


src/data/src/entity/auth_providers.rs
```diff
@@ -648,6 +648,11 @@ pub struct AuthProviderCallback {
     pub req_nonce: Option<String>,
     pub req_code_challenge: Option<String>,
     pub req_code_challenge_method: Option<String>,
+    /// RFC 8707 resource indicator from the authorization request (#1702).
+    /// `default` so a callback cached by the previous version still decodes
+    /// during a rolling upgrade.
+    #[serde(default)]
```
sebadob, line 654: The `#[serde(default)]` is unnecessary when the data type is an `Option<_>` already.

sebadob (changes requested):

PR #1706, opened 2026-08-26
Write theme saturation and lightness with explicit % units

andrewleech:
Thanks for Rauthy — it's been robust to run and straightforward to operate, and
the device-code flow made our mobile sign-in much simpler than it would
otherwise have been.

We use it as the OIDC provider for a small notes app. On mobile, users leave the
app and land on Rauthy's device and login pages. While testing our iOS build on
an older iPhone (iOS 15.8), the buttons on those pages were invisible — present
and tappable, but with nothing drawn.

`append_css` writes `--action:0 45 33`, and the stylesheets use that as
`hsl(var(--action))` and `hsla(var(--action) / .93)`. Unitless saturation and
lightness are a CSS Color 4 addition, and where it isn't supported the whole
declaration is dropped — so the button loses its background while
`--btn-text: white` still applies, giving white on white.

`frontend/src/css/theme.css` already writes the percent form
(`--action: 34 100% 40%`), so this brings the generated CSS in line with the
checked-in stylesheet rather than introducing a new convention.

Measured on the device with a probe page reporting `CSS.supports` and
`getComputedStyle`:

| declaration | iOS 15.8.8 | iOS 18.6 |
| --- | --- | --- |
| `hsl(0 45 33)` | false → `rgba(0, 0, 0, 0)` | true → `rgb(122, 46, 46)` |
| `hsl(0 45% 33%)` | true → `rgb(122, 46, 46)` | true → `rgb(122, 46, 46)` |
| `hsla(0 45 33 / .93)` | false → `rgba(0, 0, 0, 0)` | true → `rgba(122, 46, 46, 0.93)` |
| `hsla(0 45% 33% / .93)` | true → `rgba(122, 46, 46, 0.93)` | true → `rgba(122, 46, 46, 0.93)` |

Per MDN's compat data the unitless form arrived in Safari 18, Chrome 121 and
Firefox 122, so it may be worth noting this affects every iOS below 18 rather
than only older hardware — and on iOS there's no alternative engine to fall back
to. The slash-alpha syntax itself is fine on iOS 15 once the components carry
units (last row above), so the `hsla(var(--x) / …)` sites are covered by the same
change.

What this touches:

- `ThemeCss::append_css`, the backend emitter.
- `fed_cm.rs`, which builds the same strings for the browser's FedCM UI. FedCM
  shipped in Chrome ~108 but unitless `hsl()` only parses from 121, so those
  branding colours were being dropped on 108–120.
- Three places the admin UI builds `hsl()` of its own, so the branding editor
  works on the same browsers as the pages it previews: the branding preview, the
  colour swatches, and the slider track gradients. The gradients matter most —
  one invalid colour stop invalidates the whole `linear-gradient`, so all three
  slider tracks render flat and you can't see the hue you're dragging toward.

The emails benefit too, without a change: `templates/email/base.html` and
`email/custom.rs` consume the same variables, and `%` is the older and more
widely supported form.

A second commit fixes an unrelated stray digit on a line I was already touching
(`hsl(3600 …)` → `hsl(360 …)` in the hue gradient). Kept separate so you can
drop it independently.

One caveat, with no code change proposed since it's your call: the theme
endpoint is served `max-age=31104000` and busts on `last_update`. For the default
theme that's `BUILD_TIME`, so an upgrade fixes it. For a client with a row in
`themes`, `last_update` isn't touched by an upgrade, so a browser that cached the
unitless CSS keeps it until the theme is next saved. Server-side caching is fine
— `Cache::Html` is cleared at startup.

Checks: `cargo fmt --check` and `prettier -c` are clean on the changed files. I
have not run `cargo clippy -D warnings` or `just pre-pr-checks` — the Hiqlite and
Postgres tests need more than my current setup. Only literal text inside the
format strings moved, with arity and types unchanged, but please treat those as
unverified on my side. I also looked for anything that parses the emitted CSS or
reads these variables back and didn't find any — the admin editor round-trips
through the JSON API as numbers — though a second opinion there is worth more
than mine.

Unrelated, and only in case it's news: `ThemeCss::validate` looks like it calls
`validate_hsl(self.accent)` twice and never checks `self.action`. It doesn't
affect this change, since both forms clamp out-of-range values identically.

Happy to add a CHANGELOG entry, or to take this a different way if you'd prefer
— for instance leaving the variables unitless and adding units at the point of
use instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,12 @@ theme's timestamp, which is what busts the long-lived client-side cache for the 
 
 [#1706](https://github.com/sebadob/rauthy/pull/1706)
 
+### Bugfix
+
+- The last color stop of the hue slider in the Admin UI branding editor used a hue of `3600`
+  instead of `360`, so the gradient ended on red instead of spanning the full spectrum.
+  [#1706](https://github.com/sebadob/rauthy/pull/1706)
+
 ## v0.36.2
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,8 @@ theme's timestamp, which is what busts the long-lived client-side cache for the 
 - The last color stop of the hue slider in the Admin UI branding editor used a hue of `3600`
   instead of `360`, so the gradient ended on red instead of spanning the full spectrum.
   [#1706](https://github.com/sebadob/rauthy/pull/1706)
+- Theme validation checked `accent` twice and never validated `action`.
+  [#1706](https://github.com/sebadob/rauthy/pull/1706)
 
 ## v0.36.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,20 @@ RE_CLIENT_NAME: ^[\p{L}\p{M}\p{N}\p{Zs}()._-]{2,128}$
 
 [#1708](https://github.com/sebadob/rauthy/pull/1708)
 
+#### Theme CSS uses explicit percent units
+
+The generated theme CSS now writes saturation and lightness with an explicit `%`, so a color is
+emitted as `--action: 34 100% 40%` rather than `--action: 34 100 40`. Unitless values inside `hsl()`
+are a CSS Color 4 addition supported from Safari 18, Chrome 121 and Firefox 122. Browsers below that
+drop the whole declaration, which left buttons with no background while `--btn-text` still applied,
+rendering them invisible. It affects every iOS below 18, where no alternative browser engine is
+available.
+
+If you use a custom theme, save it once after upgrading even if you change nothing. That updates the
+theme's timestamp, which is what busts the long-lived client-side cache for the generated CSS.
+
+[#1706](https://github.com/sebadob/rauthy/pull/1706)
+
 ## v0.36.2
 
 ### Security
--- a/frontend/src/lib/admin/clients/branding/BrandingPreview.svelte
+++ b/frontend/src/lib/admin/clients/branding/BrandingPreview.svelte
@@ -17,6 +17,8 @@
         typ: 'light' | 'dark';
     } = $props();
 
+    const hslVar = (v: number[]) => `${v[0]} ${v[1]}% ${v[2]}%`;
+
     let isLoading = $state(false);
     let showDefault = $state(false);
 
@@ -38,13 +40,13 @@
 <div
     aria-label="Preview: All components inside are only for theme and colors preview and have no effect or interaction"
     style:--border-radius={borderRadius}
-    style:--text={theme.text.join(' ')}
-    style:--text-high={theme.text_high.join(' ')}
-    style:--bg={theme.bg.join(' ')}
-    style:--bg-high={theme.bg_high.join(' ')}
-    style:--action={theme.action.join(' ')}
-    style:--accent={theme.accent.join(' ')}
-    style:--error={theme.error.join(' ')}
+    style:--text={hslVar(theme.text)}
+    style:--text-high={hslVar(theme.text_high)}
+    style:--bg={hslVar(theme.bg)}
+    style:--bg-high={hslVar(theme.bg_high)}
+    style:--action={hslVar(theme.action)}
+    style:--accent={hslVar(theme.accent)}
+    style:--error={hslVar(theme.error)}
     style:--btn-text={theme.btn_text}
     style:--theme-sun={theme.theme_sun}
     style:--theme-moon={theme.theme_moon}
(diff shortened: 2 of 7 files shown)


sebadob:
> frontend/src/css/theme.css already writes the percent form
(--action: 34 100% 40%), so this brings the generated CSS in line with the
checked-in stylesheet rather than introducing a new convention.

Yes, makes sense. The file in the frontend is not really used at all. It's more for quick and dirty testing some things in the DEV UI if necessary. But makes total sense to have the convention with more compat.

> Per MDN's compat data the unitless form arrived in Safari 18, Chrome 121 and
Firefox 122, so it may be worth noting this affects every iOS below 18 rather
than only older hardware — and on iOS there's no alternative engine to fall back
to. The slash-alpha syntax itself is fine on iOS 15 once the components carry
units (last row above), so the hsla(var(--x) / …) sites are covered by the same
change.

I am not that familiar with Apple. Didn't know that old devices don't get new browser versions, even when they still get security updates. If iOS 15 would be EOL, I probably would not care too much, but even then it's an easy fix for old devices without breaking current ones.

> Only literal text inside the
format strings moved, with arity and types unchanged, but please treat those as
unverified on my side. I also looked for anything that parses the emitted CSS or
reads these variables back and didn't find any — the admin editor round-trips
through the JSON API as numbers — though a second opinion there is worth more
than mine.

These values are validated only once during updates, which means we know for sure that the saved data inside the DB is valid, so there is no need to re-validate all the time. They are just forwarded from that point on. The HSL values are kept as numbers only to make it more efficient and almost self-validating. Integers are quickly and easily validated.

The Admin UI branding preview (`frontend/src/lib/admin/clients/branding/BrandingPreview.svelte`) builds CSS values for preview without the `%`. However, that would only not work on iOS 15 devices, and it's very unlikely that you are using such one to update branding on the Admin page. For completeness though, these could be updated as well.

> Unrelated, and only in case it's news: ThemeCss::validate looks like it calls
validate_hsl(self.accent) twice and never checks self.action. It doesn't
affect this change, since both forms clamp out-of-range values identically.

That's clearly a bug. Would be nice if you could fix this one line as well.

> Happy to add a CHANGELOG entry (...)

Yes it would be nice if you could add it to the *UNRELEASED* section.  This makes releases a lot less work and gives you a free preview of what's coming with the next release. When you are working on a PR, it's easy to write the changelog, but when doing it during release, I need to read through everything again.

When doing this, it would also be nice a add a hint for all people that are using custom themes that they should probably at least update once, even when they don't change anything. Another save of a template will update the timestamp and therefore cache bust. We could also do it automatically in temp migrations, but that would cache bust it way more often than necessary.

---

The code looks good already, thanks.

andrewleech:
Thanks for the quick look. Both done, and I rebased onto current `main` while I was there.

**`ThemeCss::validate`** — fixed as its own commit. It now checks `action` and `accent` once each, in the struct's field order.

**CHANGELOG** — added to `UNRELEASED`: a `#### Theme CSS uses explicit percent units` note under `Changes`, and a `Bugfix` entry for each of the two fixes. The custom-theme hint is in there as you described — save a theme once after upgrading, even unchanged, so the timestamp bumps and the cache busts. No migration.

The rebase was needed for the changelog rather than for its own sake: the branch was based on a commit that predates the `UNRELEASED` section, so the entry had nowhere correct to go and would have collided with yours on merge. Only the two existing commits were rewritten; the code in them is byte-identical to what you already reviewed.

Two notes, neither needing anything from you:

The Admin UI branding preview is already in the PR — `BrandingPreview.svelte`, `HSLInputRange.svelte` and `InputRange.svelte` were all in the first push. The slider gradients were the ones worth having: one invalid colour stop invalidates the whole `linear-gradient`, so all three tracks rendered flat rather than just losing a swatch. Easy to miss if you were reading the Rust diff.

On the values being known-valid in the DB: that held for every field except `action`, which is the one `validate` never checked. It matters only at the edges — the Admin UI sliders bound the ranges, and `validate_hsl` only rejects hue > 360 or s/l > 100 — but it does mean an operator who somehow stored an out-of-range `action` through the API would now get a 400 the first time they follow the changelog advice and re-save. Rendering was never affected either way, since both the unitless and percentage forms clamp identically.

One thing I left alone deliberately: `frontend/src/hooks.server.js` has the same unitless triples in its `themeDefault` string. As far as I can tell that binding is never referenced — the page takes its theme from the stylesheet link in `app.html` — so I did not want to grow the PR for something with no behavioural effect. Happy to include it, or to leave it.

sebadob (approved):
Thanks!

sebadob:
> One thing I left alone deliberately: frontend/src/hooks.server.js has the same unitless triples in its themeDefault string. As far as I can tell that binding is never referenced — the page takes its theme from the stylesheet link in app.html — so I did not want to grow the PR for something with no behavioural effect. Happy to include it, or to leave it.

You can ignore server hooks. They only exist during development and for pre-rendering at compile time. In prod, there will never be any JS server-side.

PR #1713, opened 2026-09-03
Allow API keys to delete user passkeys

### LINKED ISSUE #1711: Allow Users/Delete API keys to delete user passkeys
`GET /auth/v1/users/{id}/webauthn` already accepts an API key with `Users: Read`, but `DELETE /auth/v1/users/{id}/webauthn/delete/{name}` currently requires a session. This prevents a non-browser administration CLI from resetting passkeys for a locked-out user.

Would you accept allowing an API key with `Users: Delete` on the delete endpoint, while preserving the existing full-admin, group-admin, self-session + `mfa_mod_token`, and CSRF paths?

Draft implementation: #1713

caniko:
## Summary

- allow API keys with `Users` + `Delete` to call the user passkey deletion endpoint
- preserve the existing full-admin, group-admin, self-session, MFA modification token, and CSRF paths
- cover rejection of `Users` + `Read` and acceptance of `Users` + `Delete`

Closes #1711

`just pre-pr-checks` was not run. Linking the full integration-test suite exceeded the local host memory limit, so this remains a draft until the required Hiqlite and PostgreSQL suites run on a capable host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,12 @@ theme's timestamp, which is what busts the long-lived client-side cache for the 
 
 [#1706](https://github.com/sebadob/rauthy/pull/1706)
 
+#### API Key Passkey Deletion
+
+API Keys with `Users` + `Delete` can now call
+`DELETE /auth/v1/users/{id}/webauthn/delete/{name}`. Passkey registration stays
+session-only.
+
 ### Bugfix
 
 - The last color stop of the hue slider in the Admin UI branding editor used a hue of `3600`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,8 +108,7 @@ theme's timestamp, which is what busts the long-lived client-side cache for the 
 #### API Key Passkey Deletion
 
 API Keys with `Users` + `Delete` can now call
-`DELETE /auth/v1/users/{id}/webauthn/delete/{name}`. Passkey registration stays
-session-only.
+`DELETE /auth/v1/users/{id}/webauthn/delete/{name}`.
 
 ### Bugfix
 
--- a/src/api/src/users.rs
+++ b/src/api/src/users.rs
@@ -1928,6 +1928,7 @@ pub async fn post_webauthn_auth_finish_login(
 ///
 /// **Permissions**
 /// - rauthy_admin
+/// - API key with Users + Delete
 /// - authenticated and logged in user for this very {id}
 #[utoipa::path(
     delete,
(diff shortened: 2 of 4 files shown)


src/api/src/users.rs
```diff
@@ -1951,40 +1952,41 @@ pub async fn delete_webauthn(
...
 
-    // Note: Currently, this is not allowed with an ApiKey on purpose.
-    // Access tiers:
-    // - full Rauthy admin: may reset MFA for any user,
-    // - group admin: may reset MFA for a user it manages,
-    // - the user itself: only with a valid `mfa_mod_token`.
-    if principal.validate_admin_session().is_ok() {
-        if principal.is_user(&id).is_err() {
-            warn!("Passkey delete from admin for user {} for key {}", id, name);
+    match principal.validate_api_key(AccessGroup::Users, AccessRights::Delete) {
```
sebadob: Unnecessarily complicated manual `Result<_>` handling. Just use the combining function `validate_api_key_or_admin_session()` which exists exactly for these cases. We don't need to know in detail if an Admin or an API key deleted it. Only an "admin" can actually have an API key. The only situation we care about here is when it was anything else than the user the key belongs to.

Honestly, this whole function is written more complicated than it should be, and it happened when someone else introduced the group admin feature. For instance, the `principal.validate_session_auth()?` check is unnecessary. You would never get here with a session in non-auth state.

When we actually allow API keys to delete here as well, we must add the info mail to users about it in the same PR. An API key is less secure than a full admin session, and this is a very critical request, which is why it was even not allowed so far.

CHANGELOG.md
```diff
@@ -105,6 +105,12 @@ theme's timestamp, which is what busts the long-lived client-side cache for the
 
 [#1706](https://github.com/sebadob/rauthy/pull/1706)
 
+#### API Key Passkey Deletion
+
+API Keys with `Users` + `Delete` can now call
+`DELETE /auth/v1/users/{id}/webauthn/delete/{name}`. Passkey registration stays
```
sebadob: No need to mention that passkey reg is session only. It's technically impossible to let an API key do it.

sebadob (changes requested):

src/api/src/users.rs
caniko: I guess we should mention these assumptions in the docstring to avoid any unwanted consequences in the future

CHANGELOG.md
caniko: Removed that sentence; the changelog now only documents the newly supported API-key deletion.

src/api/src/users.rs
caniko: Good call. I simplified the branch to use `validate_api_key_or_admin_session()` for admin/API-key requests, kept the existing delegated group-admin scope check, and removed the redundant session-auth check. Non-self deletion now emails the affected user through the existing event mailer.

sebadob (approved):
Thanks

PR #1724, opened 2026-09-12
fix typo in Postgres migration `V27__one_time_password.sql`

justprosh:
`migrations/postgres/V27__one_time_password.sql` (from #1620) declares `secret BYTEA NUT NOT NULL`. Postgres rejects that line with a syntax error (`SqlState 42601`), so a Postgres-backed Rauthy built from `main` cannot start: the migration runner stops at V27, both on a fresh database and on one at the v0.36.2 schema. The hiqlite migration (`32_one_time_password.sql`, `BLOB`) is fine, which is probably why it went unnoticed.

This drops the stray `NUT`, nothing else.

Checked against Postgres 17: a fresh database now migrates through V27–V29 and the server starts; a database at the v0.36.2 schema upgrades the same way.


sebadob:
> (...) so a Postgres-backed Rauthy built from main.

...which you hopefully never do in production, running anything from `main`, but yes this is obviously a weird error that got in during refactoring. Did not run Postgres tests yet. Thanks.

